PR #236, opened 2016-09-18
Support FetchMode.SELECT ManyToMany relationships

JWGmeligMeyling:
**Please review before merging**

Some initial work of extending the Collection persisters so that `FetchMode.SELECT` is supported for `ManyToMany` relationships as well (and potentially also `ManyToOne`). 

~~The outputted query for the extended test case is now~~

```sql
WITH RECURSIVE Issue227Test$RecursiveEntityCte(id, name, parent_id) AS
  (SELECT issue227te0_.id AS col_0_0_, issue227te0_.name AS col_1_0_, issue227te0_.parent_id AS col_2_0_
   FROM issue_227_recursive_entity issue227te0_
   WHERE issue227te0_.parent_id IS NOT NULL
   UNION ALL SELECT issue227te0_.id AS col_0_0_, issue227te0_.name AS col_1_0_, issue227te0_.parent_id AS col_2_0_
   FROM issue_227_recursive_entity issue227te0_
   CROSS JOIN Issue227Test$RecursiveEntityCte issue227te1_
   WHERE issue227te0_.id=issue227te1_.parent_id )
SELECT manytomany0_.other_scope_id AS other_sc1_2_1_,
       manytomany0_.scope_id AS scope_id2_3_1_,
       issue227te1_.id AS id1_2_0_,
       issue227te1_.name AS name2_2_0_,
       issue227te1_.parent_id AS parent_i3_2_0_
FROM many_to_many_children manytomany0_
INNER JOIN issue_227_recursive_entity issue227te1_ ON manytomany0_.scope_id=issue227te1_.id
WHERE manytomany0_.other_scope_id IN
    (SELECT issue227te1_.id
     FROM issue_227_recursive_entity issue227te0_
     WHERE issue227te0_.parent_id IS NOT NULL
     UNION ALL SELECT issue227te0_.id AS col_0_0_,
                      issue227te0_.name AS col_1_0_,
                      issue227te0_.parent_id AS col_2_0_
     FROM issue_227_recursive_entity issue227te0_
     CROSS JOIN Issue227Test$RecursiveEntityCte issue227te1_
     WHERE issue227te0_.id=issue227te1_.parent_id )
  SELECT issue227te0_.id AS id1_0_0_,
         issue227te1_.id AS id1_2_1_,
         issue227te0_.name AS name2_0_0_,
         issue227te0_.parent_id AS parent_i3_0_0_,
         issue227te1_.name AS name2_2_1_,
         issue227te1_.parent_id AS parent_i3_2_1_
  FROM Issue227Test$RecursiveEntityCte issue227te0_
  LEFT OUTER JOIN issue_227_recursive_entity issue227te1_ ON issue227te0_.parent_id=issue227te1_.id )
```

~~This query fails with the following exception:~~

`ERROR: Column count does not match; SQL statement:`

~~I think this is from the `UNION ALL SELECT` in the subselect. Still trying to figure out what would be the best way of rewriting the query in my newly added `CustomSubselectCollectionLoader`.~~

Review is very much appreciated!


integration/hibernate-4/src/main/java/com/blazebit/persistence/impl/hibernate/CustomPersisterClassResolver.java
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * @author Jan-Willem Gmelig Meyling
+ * @since 1.2.0
+ */
+@ServiceProvider(PersisterClassResolver.class)
+public class CustomPersisterClassResolver extends StandardPersisterClassResolver implements PersisterClassResolver {
+
+    @Override
+    public Class<? extends CollectionPersister> getCollectionPersisterClass(Collection metadata) {
+        return metadata.isOneToMany() ? CustomOneToManyPersister.class : CustomBasicCollectionPersister.class;
```
JWGmeligMeyling: This is very similar to what the `StandardPersisterClassResolver` actually does.

integration/hibernate-4/src/main/java/com/blazebit/persistence/impl/hibernate/Hibernate4Integrator.java
```diff
@@ -36,7 +37,8 @@ public void integrate(Configuration configuration, SessionFactoryImplementor ses
 				// TODO: check that no collections are mapped
 			}
 		}
-		CommonHibernateUtils.setCollectionPersisterClass(configuration.getClassMappings(), CustomOneToManyPersister.class);
+
+		serviceRegistry.locateServiceBinding(PersisterClassResolver.class).setService(new CustomPersisterClassResolver());
```
JWGmeligMeyling: Perhaps this is not the best way to provide Hibernate services, as I am not really familiar with the SPI. Any input?

JWGmeligMeyling:
Fixed the last problem so the test now passes on Hibernate 4. How do I run the test for all versions of Hibernate?

integration/hibernate-4/src/main/java/com/blazebit/persistence/impl/hibernate/Hibernate4Integrator.java
beikov: Honestly I have no idea how to do that properly. If the tests run through, I am fine with any approach.

JWGmeligMeyling: I've been doing some manual testing in my own project and for me it seems to work on Hibernate 5.2 as well! :smile:

beikov: The only thing that the resolver does not take into account as far as I can see is printing out a warning if the user supplies a custom persister. What do you think, is it better to print the warning or to write documentation about how the fetch support works and let the user figure out that the custom persister is the reason why it doesn't work?

beikov:
To run the tests with e.g Hibernate 5.2 run with the maven profiles "hibernate-5.2,h2". Look into the pom.xml of the testsuite projects for other available profiles.

integration/hibernate-4/src/main/java/com/blazebit/persistence/impl/hibernate/Hibernate4Integrator.java
JWGmeligMeyling: Ah yes, printing a warning using this mechanism is not easily done, because Hibernate only invokes the `PersisterClassResolver` if a custom persister is not defined. On the other hand, it seems that using the `PersisterClassResolver` is more in line with the Hibernate SPI. Furthermore it seems very rare to me that one would use a custom persister, with CTE queries, and the `FetchMode.SUBSELECT`.

So adding some notes on how the fetch support works seems sufficient to me. I will amend that to my PR.

JWGmeligMeyling:
I have added some notes to the documentation. Do you think this would be a logical place?

Issue #257, opened 2016-09-29
Allow specifying fetches for relations in entity views


beikov:
Contrary to what I thought was possible with JPQL when creating #146 it actually is possible to define fetches for sub-relations. When #190 is done, we can also support doing fetches in entity views. I propose to add a new method `String[] fetches() default {}` to `Mapping` that can be used to specify fetches for plain entities or embeddables.
Also update the documentation about entity mappings when this is done.

Issue #190, opened 2016-02-15
Allow fetching when doing relation selects


beikov:
Currently, we only allow to do fetch joins when doing entity selects.
First of all we need more tests to cover all the possible fetch scenarios like
- Selecting subrelation that uses fetch joins `SELECT b FROM A a JOIN a.toOne b LEFT JOIN FETCH b.myCollection`
- Fetching in strict mode, meaning that no aliases are allowed and no further dereference is allowed
- Arbitrary fetching depth

beikov:
I just tried out selecting a subrelation that uses fetch joins with Hibernate and it does seem to have a bug regarding element collections: https://hibernate.atlassian.net/browse/HQLPARSER-15

PR #389, opened 2017-03-15
#190 - Test and fix for fetch joins with scalar selects

beikov:
This is required for being able to implement #257 so hurry with reviewing ;)


core/impl/src/main/java/com/blazebit/persistence/impl/BaseSubqueryBuilderImpl.java
```diff
@@ -72,7 +72,7 @@ public BaseSubqueryBuilderImpl(MainQuery mainQuery, AliasManager aliasManager, J
     public T getResult() {
         return result;
     }
-    
+
```
Mobe91: please revert

PR #390, opened 2017-03-15
Fixed #198 - Support for OneToOne regarding single valued association id optimizations

### LINKED ISSUE #198: Implement proper one-to-one support
Currently we have some TODOs in the code for one-to-one mappings. 

One example is the determination of the foreign key column location for to-one mappings to decide if a join is necessary to use the id of the relation. Knowing the location can make it possible to avoid a join.

beikov:


beikov:
Merged 3e12fa1c30be2f77773ab5894d81274a930abef6

PR #394, opened 2017-03-19
fix some english prepositions

andyjefferson:
Fixes a few minor English prepositions in your front page

### DIFF
--- a/website/src/main/jbake/content/index.html
+++ b/website/src/main/jbake/content/index.html
@@ -21,7 +21,7 @@ status=published
   -->
 
 <section class="bTop clearfix" style="margin-top: 40px;">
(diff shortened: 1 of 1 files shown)


beikov:
Thanks for the fixes, but I am not sure about the _annoyed at_ part. I'm no native speaker but it doesn't feel right and apparently _annoyed of_ seems to be valid: https://www.quora.com/Which-prepositions-are-used-after-annoyed
Would you mind explaining why it should be _annoyed at_?

andyjefferson:
I personally would not use "annoyed" with reference to the Criteria API. I would say more like "frustrated with ...". The fact remains though, no native speaker (of which I am one) would ever say "annoyed of", it just doesn't exist.

beikov:
Thanks, went with "frustrated" as that sounded good :)

PR #400, opened 2017-03-23
Fixed #341 - Eq/In/Nullness Predicate rewrites and Parser/Perf. fixes

### LINKED ISSUE #341: Rewrite nullness and binary predicate for relation to use the id instead
Expressions like `alias.relation IS NULL` might generate a join for the relation. Hibernate wrongly does so, but we could easily workaround that by rendering `alias.relation.id IS NULL`. We should actually employ workarounds like that to improve cross jpa provider compatibility. On the other hand, Datanucleus and EclipseLink generate the join only when seeing expressions like `alias.relation.id IS NULL` but might not generate a join when having `alias.relation IS NULL` so we should rewrite that too.
The same optimization could be done for binary predicates i.e. rewrite `alias1.relation = alias2.relation` to `alias1.relation.id = alias2.relation.id` and vice versa depending on what avoids the join for a jpa provider.

beikov:
## Description

The use of entity type equals/in in pre-Hibernate 5.2.8 is rewritten to id equals/in.
For non-Hibernate the use of id equals/in is rewritten to entity equals/in.
Contains parser and performance fixes that were encountered during development.

## Related Issue

#341


core/parser/src/main/antlr4/imports/JPQL_lexer.g4
```diff
@@ -209,7 +209,9 @@ fragment NOT_FRAG : [Nn][Oo][Tt];
 
 fragment Date_string : DIGIT DIGIT DIGIT DIGIT '-' DIGIT DIGIT '-' DIGIT DIGIT;
 
-fragment Time_string : DIGIT DIGIT? ':' DIGIT DIGIT ':' DIGIT DIGIT ('.' DIGIT*)?;
+fragment Time_string : DIGIT DIGIT? ':' DIGIT DIGIT ':' DIGIT DIGIT;
+
+fragment Nanos_string : ('.' DIGIT*)?;
```
Mobe91: What is the purpose of pulling out `Nanos_string`?

beikov: A normal JDBC escaped time has no nanos, so this was wrong before. Only the timestamp has nanos.

beikov:
Merged by 70fec2c4a26050a61fd6f6d96d8ee82eb2967e7f

PR #403, opened 2017-03-24
Fixed #357 - Added support for a 'this' expression in entity views

### LINKED ISSUE #357: Allow mapping EmbeddableEntityView to `this`
We should allow to have EmbeddableEntityViews that map to an entity type.
When we introduce the expression for referring to `this` in #312 we can effectively use this mechanism to put some attributes into sub-objects elegantly.
Update the mapping documentation when implementing this.

beikov:
## Description

Most of the changes are tests. Basically I adapted the mapping validation checks to allow `this` and replaced the `this` in the expressions with the view root.

## Related Issue

#357

## Motivation and Context

This is the foundation for treat support.

### DIFF
--- a/entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/metamodel/AbstractAttribute.java
+++ b/entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/metamodel/AbstractAttribute.java
@@ -20,6 +20,7 @@ import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.util.*;
+import java.util.regex.Pattern;
 
 import javax.persistence.metamodel.ManagedType;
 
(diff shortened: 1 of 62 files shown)


entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/metamodel/AbstractAttribute.java
```diff
@@ -53,6 +54,9 @@
 public abstract class AbstractAttribute<X, Y> implements Attribute<X, Y> {
 
     private static final String[] EMPTY = new String[0];
+    private static final String THIS = "this";
+    private static final Pattern PLAIN_THIS_REPLACE_PATTERN = Pattern.compile("([^a-zA-Z0-9\\.])this[^\\.\\(\\[]");
```
Mobe91: unused

beikov:
Merged by https://github.com/Blazebit/blaze-persistence/commit/41d91ac577152ede0996470ff23e8bb0b8b6eb19

PR #404, opened 2017-03-25
Fixed #376 and fixed #356 - Flat views, flat view ids and function type determination

### LINKED ISSUE #356: Try to relax the requirements of @IdMapping
Right now, we always require an `@IdMapping` annotation within an entity view. Check if we can relax this requirement somehow and improve on some things
* Do not require a value to be provided (we can infer the mapping)
* Do not require the `@IdMapping` annotation explicitly if there is an attribute with the implicit mapping to the entities id attribute

We should also play around and see if it makes sense to have entity views that don't have a `@IdMapping`.
The only restrictions such an entity view would have are
* ~~Can't have collections because elements can't be correlated without ids~~
* ~~Can't be used in a collection mapping~~
* ~~Can't be updatable because they don't have an id~~
* ~~Can't be used with `PaginatedCriteriaBuilder` API because that requires an id~~

The only real limitation is that it can't have join fetched collections when no identity is available.

Update the getting started guide when removing the need for `@IdMapping`

Kind of related to https://github.com/Blazebit/blaze-persistence/issues/376

### LINKED ISSUE #376: Restructure entity view metamodel
Currently the entity view metamodel is kind of wrong and needs to be fixed. The structure of how it should be will be documented in the metamodel section of the documentation. We should adapt it to that design.
IdentifiableViewType should be removed and some methods from ManagedViewType have to be moved down to ViewType.

Actually we should also rename the EmbeddableViewType as it actually represents just a view that has no id. @Mobe91 have an idea for a name? I thought maybe _FlatViewType_ or so. By the way, the `@EmbeddableEntityView` annotation will be removed as it's functionality can be modeled with a normal `@EntityView` after relaxing id requirements with #356

beikov:
## Description

Restructured entity view metamodel implementation so that the real view types are actually known at the attribute level. This makes nested model checks easier to implement.
Implemented function type determination for being able to determine the distinct count of a name for a test.
Replaced embeddable entity views with the flat view concept which is essentially a normal view without an id.

## Related Issue

#376 and #356

## Motivation and Context

This is a clean split between id types(`ViewType`) and non-id types(`FlatViewType`). It allows to apply a flat view on a query which wasn't possible with embeddable view types before. Embeddables were a stupid concept for our case anyway as we don't have such a thing. Instead, we see this as _flat view_ which at the same time can also be a view root.
 but at the same time offers some convenience for mapping ids.


beikov:
Merged by https://github.com/Blazebit/blaze-persistence/commit/08cce8f90bccbbbcb3a284ca0b41dfe39df18b79

Issue #419, opened 2017-05-24
Adapt keyset predicate to be easier for DBMS to optimize


beikov:
Markus Winand explains this very nicely in his [article](http://use-the-index-luke.com/de/sql/partielle-ergebnisse/blaettern) that although the current form for keyset predicates that is implemented might be easier understandable, DBMS might not see through it that easily. Since DBMS can only use top level conjuncts for index lookups, we should rewrite the predicate generation to that form so we get more efficient execution plans.

Mobe91:
fixed via a48439d

Issue #123, opened 2015-04-26
Check if we can do something about polymorphic query problems


beikov:
There are some problems that we encountered when doing polymorphic property access in queries where the target property is only available in one of the possible types. This seems to only happen in hibernate when the base class is abstract.

http://stackoverflow.com/questions/18392252/access-subclass-properties-in-criteria-builder-jpa
http://stackoverflow.com/questions/6903783/jpa-criteria-api-query-property-of-subclass
http://stackoverflow.com/questions/23595594/exception-in-query-downcasting-to-subclass-with-treat-as
http://www.jroller.com/RickHigh/entry/round_2_jpa_hibernate_accessing

beikov:
Seems we will need secondary joins a.k.a. multiple from clauses in order to make that work in a spec compliant way.

beikov:
Let's call expressions that access properties from a subclass ```polymorphic expressions``` and joins with relations from a subclass ```polymorphic joins```.
When having a polymorphic join the transformation to make it JPA spec compliant requires a secondary join for the concrete type from which the relation comes from. The joins have to be rewritten to use the newly created join node from the secondary join. Additionally a where clause for each type has to be created that checks for id equality between the base entity and the concrete entity.
When a polymorphic join or expression is used in a clause it first has to be rewritten to use the newly created join node. The next step is to create a case when expression which emits the expression only when the type of the base entity is the concrete entity.

Here an example:

```
@Entity
@Inheritance(strategy = InheritanceType.SINGLE_TABLE)
@DiscriminatorColumn(name="OBJECT_TYPE", discriminatorType = DiscriminatorType.STRING)
public abstract class Document {
    
    @Id
    @GeneratedValue
    private Long id;
    private String name;
}

@Entity
@DiscriminatorValue("CONTRACT")
public class Contract extends Document {
    
    @ManyToOne(fetch = FetchType.LAZY)
    private Person contractOwner;
}

@Entity
@DiscriminatorValue("MEETING")
public class MeetingSheet extends Document {
    
    @ManyToOne(fetch = FetchType.LAZY)
    private Person writer;
}

@Entity
public class Person {
    
    @Id
    @GeneratedValue
    private Long id;
    private String name;
}
```

The query following polymorphic query will be supported.

```
SELECT CASE WHEN contractOwner IS NOT NULL THEN contractOwner.name ELSE writer.name END 
FROM Document d 
LEFT JOIN d.contractOwner contractOwner 
[...]

PR #431, opened 2017-08-19
[#123] Adapt keyset predicate to be easier for DBMS to optimize

Mobe91:
## Description
Implements the improvements suggested in #419

### DIFF
--- a/core/api/src/main/java/com/blazebit/persistence/spi/DbmsDialect.java
+++ b/core/api/src/main/java/com/blazebit/persistence/spi/DbmsDialect.java
@@ -28,6 +28,7 @@ import java.util.Map;
  * Interface for implementing some dbms specifics.
  *
  * @author Christian Beikov
+ * @author Moritz Becker
  * @since 1.1.0
  * @see CriteriaBuilderConfiguration#registerDialect(java.lang.String, com.blazebit.persistence.spi.DbmsDialect)
  */
@@ -203,6 +204,23 @@ public interface DbmsDialect {
      * @since 1.2.0
      */
     public boolean supportsJoinsInRecursiveCte();
+
+    /**
+     * Returns true if the dbms supports row value constructor syntax, false otherwise.
+     *
+     * @return Whether row value constructor syntax is supported by the dbms
+     * @since 1.2.0
+     */
+    public boolean supportsRowValueConstructor();
+
+    /**
+     * Returns true if the dbms supports all <, <=, >, >=, =, <> comparison operations for row values.
+     * Note that some DBMS only support = and <> operators.
+     *
+     * @return Whether full row value comparison operations are supported by the dbms
+     * @since 1.2.0
+     */
+    public boolean supportsFullRowValueComparison();
 
     /**
      * Returns the sql type for the java class type for usage in cast expressions.
--- a/core/impl/src/main/java/com/blazebit/persistence/impl/AbstractCommonQueryBuilder.java
+++ b/core/impl/src/main/java/com/blazebit/persistence/impl/AbstractCommonQueryBuilder.java
@@ -254,7 +254,7 @@ public abstract class AbstractCommonQueryBuilder<QueryResultType, BuilderType, S
 
         this.selectManager = new SelectManager<QueryResultType>(queryGenerator, parameterManager, this.joinManager, this.aliasManager, subqueryInitFactory, expressionFactory, jpaProvider, mainQuery, groupByExpressionGatheringVisitor, resultClazz);
         this.orderByManager = new OrderByManager(queryGenerator, parameterManager, subqueryInitFactory, this.aliasManager, jpaProvider, groupByExpressionGatheringVisitor);
-        this.keysetManager = new KeysetManager(queryGenerator, parameterManager);
+        this.keysetManager = new KeysetManager(queryGenerator, parameterManager, jpaProvider, dbmsDialect);
 
         final SizeTransformationVisitor sizeTransformationVisitor = new SizeTransformationVisitor(mainQuery, subqueryInitFactory, joinManager, jpaProvider);
         this.transformerGroups = Arrays.<ExpressionTransformerGroup<?>>asList(
@@ -1878,7 +1878,11 @@ public abstract class AbstractCommonQueryBuilder<QueryResultType, BuilderType, S
         } else {
             sbSelectFrom.append(" WHERE ");
 
-            keysetManager.buildKeysetPredicate(sbSelectFrom);
+            if (mainQuery.getQueryConfiguration().isOptimizedKeysetPredicateRenderingEnabled()) {
+                keysetManager.buildOptimizedKeysetPredicate(sbSelectFrom);
+            } else {
+                keysetManager.buildKeysetPredicate(sbSelectFrom);
+            }
 
             if (whereManager.hasPredicates()) {
                 sbSelectFrom.append(" AND ");
@@ -1892,7 +1896,7 @@ public abstract class AbstractCommonQueryBuilder<QueryResultType, BuilderType, S
         groupByManager.buildGroupByClauses(clauses);
 
         int size = transformerGroups.size();
-        for (int i = 0; i < size; i++) {
+        for (int i = 0; i < transformerGroups.size(); i++) {
             ExpressionTransformerGroup<?> transformerGroup = transformerGroups.get(i);
             clauses.addAll(transformerGroup.getRequiredGroupByClauses());
         }
@@ -1905,7 +1909,7 @@ public abstract class AbstractCommonQueryBuilder<QueryResultType, BuilderType, S
                 havingManager.buildGroupByClauses(clauses);
             }
             if (mainQuery.getQueryConfiguration().isImplicitGroupByFromOrderByEnabled()) {
-                orderByManager.buildGroupByClauses(clauses, false);
+                orderByManager.buildGroupByClauses(clauses);
             }
         }
 
--- a/core/impl/src/main/java/com/blazebit/persistence/impl/keyset/KeysetManager.java
+++ b/core/impl/src/main/java/com/blazebit/persistence/impl/keyset/KeysetManager.java
@@ -17,6 +17,9 @@
 package com.blazebit.persistence.impl.keyset;
 
 import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Iterator;
 import java.util.List;
 
 import com.blazebit.persistence.Keyset;
@@ -26,10 +29,14 @@ import com.blazebit.persistence.impl.ParameterManager;
 import com.blazebit.persistence.impl.ResolvingQueryGenerator;
 import com.blazebit.persistence.impl.SimpleQueryGenerator;
 import com.blazebit.persistence.impl.expression.Expression;
+import com.blazebit.persistence.impl.function.rowvalue.RowValueComparisonFunction;
+import com.blazebit.persistence.spi.DbmsDialect;
+import com.blazebit.persistence.spi.JpaProvider;
 
 /**
  *
  * @author Christian Beikov
+ * @author Moritz Becker
  * @since 1.0
  */
 public class KeysetManager extends AbstractKeysetBuilderEndedListener {
@@ -38,11 +45,15 @@ public class KeysetManager extends AbstractKeysetBuilderEndedListener {
 
     private final ResolvingQueryGenerator queryGenerator;
     private final ParameterManager parameterManager;
+    private final JpaProvider jpaProvider;
+    private final DbmsDialect dbmsDialect;
     private List<OrderByExpression> orderByExpressions;
 
-    public KeysetManager(ResolvingQueryGenerator queryGenerator, ParameterManager parameterManager) {
+    public KeysetManager(ResolvingQueryGenerator queryGenerator, ParameterManager parameterManager, JpaProvider jpaProvider, DbmsDialect dbmsDialect) {
         this.queryGenerator = queryGenerator;
         this.parameterManager = parameterManager;
+        this.jpaProvider = jpaProvider;
+        this.dbmsDialect = dbmsDialect;
     }
 
     public boolean hasKeyset() {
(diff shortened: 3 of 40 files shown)


core/impl/src/main/java/com/blazebit/persistence/impl/keyset/KeysetManager.java
```diff
@@ -57,15 +68,167 @@ public void initialize(List<OrderByExpression> orderByExpressions) {
...
+            if (orderByExpression.isNullable()) {
+                hasNullableOrderBys = true;
+                break;
+            }
+        }
+
+        extractedNonNullableExpression = orderByExpressions.get(0);
+
+        if (hasNullableOrderBys || !dbmsDialect.supportsRowValueConstructor() || !dbmsDialect.supportsFullRowValueComparison()) {
+            if (extractedNonNullableExpression.isNullable() && (!extractedNonNullableExpression.isNullFirst() || key[0] == null)) {
```
beikov: Could you please document the rationale behind this condition as a comment?

core/impl/src/main/java/com/blazebit/persistence/impl/keyset/KeysetManager.java
```diff
@@ -57,15 +68,167 @@ public void initialize(List<OrderByExpression> orderByExpressions) {
...
+                buildKeysetPredicate0(keysetMode, key, sb, orderByExpressions);
+            } else {
+                applyOptimizedKeysetNotNullItem(extractedNonNullableExpression, sb, 0, key[0], keysetMode, orderByExpressions.size() <= 1 && extractedNonNullableExpression.isUnique(), false);
```
beikov: Extract the condition for an expression being considered `unique` to a variable and document why `orderByExpressions.size() <= 1` is a requirement

core/api/src/main/java/com/blazebit/persistence/spi/DbmsDialect.java
```diff
@@ -205,6 +206,22 @@
...
     /**
+     * Returns true if the dbms supports row value constructor syntax, false otherwise.
+     *
+     * @return Whether row value constructor syntax is supported by the dbms
+     * @since 1.2.0
+     */
+    public boolean supportsRowValueConstructor();
+
+    /**
+     * Returns true if the dbms supports <, <=, >, >=, =, <> comparison operations for row values.
```
beikov: The documentation is misleading. If the DBMS supports the syntax it has to at least support equality checks i.e. operators `=` and `<>`. If you have no idea for what you might use the `supportsRowValueConstructor` I suggest you remove it and just keep this method.
Also note that there is a nice overview of the support in various DBMS here: https://blog.jooq.org/2012/12/26/row-value-expressions-and-the-between-predicate/

core/impl/src/main/java/com/blazebit/persistence/impl/function/rowvalue/RowValueConstructorComparisonFunction.java
```diff
@@ -0,0 +1,65 @@
...
+package com.blazebit.persistence.impl.function.rowvalue;
+
+import com.blazebit.persistence.spi.FunctionRenderContext;
+import com.blazebit.persistence.spi.JpqlFunction;
+
+/**
+ * @author Moritz Becker
+ * @since 1.2.0
+ */
+public class RowValueConstructorComparisonFunction implements JpqlFunction {
```
beikov: The function should be documented in `18_jpql_functions.adoc`

core/impl/src/main/java/com/blazebit/persistence/impl/keyset/KeysetManager.java
```diff
@@ -57,15 +68,167 @@ public void initialize(List<OrderByExpression> orderByExpressions) {
...
+    }
+
+    private void buildOptimizedPredicate0(KeysetMode keysetMode, Serializable[] key, StringBuilder sb, List<OrderByExpression> orderByExpressions) {
+        int expressionCount = orderByExpressions.size();
+        int brackets = 1;
+        sb.append('(');
+
+        SimpleQueryGenerator.BooleanLiteralRenderingContext oldBooleanLiteralRenderingContext = queryGenerator.setBooleanLiteralRenderingContext(SimpleQueryGenerator.BooleanLiteralRenderingContext.CASE_WHEN);
+
+        for (int i = 1; i < expressionCount; i++) {
```
beikov: Can you please document from a high level perspective what this loop is doing in a few sentences as comment?

core/testsuite/src/test/java/com/blazebit/persistence/testsuite/OptimizedKeysetPaginationRowValueConstructorTest.java
```diff
@@ -0,0 +1,414 @@
...
+        result = crit.page(result.getKeysetPage(), 0, 2).getResultList();
+
+        assertEquals(2, result.getSize());
+        assertEquals("doc4", result.get(0).get(0));
+        assertEquals("doc3", result.get(1).get(0));
+    }
+
+    @Test
+    public void forwardBackwardsPaginationResultSetOrder() {
+        final Document doc5 = new Document("doc5");
```
beikov: I suppose the MySQL issues(https://travis-ci.org/Blazebit/blaze-persistence/jobs/266267687) might be related to the fact that you are using `transactional` to persist new entities after you started the TX for your test. To fix this I recommend following the suggestion from above i.e. using the same data set throughout a single test class.

core/impl/src/main/java/com/blazebit/persistence/impl/AbstractCommonQueryBuilder.java
```diff
@@ -1891,9 +1895,7 @@ protected void appendGroupByClause(StringBuilder sbSelectFrom) {
         Set<String> clauses = new LinkedHashSet<String>();
         groupByManager.buildGroupByClauses(clauses);
 
-        int size = transformerGroups.size();
```
Mobe91: * I consider this micro-optimization. As long as the performance of the loops is not proven to be a problem, I would go for readability vs micro-optimization.
* Apart from that I think that the for-each loop syntax considerably enhances readability as it removes 1-2 lines of nonsense code at every use-side.
* In this concrete instance I also don't understand why you store the `size` in a separate variable.
* Also, `transformerGroups` is of type `java.util.List`. So if, for some reason, say a `LinkedList` is used in the future instead of an `ArrayList`, the performance of the classical for-loop will suddenly suck (due to the lack of constant time random access) which might not be obvious to and may be overlooked by whoever introduces such a change.

beikov: It is a micro optimization, but why not help the compiler/interpreter and GC when possible? I generally agree with you that if the implementation isn't known that an enhanced-for loop is a better approach, but I don't consider an indexed loop to be unreadable. The implementation is controlled by us and such optimizations are useful in general. I like to avoid creating garbage i.e. the iterator object whenever possible.

core/testsuite/src/test/java/com/blazebit/persistence/testsuite/KeysetPaginationTest.java
```diff
@@ -39,14 +41,15 @@
...
+ * @author Moritz Becker
  * @since 1.0
  */
 public class KeysetPaginationTest extends AbstractCoreTest {
 
     private Person o4;
 
-    @Override
-    public void setUpOnce() {
+    @Before
```
Mobe91: In this case I will need to extract the `keysetPaginationPage1EdgeCaseFallback` into a separate test class that extends the abstract keyset pagination test class since it needs to be able to modify the database contents. Do you comply?

beikov: Either you do that or you try to construct the scenario with the existing data.

core/impl/src/main/java/com/blazebit/persistence/impl/keyset/KeysetManager.java
Mobe91: Would have been nice to have such a comment for the loop starting on line 243 :)

core/api/src/main/java/com/blazebit/persistence/spi/DbmsDialect.java
Mobe91: How is this misleading? The description states that this property is true if all of the presented operators are supported. Cannot be expressed clearer IMHO. The description implies that the sole support of <>, = operators leads to the property being false.
Since the support for row value constructor is independent from the support for row value comparison, I figured that it makes sense to introduce separate properties. But you are right, the `supportsFullRowValueComparison` would suffice at the moment.

beikov: I just meant that the documentation suggests `=`and `<>` are only supported if this method returns true. Anyways, I think you got the gist.

core/parser/src/main/java/com/blazebit/persistence/impl/SimpleQueryGenerator.java
```diff
@@ -535,7 +535,7 @@ public void visit(FunctionExpression expression) {
         BooleanLiteralRenderingContext oldBooleanLiteralRenderingContext = setBooleanLiteralRenderingContext(BooleanLiteralRenderingContext.PLAIN);
         ParameterRenderingMode oldParameterRenderingMode = setParameterRenderingMode(ParameterRenderingMode.PLACEHOLDER);
         boolean hasExpressions = expression.getExpressions().size() > 0;
-        String functionName = expression.getFunctionName();
+        String functionName = expression.getFunctionName().toUpperCase();
```
Mobe91: You are right, this is a problem. Your suggested solution sounds good but the `SimpleQueryGenerator` is used in a number of places so the case sensitivity problem would remain for these use sites...

beikov: The `ResolvingQueryGenerator` is the _last bastion_ after which a query gets passed to a JPA provider so it is sufficient to handle it only there. Everything else is just intermediate processing.

core/impl/src/main/java/com/blazebit/persistence/impl/keyset/KeysetManager.java
```diff
@@ -57,15 +68,175 @@ public void initialize(List<OrderByExpression> orderByExpressions) {
...
+        }
+
+        extractedNonNullableExpression = orderByExpressions.get(0);
+
+        // We can only use row value based keyset predicates if the dbms supports row values and row value comparison and
+        // if all order bys are non-nullable because null elements would break the row value comparison.
+        if (hasNullableOrderBys || !dbmsDialect.supportsRowValueConstructor() || !dbmsDialect.supportsFullRowValueComparison()) {
+            // if the first order by is nullable, we cannot render an optimized form because we would need to include
+            // null handling involving disjunction on the top predicate level
+            if (extractedNonNullableExpression.isNullable() && (!extractedNonNullableExpression.isNullFirst() || key[0] == null)) {
```
beikov: I still don't understand the condition on this line. If `key[0]` is != null, this would render the optimized form even if the expression is nullable, given that null first is configured. I don't understand what the null precedence and the actual value has to do with the actual rendering.

core/impl/src/main/java/com/blazebit/persistence/impl/keyset/KeysetManager.java
```diff
@@ -138,11 +309,63 @@ public void buildKeysetPredicate(StringBuilder sb) {
         queryGenerator.setBooleanLiteralRenderingContext(oldBooleanLiteralRenderingContext);
     }
 
-    private void applyKeysetNotNullableItem(OrderByExpression orderByExpr, StringBuilder sb, Expression expr, int i, Serializable[] key, KeysetMode keysetMode) {
+    private void applyOptimizedKeysetNotNullItem(OrderByExpression orderByExpr, StringBuilder sb, int i, Serializable keyElement, KeysetMode keysetMode, boolean negated, boolean lastOrderBy) {
```
beikov: `lastOrderBy` is unused in that method.

core/impl/src/main/java/com/blazebit/persistence/impl/keyset/KeysetManager.java
```diff
@@ -57,15 +68,175 @@ public void initialize(List<OrderByExpression> orderByExpressions) {
...
+                hasNullableOrderBys = true;
+                break;
+            }
+        }
+
+        extractedNonNullableExpression = orderByExpressions.get(0);
+
+        // We can only use row value based keyset predicates if the dbms supports row values and row value comparison and
+        // if all order bys are non-nullable because null elements would break the row value comparison.
+        if (hasNullableOrderBys || !dbmsDialect.supportsRowValueConstructor() || !dbmsDialect.supportsFullRowValueComparison()) {
```
beikov: Drop the condition `!dbmsDialect.supportsRowValueConstructor()`

core/testsuite/src/test/java/com/blazebit/persistence/testsuite/OptimizedKeysetPaginationRowValueConstructorTest.java
beikov: Move that to a separate test

core/testsuite/src/test/java/com/blazebit/persistence/testsuite/OptimizedKeysetPaginationRowValueConstructorTest.java
```diff
@@ -0,0 +1,406 @@
...
+
+        keyset = result.getKeysetPage();
+        crit = crit.page(keyset, 1, 1);
+        result = crit.getResultList();
+        assertEquals(1, result.size());
+        assertEquals("doc2 - Karl2", result.get(0));
+    }
+
+    @Test
+    public void keysetPaginationPage1EdgeCaseFallback() {
```
beikov: Move to a separate test

core/testsuite/src/test/java/com/blazebit/persistence/testsuite/OptimizedKeysetPaginationRowValueConstructorTest.java
```diff
@@ -178,6 +175,12 @@ public void forwardBackwardsPaginationResultSetOrder() {
                         + " ORDER BY " + renderNullPrecedence("owner_1.name", "DESC", "LAST") + ", " + renderNullPrecedence("d.name", "DESC", "LAST") + ", " + renderNullPrecedence("d.id", "ASC", "LAST"),
                 pcb.getPageIdQueryString()
         );
+
+        List<Document> docs = em.createQuery("SELECT d FROM Document d INNER JOIN d.owner o ORDER BY o.name DESC NULLS LAST, d.name DESC NULLS LAST, d.id ASC", Document.class).getResultList();
+        for (Document doc : docs) {
+            System.out.println(String.format("%s | %s | %d", doc.getOwner().getName(), doc.getName(), doc.getId()));
```
beikov: Please no system out printing in tests

core/testsuite/src/main/java/com/blazebit/persistence/testsuite/AbstractCoreTest.java
```diff
@@ -76,24 +77,7 @@ protected CriteriaBuilderConfiguration configure(CriteriaBuilderConfiguration co
     }
     
     protected Set<String> getRegisteredFunctions() {
```
beikov: Is there any need for exposing this anymore? I think the method can be removed since it shouldn't be overridden anymore.

core/testsuite/src/main/java/com/blazebit/persistence/testsuite/AbstractCoreTest.java
```diff
@@ -368,6 +348,10 @@ protected String singleValuedAssociationIdNullnessPath(String path, String id) {
             return path;
         }
     }
+
+    protected String resolveRegisteredFunctionName(String functionName) {
+        return getIgnoreCase(getRegisteredFunctions(), functionName);
```
beikov: Do a map lookup in the registered functions like `STATIC_EMF_INTEGRATOR.getRegisteredFunctions(emf).get(functionName.toLowerCase())` instead.

Mobe91: This won't work since it's not a case insensitive lookup.

Mobe91: I could do this:
```java
getIgnoreCase(STATIC_EMF_INTEGRATOR.getRegisteredFunctions(emf).keySet(), functionName)
```
or alternatively, I need to create a separate case insensitive map from `STATIC_EMF_INTEGRATOR.getRegisteredFunctions(emf)`.

beikov: Sounds ok, just make sure you drop the `getRegisteredFunctions` from the test

PR #465, opened 2017-11-03
Test for C > B > A EntityViewInheritance

JWGmeligMeyling:
This fails that C is invalidly mapped to an AView and not a CView.

The test cases describe three different scenario's, not all of which I am certain whether this is a feature Blaze-persistence should support. Nevertheless I think that at least two out of the three scenario's should be supported.


JWGmeligMeyling:
Rebased. This issue not necessairily describes a problem with entity viewinhertiance, but rather a limitation. Currently it behaves slightly different to Hibernate / JPA Inheritance. Especially in the case where a view C > B > A exists and an entity should be mapped as a C but is mapped as a B.

beikov:
Same here with `MultiLevelViewInheritance.java`

JWGmeligMeyling:
I've added the license header.

beikov:
Kind of related to #471. I'm still debugging and trying to figure this out.

beikov:
Superseded by #478

Issue #420, opened 2017-05-30
Implicit joins coming from subqueries are wrongly removed in paginated query


beikov:
The issue was with an entity view where a subquery provider had an expression like `OUTER(positions.id.headId)` which involves a join on the relation `positions` that is only used in the subquery. Unfortunately the clause dependency analysis doesn't pick up correlated join nodes so the join is removed from the object query of a PaginationCriteriaBuilder. The workaround is to create a tautological predicate in the where clause of the outer query that simply references the join nodes, so that a clause dependency is registered.

PR #489, opened 2017-12-04
[#420] - Fix wrong query generation for cases when a subquery is the reason for joins on outer levels

beikov:
## Description

The issue is quite simple, when a subquery contains an expression like shown in the test `OUTER(contacts.id)`, which involves a relation, the paginated query is wrong. The problem is that the join node `contacts` does not have the clause dependency on the parent queries clause. I fixed this by doing the implicit join with the parent JoinVisitor.

## Related Issue

#420


core/impl/src/main/java/com/blazebit/persistence/impl/JoinVisitor.java
```diff
@@ -74,6 +76,24 @@ private void visit(PathExpression expression, boolean idRemovable) {
...
+                JoinNode baseNode = (JoinNode) expression.getBaseNode();
+                AliasManager aliasOwner = baseNode.getAliasInfo().getAliasOwner();
+                if (aliasOwner != joinManager.getAliasManager()) {
+                    parentVisitor.addClauseDependencies(baseNode, aliasOwner);
+                }
+            }
+        }
+    }
+
+    void addClauseDependencies(JoinNode node, AliasManager aliasOwner) {
```
Mobe91: Why is this method not private?

beikov: Otherwise I couldn't call the method on a different instance.

Mobe91: You are only calling it on a different instsance of the same type, so it can be private.

beikov:
Merged d4d52b83f0e2a3b839b427be35456456b87ee471

Issue #521, opened 2018-03-01
EntityMetamodel generation fails when SecondaryTable is used with an inheritance strategy


JWGmeligMeyling:
The underlying problem is that we check the subclass tables and not the regular tables of entity persister:

```
entityPersister.getSubclassTableSpan() == 5
entityPersister.getTableSpan() == 3
```

With the code being used:

```java
tables = new Table[((JoinedSubclassEntityPersister) entityPersister).getTableSpan()];
for (int i = 0; i < tables.length; i++) {
      tables[i] = database.getTable(entityPersister.getSubclassTableName(i));
}
```

Stacktrace:

```
17:32:27,504 ERROR [org.jboss.msc.service.fail] (ServerService Thread Pool -- 26) MSC000001: Failed to start service jboss.deployment.unit."iris-server.war".component.CriteriaBuilderFactoryProducer.START: org.jboss.msc.service.StartException in service jboss.deployment.unit."iris-server.war".component.CriteriaBuilderFactoryProducer.START: java.lang.IllegalStateException: WFLYEE0042: Failed to construct component instance
	at org.jboss.as.ee.component.ComponentStartService$1.run(ComponentStartService.java:57)
	at java.util.concurrent.Executors$RunnableAdapter.call$$$capture(Executors.java:511)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java)
	at java.util.concurrent.FutureTask.run$$$capture(FutureTask.java:266)
	at java.util.concurrent.FutureTask.run(FutureTask.java)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:748)
	at org.jboss.threads.JBossThread.run(JBossThread.java:320)
Caused by: java.lang.IllegalStateException: WFLYEE0042: Failed to construct component instance
	at org.jboss.as.ee.component.BasicComponent.constructComponentInstance(BasicComponent.java:163)
	at org.jboss.as.ee.component.BasicComponent.constructComponentInstance(BasicComponent.java:134)
	at org.jboss.as.ee.component.BasicComponent.createInstance(BasicComponent.java:88)
	at org.jboss.as.ejb3.component.singleton.SingletonComponent.getComponentInstance(SingletonComponent.java:124)
[...]

PR #522, opened 2018-03-01
Reproducer for issue #521 (issue not present since Alpha4)

JWGmeligMeyling:

### DIFF
--- a/core/testsuite/src/main/java/com/blazebit/persistence/testsuite/entity/SecondaryTableEntityBase.java
+++ b/core/testsuite/src/main/java/com/blazebit/persistence/testsuite/entity/SecondaryTableEntityBase.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2014 - 2018 Blazebit.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


beikov:
For this test to work properly with DataNucleus, the entity classes need to be top level classes in `src/main/java` that are registered in the `src/main/resources/META-INF/orm.xml`.
I suggest you name the entities something like `SecondaryTableEntityBase` and `SecondaryTableEntitySub` and put them into the package where the other entities are.

JWGmeligMeyling:
Allright! Didn't want to do that directly, because it would potentially cause all tests to fail rather than just the reproducer, but it should just pass now anyway.

beikov:
Yeah it passed for Hibernate 5.2 and EclipseLink already, so it's probably fine.

JWGmeligMeyling:
I've moved the files to `src/main/java`. It however doesn't seem to pass locally, I think I missed something in `persistence.xml` but I can't seem to find it.

beikov:
Looks fine to me. Let's see what Travis CI reports.

beikov:
Ahh, the test class needs to specify the entity classes it uses by providing them via an overridden `getEntityClasses`. These classes are registered in the persistence unit. The orm.xml is just for enhancers to know which classes to enhance.

JWGmeligMeyling:
Fixed 😄 Thought it wouldn't be necessary if the entities were already listed in the test `persistence.xml`.

JWGmeligMeyling:
Is the test suite failing in general or did I miss something?

beikov:
It's a checkstyle error because of the use of a `.*` import: https://travis-ci.org/Blazebit/blaze-persistence/jobs/348301799

JWGmeligMeyling:
Annoying IDE configs all the time! :)

Issue #519, opened 2018-02-24
Hibernate Subselect Entity not supported by EntityMetamodelImpl


JWGmeligMeyling:
### Description
After playing a bit with table functions (see also #181), I had the idea to experiment with mapping a `generate_series` query to an entity using the [`@Subselect`](https://docs.jboss.org/hibernate/orm/5.2/javadocs/org/hibernate/annotations/Subselect.html) annotation. Find below the produced stack trace.

A trivial work around is to wrap the query into to a view and map that instead. I am uncertain whether a fix for this is worthwhile, but it could perhaps be a step towards implementing table functions (#181).

```java
package com.pallasathenagroup.entities.series;

import lombok.Data;
import lombok.EqualsAndHashCode;
import org.hibernate.annotations.Immutable;
import org.hibernate.annotations.Subselect;

import javax.persistence.Column;
import javax.persistence.Entity;
import javax.persistence.Id;
import java.time.LocalDateTime;

@Data
@Entity
@Immutable
@Subselect("select start_date_time, start_date_time + interval '1 year' as end_date_time " +
    "from generate_series('2016-09-01'::timestamp, date_trunc('year', now()) + interval '8 months', '1 year') as start_date_time")
@EqualsAndHashCode(of = {"start"})
public class CollegeYearSeries implements Comparable<CollegeYearSeries> {

    @Id
    @Column(name = "start_date_time")
    private LocalDateTime start;

    @Column(name = "end_date_time")
    private LocalDateTime end;

    @Override
    public int compareTo(CollegeYearSeries o) {
        return getStart().compareTo(o.getStart());
    }
}
```

```
Caused by: java.lang.NullPointerException
	at com.blazebit.persistence.impl.hibernate.HibernateExtendedQuerySupport.getColumnTypes(HibernateExtendedQuerySupport.java:246)
	at com.blazebit.persistence.impl.EntityMetamodelImpl.<init>(EntityMetamodelImpl.java:110)
	at com.blazebit.persistence.impl.CriteriaBuilderFactoryImpl.<init>(CriteriaBuilderFactoryImpl.java:81)
	at com.blazebit.persistence.impl.CriteriaBuilderConfigurationImpl.createCriteriaBuilderFactory(CriteriaBuilderConfigurationImpl.java:599)
	at com.pallasathenagroup.iris.producers.CriteriaBuilderFactoryProducer.init(CriteriaBuilderFactoryProducer.java:32)
[...]

PR #523, opened 2018-03-01
Reproducer issue #519: Hibernate Subselect Entity not supported by EntityMetamodelImpl

JWGmeligMeyling:

### DIFF
--- a/core/testsuite/src/test/hibernate/com/blazebit/persistence/testsuite/Issue519Test.java
+++ b/core/testsuite/src/test/hibernate/com/blazebit/persistence/testsuite/Issue519Test.java
@@ -1,0 +1,117 @@
+/*
+ * Copyright 2014 - 2018 Blazebit.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


JWGmeligMeyling:
The test fails because `entityPersister.getSubclassTableName(i)` returns `"( SELECT 5 as b, a.a_id FROM a )"`, and looking up the `Table` for this in `Database` returns `null`.  I think this is actually quite hard to resolve as it seems that Hibernate provides no way do determine the column types for this sub select result.

JWGmeligMeyling:
I've changed the test because the Hibernate was invalid causing the reproducer to fail even with the bug fixed.

JWGmeligMeyling:
The tests fail because the SQL query used in the test entity is not valid SQL for all database vendors. I think for example for MySQL, columns have to be wrapped between ` quotes. Perhaps we can come up with a SQL fragment that is valid across all database vendors. Otherwise we may have to restrict the test to be run only on a couple of databases instead.

beikov:
Superseded by https://github.com/Blazebit/blaze-persistence/pull/525

Issue #484, opened 2017-11-26
Implicit joins for aggregates do not work


JWGmeligMeyling:
### Description

As discussed on the Slack channel:

```java
return criteriaBuilderFactory.create(entityManager, Tuple.class)
    .from(EmployeePayment.class)
    .select("FUNCTION('YEAR', created)", "creationYear")
    .select("FUNCTION('MONTH', created)", "creationMonth")
    .select("FUNCTION('YEAR', computedOver)", "computedOverYear")
    .select("FUNCTION('MONTH', computedOver)", "computedOverMonth")
    .select("type", "paymentType")
    .select("SUM(amount)", "amount")
    .orderBy("creationYear", true, false)
    .orderBy("creationMonth", true, false)
    .orderBy("computedOverYear", true, false)
    .orderBy("computedOverMonth", true, false)
    .orderBy("paymentType", true, false)
    .getResultList()
```

Fails with:

```
java.lang.RuntimeException: The select expression 'SUM(amount)' is not a simple path expression! No idea how to implicit join that.

    at com.blazebit.persistence.impl.JoinManager.implicitJoin(JoinManager.java:1538)
    at com.blazebit.persistence.impl.JoinManager.implicitJoin(JoinManager.java:1389)
    at com.blazebit.persistence.impl.JoinVisitor.visit(JoinVisitor.java:76)
    at com.blazebit.persistence.impl.JoinVisitor.visit(JoinVisitor.java:68)
    at com.blazebit.persistence.impl.expression.PathExpression.accept(PathExpression.java:75)
    at com.blazebit.persistence.impl.expression.VisitorAdapter.visit(VisitorAdapter.java:105)
    at com.blazebit.persistence.impl.JoinVisitor.visit(JoinVisitor.java:97)
    at com.blazebit.persistence.impl.expression.AggregateExpression.accept(AggregateExpression.java:71)
    at com.blazebit.persistence.impl.SelectManager.acceptVisitor(SelectManager.java:174)
    at com.blazebit.persistence.impl.AbstractCommonQueryBuilder.applyImplicitJoins(AbstractCommonQueryBuilder.java:1401)
    at com.blazebit.persistence.impl.AbstractCommonQueryBuilder.prepareAndCheck(AbstractCommonQueryBuilder.java:1811)
    at com.blazebit.persistence.impl.AbstractCommonQueryBuilder.getBaseQueryStringWithCheck(AbstractCommonQueryBuilder.java:1498)
[...]

PR #529, opened 2018-03-10
[#484] fixed use of select alias in its defining expression

Mobe91:


core/impl/src/main/java/com/blazebit/persistence/impl/SelectManager.java
```diff
@@ -169,19 +169,19 @@ public boolean containsSizeSelect() {
     }
 
     void acceptVisitor(Visitor v) {
-        List<SelectInfo> infos = selectInfos;
-        int size = selectInfos.size();
-        for (int i = 0; i < size; i++) {
-            final SelectInfo selectInfo = infos.get(i);
+        for (SelectInfo selectInfo : selectInfos) {
```
beikov: The old-school for loops are done in many places on purpose as it takes some time until an enhanced for loop is compiled down to the optimal code. Using the index based loop is generally better during the interpretation phase. Please revert back to the old style in this class.

Mobe91: Are you serious?? This code looks like crap when I revert back.

beikov: I am serious ;P

Mobe91: When I look at code of other projects, they tend do not trash their own code for the sake of probably unnoticable performance gains that are neutralized over time anyway.
So I don't like this but I think discussing it further makes no sense, so I'll just revert...

Mobe91: So the original version is like this:
```java
List<SelectInfo> infos = selectInfos;
int size = selectInfos.size();
for (int i = 0; i < size; i++) {
	final SelectInfo selectInfo = infos.get(i);
	selectInfo.getExpression().accept(v);
}
```
Can I at least write it like this, or am I overlooking another "performance optimization":
```java
for (int i = 0; i < selectInfos.size(); i++) {
	final SelectInfo selectInfo = selectInfos.get(i);
	selectInfo.getExpression().accept(v);
}
```

beikov: Should be ok this way ;)
But note that there are compile errors: https://travis-ci.org/Blazebit/blaze-persistence/jobs/351678451

Mobe91: Hmm didn't test it again after rebasing.. guess the rebase must have destroyed it :/

PR #541, opened 2018-03-21
[#255] Allow to omit the count query

beikov:
## Related Issue

#255

PR #543, opened 2018-03-28
Changed MacroConfiguration comparator

JWGmeligMeyling:
I have a hard time believing that it is intentional for one macro with the same string length to remove another macro from the macro map, so I think a case insensitive comparator is more appropriate.

> a sorted map performs all key comparisons using its compareTo (or compare) method, so two keys that are deemed equal by this method are, from the standpoint of the sorted map, equal.


beikov:
Superseded by https://github.com/Blazebit/blaze-persistence/pull/541

PR #559, opened 2018-05-01
[#513] fixes #513, prevent literal group bys, fixed test logging

### LINKED ISSUE #513: SIZE(collection) expression has some issues when used within a CTE query
### Description
I've observed that using JPQL expression that create aggregations under the hood do not work in (at least) CTE queries. So one can not do: ```with(SomeCTE.class).from(Entity.class, "ent").bind('attribute').select("SIZE(ent.someList)")```. It will fail reporting it misses a grouping column. I think that this usage should be supported, as it seems a quite regular JPQL expression. I am not sure where exactly it breaks. Perhaps we do not detect that grouping is necessary for this query? 

Another, more complex, similar issue  occurs when such an expression in your own grouping query, such as ```with(SomeCTE.class).from(Entity.class, "ent").bind('attribute').select("SUM(SIZE(ent.someList))").groupBy("ent.someGroupKey")```.  I am uncertain whether this last case should ever be allowed because you use a nested aggregation function. 

Both cases can be easily worked around by writing your own join query / count aggregate. In my case ```with(SomeCTE.class).from(Entity.class, "ent").innerJoinDefault("ent.someList", "b").bind('attribute').select("COUNT(b.id)")```. 

### Environment

Version:            Alpha3
JPA-Provider:       Hibernate 5.2.12
DBMS:               Postgresql
Application Server:

Mobe91:


core/impl/src/main/java/com/blazebit/persistence/impl/GroupByExpressionGatheringVisitor.java
```diff
@@ -91,6 +92,9 @@ private boolean setCollect(boolean collect) {
     }
 
     public Set<Expression> extractGroupByExpressions(Expression expression) {
+        if (expression instanceof LiteralExpression) {
```
beikov: Please add a comment here stating that only plain literals don't make sense.

core/parser/src/main/java/com/blazebit/persistence/parser/expression/CombiningResultVisitorAdapter.java
```diff
@@ -0,0 +1,298 @@
...
+import com.blazebit.persistence.parser.predicate.MemberOfPredicate;
+import com.blazebit.persistence.parser.predicate.Predicate;
+
+import java.util.List;
+
+/**
+ * @author Moritz Becker
+ * @since 1.2.0
+ */
+public abstract class CombiningResultVisitorAdapter<T> implements Expression.ResultVisitor<T> {
```
beikov: Although the idea of a combining visitor is interesting, it is IMO inappropriate for the size detection as the detection can stop traversing the expression tree as soon as a size expression was found, whereas the combining visitor will still traverse the whole tree. Please stick with the AbortableVisitorAdapter.

core/testsuite/src/test/java/com/blazebit/persistence/testsuite/SizeTransformationTest.java
```diff
@@ -356,4 +359,40 @@ public void testSizeTransformationMultiLevelCorrelatedFromClause() {
             cb.getResultList();
     }
 
+    // from issue #513
+    @Test
```
beikov: Some DBs seem to have a problem with that expression:

* DB2 - https://travis-ci.org/Blazebit/blaze-persistence/jobs/374092236#L1515
* SQL Server - https://travis-ci.org/Blazebit/blaze-persistence/jobs/374092224#L1537

We need to investigate if there is a way to make it work, otherwise we will ignore the test for the problematic DBMS

core/parser/src/main/java/com/blazebit/persistence/parser/expression/CombiningResultVisitorAdapter.java
```diff
@@ -0,0 +1,298 @@
...
```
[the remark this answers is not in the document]
Mobe91: It might prove useful in the future

core/impl/src/main/java/com/blazebit/persistence/impl/transform/SizeTransformationVisitor.java
```diff
@@ -359,6 +360,7 @@ private Expression wrapSubqueryConditionally(SubqueryExpression subquery, boolea
             // see https://hibernate.atlassian.net/browse/HHH-9331
             List<Expression> coalesceArguments = new ArrayList<>(1);
             coalesceArguments.add(subquery);
+            coalesceArguments.add(new NullExpression());
```
beikov: Seems DataNucleus doesn't like this. Please either report the problem to them and add an issue key to `@Category(NoDatanucleus.class)` or fix it somehow. Here is the error: https://travis-ci.org/Blazebit/blaze-persistence/jobs/375353382#L1572

core/testsuite/src/test/java/com/blazebit/persistence/testsuite/SizeTransformationTest.java
```diff
@@ -391,7 +392,7 @@ public void work(EntityManager em) {
                 .from(Document.class, "d")
                 .select("SUM(SIZE(d.people))");
 
-        String expected = "SELECT SUM(COALESCE((SELECT COUNT(*) FROM d.people person))) FROM Document d";
+        String expected = "SELECT SUM(COALESCE((SELECT COUNT(*) FROM d.people person),NULL)) FROM Document d";
         assertEquals(expected, cb.getQueryString());
         assertEquals(Long.valueOf(4L), cb.getResultList().get(0));
```
beikov: Seems the result of a `SUM` is BigDecimal when using EclipseLink: https://travis-ci.org/Blazebit/blaze-persistence/jobs/375353385#L1478
I guess this is legal, you should use `Number` as the return type and do a proper conversion by using `toLong()`

Issue #545, opened 2018-03-29
SQL Type determination for Subselect and Formula fails on custom types


JWGmeligMeyling:
If one uses a custom type (`@Type`) in a `@Subselect` or `@Formula` entity, metamodel generation fails. This is because for custom types as the `dictatedSizes()` method returns `Object[] { null }`. Therefore, the workaround proposed in #519 does not work here.

Only seen in Hibernate, not sure how custom types affect other ORMs as well.

(Also reported via Slack)

### Environment

Version:            
JPA-Provider:       
DBMS:               
Application Server:

PR #585, opened 2018-06-28
[#545] Fallback to default Size if not Sized

JWGmeligMeyling:

### DIFF
--- a/core/testsuite/src/test/hibernate/com/blazebit/persistence/testsuite/Issue545Test.java
+++ b/core/testsuite/src/test/hibernate/com/blazebit/persistence/testsuite/Issue545Test.java
@@ -1,0 +1,91 @@
+/*
+ * Copyright 2014 - 2018 Blazebit.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


core/testsuite/src/test/hibernate/com/blazebit/persistence/testsuite/Issue545Test.java
```diff
@@ -0,0 +1,92 @@
...
+                entityWithCustomType.customValue = 'c';
+                em.persist(entityWithCustomType);
+                em.flush();
+                em.clear();
+            }
+        });
+    }
+
+    @Test
+    @Category(NoOracle.class)
```
JWGmeligMeyling: Oracle failed because the identifier was too long. Probably because Hibernate changes the identifier to `_formula_x_x` ...

beikov: Can we do anything about the identifier length i.e. annotate something?

beikov: The issue might be that you don't define a table name explicitly for the entity.

Issue #556, opened 2018-04-23
Investigate correlation mapping issues with using an embedded id as correlation basis


beikov:
Using a correlation mapping with a correlation basis that resolves to an embedded id type does not work properly. The error I saw was something like `The correlation basis 'id' in the context of the managed type 'com.example.SomeEntity' resolved to the non-identifiable type 'com.example.MyId'!`
See if this can be fixed or at least improve the error message because I think using `THIS` as correlation basis is an alternative.

PR #586, opened 2018-06-28
[#556] Test and fix for the use of a correlation basis that has an embeddable type

beikov:
## Related Issue

#556


beikov:
Merged by 4a5787e24563463278eab389f8de74865051d2a8

PR #589, opened 2018-07-03
[#574] Test and fix for handling of quoted identifiers

### LINKED ISSUE #574: Quoted identifiers mess up alias of VALUES clause
Using the VALUES clause with Hibernate when quoted identifiers are activated currently produces broken SQL because the alias for the value column contains the table name.

```
select valuesenti0_."value" as col_0_0_ from (VALUES (cast(? as bigint))) valuesenti0_(valuesenti0_."value")
```

I guess the extraction of the column name can't handle the quotes.

beikov:
## Related Issue

#574


beikov:
Merged by d3230aa34be690d97aff6977cdf22efb8c9964a7

Issue #557, opened 2018-04-23
Investigate non-id attribute access in VIEW_ROOT problem when using SUBSELECT fetch strategy


beikov:
The problem is with an entity view with a correlation mapping that uses the SUBSELECT fetch strategy. The correlation provider uses an expression like `VIEW_ROOT(id.someProperty)` which seems to be problematic. We should add a few tests that use embedded ids to make sure this works properly.

PR #593, opened 2018-07-04
[#557] Add test and fix for various uses the view root macro

beikov:
## Related Issue

#557


beikov:
Merged by 1c834ca504dffac8eb52b955d081bb025dddf4e0

PR #594, opened 2018-07-04
Fix various small issues that were introduced while working on issues for this version

beikov:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -58,6 +58,7 @@
         <module>examples</module>
         <module>checkstyle-rules</module>
         <module>archetype</module>
+        <module>dist</module>
     </modules>
 
     <issueManagement>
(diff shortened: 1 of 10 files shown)


pom.xml
```diff
@@ -224,32 +225,6 @@
                 </plugins>
             </build>
         </profile>
-        <profile>
-            <id>blazebit-release</id>
-            <properties>
-                <!-- We can only release with a JDK 9 since we build MR JARs -->
-                <jdk.min.version>1.9</jdk.min.version>
```
JWGmeligMeyling: Don't we require this still? Or do you take care of that locally anyway?

JWGmeligMeyling: Nevermind, I just notice its just moved.

PR #599, opened 2018-07-11
[#325] Add support for EntityViewSetting parameters

### LINKED ISSUE #325: Allow providing custom EntityViewSetting to spring data repositories
Right now, the repositories are bound to a specific entity view type that is defined statically.
Since one of the nice things about entity views is, that the type, filters and sortings can be dynamically defined by the client, we should also allow passing in the entire `EntityViewSetting` object or maybe even just subsets of it into a repository via parameters.

heruan:
## Description

- Add `OptionalParam` annotation to denote parameters to be bound via `EntityViewSetting::addOptionalParameter`
- ~~Add also support for `Specification` parameters~~

## Related Issue
Partly fixes #325.

## Motivation and Context
Provides ability to customize `EntityViewSetting` via Spring repository interfaces.

### DIFF
--- a/integration/spring-data/base/src/main/java/com/blazebit/persistence/spring/data/base/query/AbstractPartTreeBlazePersistenceQuery.java
+++ b/integration/spring-data/base/src/main/java/com/blazebit/persistence/spring/data/base/query/AbstractPartTreeBlazePersistenceQuery.java
@@ -20,10 +20,12 @@ package com.blazebit.persistence.spring.data.base.query;
 import com.blazebit.persistence.CriteriaBuilderFactory;
 import com.blazebit.persistence.PagedList;
 import com.blazebit.persistence.PaginatedCriteriaBuilder;
-import com.blazebit.persistence.criteria.BlazeCriteriaQuery;
 import com.blazebit.persistence.criteria.BlazeCriteria;
+import com.blazebit.persistence.criteria.BlazeCriteriaQuery;
+import com.blazebit.persistence.spring.data.base.query.JpaParameters.JpaParameter;
 import com.blazebit.persistence.view.EntityViewManager;
 import com.blazebit.persistence.view.EntityViewSetting;
+
 import org.springframework.data.domain.Pageable;
 import org.springframework.data.domain.Sort;
 import org.springframework.data.jpa.provider.PersistenceProvider;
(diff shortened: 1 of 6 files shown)


integration/spring-data/base/src/main/java/com/blazebit/persistence/spring/data/base/query/AbstractPartTreeBlazePersistenceQuery.java
```diff
@@ -372,6 +385,20 @@ private Sort getDynamicSort(Object[] values) {
...
+
+        private <T> void applySpecification(CriteriaQuery<T> criteriaQuery, Object[] values) {
+            JpaParameter specificationParameter = parameters.getSpecificationParameter();
+            if (specificationParameter != null) {
+                int specificationParameterIndex = specificationParameter.getIndex();
+                Specification<T> specification = (Specification<T>) values[specificationParameterIndex];
+                Root<T> root = (Root<T>) criteriaQuery.getRoots().iterator().next();
+                CriteriaBuilder criteriaBuilder = getEntityManager().getCriteriaBuilder();
+                Predicate predicate = specification.toPredicate(root, criteriaQuery, criteriaBuilder);
+                if (predicate != null) {
```
beikov: Either you always replace the original predicate i.e. the one created by the PartTree by also setting null or you always add the predicate by doing something like this `criteriaQuery.where(criteriaBuilder.and(criteriaQuery.getWhere(), predicate))`

integration/spring-data/base/src/main/java/com/blazebit/persistence/spring/data/base/query/AbstractPartTreeBlazePersistenceQuery.java
```diff
@@ -372,6 +385,20 @@ private Sort getDynamicSort(Object[] values) {
...
                     : null;
         }
+
+        private <T> void applySpecification(CriteriaQuery<T> criteriaQuery, Object[] values) {
+            JpaParameter specificationParameter = parameters.getSpecificationParameter();
+            if (specificationParameter != null) {
+                int specificationParameterIndex = specificationParameter.getIndex();
+                Specification<T> specification = (Specification<T>) values[specificationParameterIndex];
+                Root<T> root = (Root<T>) criteriaQuery.getRoots().iterator().next();
+                CriteriaBuilder criteriaBuilder = getEntityManager().getCriteriaBuilder();
```
beikov: Cast the criteriaQuery to `BlazeCriteriaQuery` instead and use `getCriteriaBuilder()` to access the JPA CriteriaBuilder object.

heruan:
> Essentially, you have to add `parameters.potentiallySortsDynamically() || parameters.hasSpecification()`

Not sure where to add this check 😕

integration/spring-data/base/src/main/java/com/blazebit/persistence/spring/data/base/query/AbstractPartTreeBlazePersistenceQuery.java
```diff
@@ -76,8 +78,9 @@ public AbstractPartTreeBlazePersistenceQuery(EntityViewAwareJpaQueryMethod metho
         this.tree = new PartTree(method.getName(), domainClass);
         this.parameters = method.getJpaParameters();
 
-        this.query = isCountProjection(tree) ? new AbstractPartTreeBlazePersistenceQuery.CountQueryPreparer(persistenceProvider, parameters.potentiallySortsDynamically())
-                : new AbstractPartTreeBlazePersistenceQuery.QueryPreparer(persistenceProvider, parameters.potentiallySortsDynamically());
+        boolean recreateQueries = parameters.potentiallySortsDynamically();
```
beikov: This should be `boolean recreateQueries = parameters.potentiallySortsDynamically() || parameters.hasOptionalParameters();` no? Otherwise it would be cached and a second invocation wouldn't add the optional parameters.

PR #627, opened 2018-08-01
[#626] Avoid using JPQL keywords as root aliases

### LINKED ISSUE #626: Avoid using JPQL keywords as root aliases
It would be useful to avoid creating an alias which is also a JPQL keyword, which might throw an exception.

heruan:
## Description
Adds a `Keywords` util class with a list of JPQL keywords from the JPA BNF and checks root aliases for inclusion in the list, in which case the alias generation is delegated to the alias manager.

## Related Issue
Fixes #626


beikov:
I added your commit to https://github.com/Blazebit/blaze-persistence/pull/611

PR #642, opened 2018-08-10
[#622] Add support for EntityViewSetting in Spring

### LINKED ISSUE #622: Add support for using EntityViewSetting in Spring Data Repository methods
Follow up of #564
Add support for using an `EntityViewSetting` parameter in Spring Data repositories.

heruan:
## Description
Adds `EntityViewSettingProcessor` interface which can be used as a parameter of Spring Data repository to customize the `EntityViewSetting` during the query creation.

## Related Issue
Fixes #622

### DIFF
--- a/integration/spring-data/base/src/main/java/com/blazebit/persistence/spring/data/base/query/AbstractPartTreeBlazePersistenceQuery.java
+++ b/integration/spring-data/base/src/main/java/com/blazebit/persistence/spring/data/base/query/AbstractPartTreeBlazePersistenceQuery.java
@@ -24,6 +24,7 @@ import com.blazebit.persistence.criteria.BlazeCriteria;
 import com.blazebit.persistence.criteria.BlazeCriteriaBuilder;
 import com.blazebit.persistence.criteria.BlazeCriteriaQuery;
 import com.blazebit.persistence.spring.data.base.query.JpaParameters.JpaParameter;
+import com.blazebit.persistence.spring.data.repository.EntityViewSettingProcessor;
 import com.blazebit.persistence.view.EntityViewManager;
 import com.blazebit.persistence.view.EntityViewSetting;
 
@@ -85,13 +86,13 @@ public abstract class AbstractPartTreeBlazePersistenceQuery extends AbstractJpaQ
 
         this.parameters = method.getJpaParameters();
         String methodName = method.getName();
-        boolean matchesSpecificationSignature = parameters.hasSpecificationParameter()
-            && QUERY_PATTERN.matcher(methodName).matches();
-        String source = matchesSpecificationSignature ? "" : methodName;
+        boolean skipMethodNamePredicateMatching = QUERY_PATTERN.matcher(methodName).matches();
+        String source = skipMethodNamePredicateMatching ? "" : methodName;
         this.tree = new PartTree(source, domainClass);
 
+        boolean hasEntityViewSettingProcessorParameter = parameters.hasEntityViewSettingProcessorParameter();
         boolean recreateQueries = parameters.potentiallySortsDynamically() || entityViewClass != null
-            || matchesSpecificationSignature;
+            || skipMethodNamePredicateMatching || hasEntityViewSettingProcessorParameter;
         this.query = isCountProjection(tree) ? new AbstractPartTreeBlazePersistenceQuery.CountQueryPreparer(persistenceProvider,
             recreateQueries) : new AbstractPartTreeBlazePersistenceQuery.QueryPreparer(persistenceProvider, recreateQueries);
     }
@@ -310,7 +311,15 @@ public abstract class AbstractPartTreeBlazePersistenceQuery extends AbstractJpaQ
             return binder.bind(jpaQuery);
         }
 
-        protected void processSetting(EntityViewSetting<?, ?> setting, Object[] values) {
+        @SuppressWarnings("unchecked")
+        protected <T> void processSetting(EntityViewSetting<T, ?> setting, Object[] values) {
+            int entityViewSettingProcessorIndex = parameters.getEntityViewSettingProcessorIndex();
+            if (entityViewSettingProcessorIndex >= 0) {
+                EntityViewSettingProcessor<T> processor = (EntityViewSettingProcessor<T>) values[entityViewSettingProcessorIndex];
+                if (processor != null) {
+                    setting = processor.acceptEntityViewSetting(setting);
+                }
+            }
             for (JpaParameter parameter : parameters.getOptionalParameters()) {
                 String parameterName = parameter.getParameterName();
                 Object parameterValue = values[parameter.getIndex()];
--- a/integration/spring-data/base/src/main/java/com/blazebit/persistence/spring/data/repository/EntityViewSettingProcessor.java
+++ b/integration/spring-data/base/src/main/java/com/blazebit/persistence/spring/data/repository/EntityViewSettingProcessor.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2014 - 2018 Blazebit.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.blazebit.persistence.spring.data.repository;
+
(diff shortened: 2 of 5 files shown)


integration/spring-data/base/src/main/java/com/blazebit/persistence/spring/data/repository/EntityViewSettingProcessor.java
```diff
@@ -0,0 +1,28 @@
...
+
+import com.blazebit.persistence.view.EntityViewSetting;
+
+/**
+ * @author Giovanni Lovato
+ * @since 1.3.0
+ */
+public interface EntityViewSettingProcessor<T> {
+
+    void acceptEntityViewSetting(EntityViewSetting<T, ?> setting);
```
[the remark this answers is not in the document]
beikov: How about allowing to return the final `EntityViewSetting` that should be used? That way, one could decide to make use of an entity view subtype.

integration/spring-data/base/src/main/java/com/blazebit/persistence/spring/data/base/query/AbstractPartTreeBlazePersistenceQuery.java
```diff
@@ -304,12 +306,18 @@ Query createPaginatedQuery(Object[] values, boolean withCount) {
...
+        protected <T> void processSetting(EntityViewSetting<T, ?> setting, Object[] values) {
             for (JpaParameter parameter : parameters.getOptionalParameters()) {
                 String parameterName = parameter.getParameterName();
                 Object parameterValue = values[parameter.getIndex()];
                 setting.addOptionalParameter(parameterName, parameterValue);
             }
+            int entityViewSettingProcessorIndex = parameters.getEntityViewSettingProcessorIndex();
+            if (entityViewSettingProcessorIndex >= 0) {
+                EntityViewSettingProcessor<T> processor = (EntityViewSettingProcessor<T>) values[entityViewSettingProcessorIndex];
+                processor.acceptEntityViewSetting(setting);
```
beikov: Put a not-null check around

integration/spring-data/testsuite/src/test/java/com/blazebit/persistence/spring/data/testsuite/repository/DocumentRepository.java
```diff
@@ -78,4 +79,6 @@
     Page<DocumentView> findByNameOrderById(String name, Pageable pageable, @OptionalParam("optionalParameter") String optionalParameter);
 
     List<DocumentView> findAll(Specification<Document> specification, @OptionalParam("optionalParameter") String optionalParameter);
+
+    List<DocumentView> findAll(EntityViewSettingProcessor<DocumentView> processor);
```
beikov: Extract the `getId()` part into an interface called `IdHolderView` and add a method `<T extends IdHolderView<Long>> List<T> findAll2(EntityViewSettingProcessor<T> processor)`.

It should be possible to is like this

```
List<DocumentView> result = documentRepository.findAll2(new EntityViewSettingProcessor<DocumentView>() {
            @Override
            public EntityViewSetting<DocumentView, ?> acceptEntityViewSetting(EntityViewSetting<DocumentView, ?> setting) {
                setting.addOptionalParameter("optionalParameter", "Foo");
                return setting;
            }
});
```

integration/spring-data/base/src/main/java/com/blazebit/persistence/spring/data/base/query/AbstractPartTreeBlazePersistenceQuery.java
```diff
@@ -87,11 +88,12 @@ public AbstractPartTreeBlazePersistenceQuery(EntityViewAwareJpaQueryMethod metho
         String methodName = method.getName();
         boolean matchesSpecificationSignature = parameters.hasSpecificationParameter()
             && QUERY_PATTERN.matcher(methodName).matches();
-        String source = matchesSpecificationSignature ? "" : methodName;
+        boolean hasEntityViewSettingProcessorParameter = parameters.hasEntityViewSettingProcessorParameter();
+        String source = (matchesSpecificationSignature || hasEntityViewSettingProcessorParameter) ? "" : methodName;
```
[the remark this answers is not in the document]
heruan: Right, it shouldn't. I did it this way to allow `findAll`-like method names, but I'll do it the right way by ignore the method name just when it matches the `QUERY_PATTERN`.

integration/spring-data/base/src/main/java/com/blazebit/persistence/spring/data/repository/EntityViewSettingProcessor.java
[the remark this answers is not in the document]
heruan: 👍

integration/spring-data/testsuite/src/test/java/com/blazebit/persistence/spring/data/testsuite/repository/DocumentRepository.java
heruan: Not sure what do you mean with the `IdHolderView` part. I've updated the PR with all the other changes, I'll push this as soon as I understand better what should I change!

integration/spring-data/base/src/main/java/com/blazebit/persistence/spring/data/base/query/AbstractPartTreeBlazePersistenceQuery.java
```diff
@@ -310,12 +312,20 @@ Query createPaginatedQuery(Object[] values, boolean withCount) {
...
             for (JpaParameter parameter : parameters.getOptionalParameters()) {
                 String parameterName = parameter.getParameterName();
                 Object parameterValue = values[parameter.getIndex()];
                 setting.addOptionalParameter(parameterName, parameterValue);
             }
+            int entityViewSettingProcessorIndex = parameters.getEntityViewSettingProcessorIndex();
+            if (entityViewSettingProcessorIndex >= 0) {
+                EntityViewSettingProcessor<T> processor = (EntityViewSettingProcessor<T>) values[entityViewSettingProcessorIndex];
+                if (processor != null) {
+                    processor.acceptEntityViewSetting(setting);
```
beikov: Use `setting = processor.acceptEntityViewSetting(setting);` instead to allow the processor to override the setting.

integration/spring-data/testsuite/src/test/java/com/blazebit/persistence/spring/data/testsuite/repository/DocumentRepository.java
beikov: I think I wanted to make sure that this kind of usage works i.e. have another findAll method that allows to return different subtypes. AFAICS the current logic would try to interpret `2` as predicate when the method is named `findAll2`. Now what do you say, would it be a good idea to allow multiple of these findAll methods or should a user just have a single one?

integration/spring-data/base/src/main/java/com/blazebit/persistence/spring/data/base/query/AbstractPartTreeBlazePersistenceQuery.java
heruan: What if I move the processor before applying the optional parameters, so that the processor can override the setting and the optional parameters are then applied to the new setting?

integration/spring-data/testsuite/src/test/java/com/blazebit/persistence/spring/data/testsuite/repository/DocumentRepository.java
heruan: The question here is how much we want to relax the Spring naming convention of method names. By default a Spring repository query method is supposed to have at least one predicate in the method name; we are now permitting no predicate embedded in the method name if:

- there is a `Specification` (any method name);
- the method name matches `^(find|read|get|query|stream)All$`.

What if we simply check the presence of the `By` keyword and skip the method name parsing if it's not present? Unfortunately Spring supports a (presumably undocumented) short syntax such as `ageGreaterThan` in place of `findByAgeGreaterThan`, so that might be a problem too.

integration/spring-data/base/src/main/java/com/blazebit/persistence/spring/data/base/query/AbstractPartTreeBlazePersistenceQuery.java
beikov: I'm a bit confused, isn't it like that currently?

heruan: Well, it wasn't but now is 😄 I pushed the change already 🙂

integration/spring-data/testsuite/src/test/java/com/blazebit/persistence/spring/data/testsuite/repository/DocumentRepository.java
heruan: Now only method names matching the pattern are ignored.

Issue #612, opened 2018-07-25
Prepare Hibernate 5.4 integration


beikov:
Since the Hibernate master bumped up to 5.4-SNAPSHOT we should check and see if the 5.3 integration still works and if not, provide an integration for the upcoming 5.4. If we hit some roadblocks, we should report that and maybe try to fix the issues in Hibernate before a final release.

PR #643, opened 2018-08-15
[#612] Shaded artifact for Hibernate 5.4 integration

JWGmeligMeyling:


integration/hibernate-5.4/pom.xml
```diff
@@ -0,0 +1,65 @@
...
+        <groupId>com.blazebit</groupId>
+        <artifactId>blaze-persistence-integration</artifactId>
+        <version>1.3.0-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+
+    <artifactId>blaze-persistence-integration-hibernate-5.4</artifactId>
+    <packaging>jar</packaging>
+
+    <name>Blazebit Persistence Integration Hibernate 5.4</name>
```
beikov: Could you please also add the automatic module name property?

JWGmeligMeyling: You mean https://github.com/Blazebit/blaze-persistence/blob/master/integration/hibernate-5.3/pom.xml#L34 ? I suppose I'll also have to add https://github.com/Blazebit/blaze-persistence/blob/master/integration/hibernate-5.3/pom.xml#L103-L132 then. This would by the way point to the 53 classes, as these aren't renamed by the shading. Is that sensible for the 5.4 integration artifact?

beikov: Yep, both of these sections are necessary. There is only one integration supposed to be used, so it's ok to have the same definition.

Issue #399, opened 2017-03-23
Fully support and test IdClass usages


beikov:
Currently we often assume that there is only a single id mapped in a JPA type although there could be multiple. We should fix these cases and write some tests that use `IdClass` to make sure it works.

JWGmeligMeyling:
After a quick search, the most problematic method seems to be `public static SingularAttribute<?, ?> getIdAttribute(IdentifiableType<?> entityType)`. It has about 40 uses throughout the code, so a fix for this is going to be a quite scattered throughout code.

Primarily I think this method fails to address two things:

- No exception if multiple of the singular attributes are an id attribute, because we just return and use the first found id property, which will consecutively be used throughout the queries, potentially reading and writing rows that were not intended to be touched.
- No special handling for `IdentifiableType#getIdClassAttributes()` which returns all SingularAttributes if `@IdClass` is used (which is mandatory if multiple `@Id` attributes exist)

We probably need a new method `public static Set<SingularAttribute<?, ?>> getIdAttributes(IdentifiableType<?> entityType)`.

For this I primarily see two approaches:

1. Add this additional method and use where it makes sense / where the most blocking exceptions happen (this could lead to further IdClass incompatibilities) 
2. Deprecate `public static SingularAttribute<?, ?> getIdAttribute(IdentifiableType<?> entityType)`, make it return `getIdAttributes()[0]` if `size() == 0`, else throw an exception (to prevent unexpected edge cases). Slowly rewrite all places where `getIdAttribute` is used.

JWGmeligMeyling:
I'll extend my PR gradually as I try rewriting all occasions of `getIdAttribute()`. I mainly made this test commit to check whether any existing tests pass incorrectly, but that doesn't seem to be the case. 

Note to self: also look at uses of `private boolean isId(Type<?> type, Expression idExpression)` in `JoinManager`.

ieugen:
I just ran into this issue while trying out this library. This is a show stopper for us.

beikov:
@ieugen can you elaborate what it is that isn't working for you? The support for keyset pagination with id class attributes is currently on master and will be in the next release.

JWGmeligMeyling:
I can revive the PR after I finish working on #571.

ieugen:
Well, my ap won't even start. I have a bunch of entities, one with composit pk using IdClass in Postgres, as i  provided link. Lets call it CompositA.

[...]

PR #654, opened 2018-09-04
Support IdentifiableValues using IdClass

JWGmeligMeyling:
Looking for intermediate feedback, see comments in code

### DIFF
--- a/core/testsuite/src/test/java/com/blazebit/persistence/testsuite/IdClassValueQueryTest.java
+++ b/core/testsuite/src/test/java/com/blazebit/persistence/testsuite/IdClassValueQueryTest.java
@@ -1,0 +1,96 @@
+/*
+ * Copyright 2014 - 2018 Blazebit.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.blazebit.persistence.testsuite;
+
+import com.blazebit.persistence.CriteriaBuilder;
+import com.blazebit.persistence.testsuite.entity.IdClassEntity;
+import com.blazebit.persistence.testsuite.tx.TxVoidWork;
+import org.junit.Test;
+
+import javax.persistence.EntityManager;
+import javax.persistence.TypedQuery;
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ *
+ * @author Alex Ben Hassine
+ * @since 1.3.0
+ */
+public class IdClassValueQueryTest extends AbstractCoreTest {
+
+    @Override
+    protected Class<?>[] getEntityClasses() {
+        return new Class<?>[]{
+                IdClassEntity.class
+        };
+    }
+
+    @Override
+    public void setUpOnce() {
+        cleanDatabase();
+        transactional(new TxVoidWork() {
+            @Override
+            public void work(EntityManager em) {
+                IdClassEntity e1 = new IdClassEntity(1,"1",  1);
+                IdClassEntity e2 = new IdClassEntity(2,"2",  2);
+                IdClassEntity e3 = new IdClassEntity(3,"3",  3);
+                IdClassEntity e4 = new IdClassEntity(4,"4",  4);
+                IdClassEntity e5 = new IdClassEntity(5,"5",  5);
+
+                em.persist(e1);
+                em.persist(e2);
+                em.persist(e3);
+                em.persist(e4);
+                em.persist(e5);
+            }
+        });
+    }
+
+    @Test
+    public void testValueQuery() {
+        IdClassEntity e1 = new IdClassEntity(1,"1",  1);
+        IdClassEntity e2 = new IdClassEntity(2,"2",  2);
+        IdClassEntity e3 = new IdClassEntity(3,"3",  3);
+        IdClassEntity e4 = new IdClassEntity(4,"4",  4);
+        IdClassEntity e5 = new IdClassEntity(5,"5",  5);
+
+        List<IdClassEntity> entities = new ArrayList<>();
+        entities.add(e1);
+        entities.add(e2);
+        entities.add(e3);
+        entities.add(e4);
+        entities.add(e5);
+
+        int i = 0;
+        CriteriaBuilder<Integer> cb =  cbf.create(em, Integer.class)
+                .fromIdentifiableValues(IdClassEntity.class,"myValue", entities)
+                .select("myValue.key1");
+        TypedQuery<Integer> typedQuery = cb.getQuery();
+        List<Integer> idClassEntities = typedQuery.getResultList();
+
+        for (Integer a : idClassEntities) {
+            i = i + a;
+        }
+
+        assertEquals(15, i);
+
+    }
+
+}


core/impl/src/main/java/com/blazebit/persistence/impl/AbstractCommonQueryBuilder.java
```diff
@@ -1829,7 +1829,7 @@ protected String getBaseQueryStringWithCheck() {
         for (JoinNode node : joinManager.getEntityFunctionNodes()) {
             Class<?> clazz = node.getJavaType();
             int valueCount = node.getValueCount();
-            boolean identifiableReference = node.getValuesIdName() != null;
+            boolean identifiableReference = node.getValuesIdNames() != null && !node.getValuesIdNames().isEmpty();
```
JWGmeligMeyling: Should `null` default to `Collections.emptySet()`?

core/impl/src/main/java/com/blazebit/persistence/impl/JoinManager.java
```diff
@@ -691,8 +699,12 @@ private static boolean isNoJoinValuesNode(JoinNode rootNode) {
...
                     if (type instanceof EntityType<?>) {
                         sb.append(((EntityType) type).getName());
-                        if (rootNode.getValuesIdName() != null) {
-                            sb.append('.').append(rootNode.getValuesIdName());
+
+
+                        Set<String> valuesIdNames = rootNode.getValuesIdNames();
+                        if (valuesIdNames != null && ! valuesIdNames.isEmpty()) {
+                            String next = valuesIdNames.iterator().next();
+                            sb.append(".").append(next);
```
JWGmeligMeyling: I suppose that this has to be altered, but how can one hit this code?

core/impl/src/main/java/com/blazebit/persistence/impl/JoinManager.java
```diff
@@ -851,9 +863,22 @@ void renderValuesClausePredicate(StringBuilder sb, JoinNode rootNode, String ali
             String[] attributes = rootNode.getValuesAttributes();
             String prefix = rootNode.getAlias();
 
+
+            // values ( 1,2,3)
```
JWGmeligMeyling: @AFbenHassine will remove comments and such

core/impl/src/main/java/com/blazebit/persistence/impl/JoinManager.java
```diff
@@ -385,12 +390,15 @@ public Object getValue(Object target) {
                 parameterNames[j][0] = prefix + '_' + attributeName + '_' + j;
             }
         } else if (identifiableReference) {
-            Attribute<?, ?> attribute = attributeSet.iterator().next();
-            String attributeName = attribute.getName();
-            attributes[0] = attributeName;
-            pathExpressions[0] = com.blazebit.reflection.ExpressionUtils.getExpression(clazz, attributeName);
-            for (int j = 0; j < valueCount; j++) {
-                parameterNames[j][0] = prefix + '_' + attributeName + '_' + j;
+            int i = 0;
```
JWGmeligMeyling: Functionally the same as else clause, can be removed.

core/impl/src/main/java/com/blazebit/persistence/impl/AbstractCommonQueryBuilder.java
beikov: Sounds reasonable.

core/impl/src/main/java/com/blazebit/persistence/impl/JoinManager.java
beikov: I agree with that

core/impl/src/main/java/com/blazebit/persistence/impl/JoinManager.java
beikov: When you use `fromIdentifiableValues()` and the generate the _logical query string_ via `CriteriaBuilder.getQueryString()`

core/impl/src/main/java/com/blazebit/persistence/impl/JoinManager.java
```diff
@@ -862,23 +887,35 @@ void renderValuesClausePredicate(StringBuilder sb, JoinNode rootNode, String ali
...
+                if (attributes.length > 1) {
+                    sb.append(" )");
+                }
+
+                if ( i+1 < valueCount) {
                     sb.append(" OR ");
                 }
             }
-
-            sb.setLength(sb.length() - " OR ".length());
```
beikov: I prefer a single end operation over a condition in a loop if possible. This code pattern is used all over the place.

core/impl/src/main/java/com/blazebit/persistence/impl/JoinManager.java
```diff
@@ -862,23 +887,35 @@ void renderValuesClausePredicate(StringBuilder sb, JoinNode rootNode, String ali
...
                     sb.append(" = ");
 
                     sb.append(':');
                     sb.append(prefix);
                     sb.append('_');
                     sb.append(attributes[j]);
                     sb.append('_').append(i);
+
+                    if (j+1 < attributes.length) {
+                        sb.append(" AND ");
```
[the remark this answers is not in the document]
beikov: This is just a _template query_ used to generate the proper parameter metadata, so usually, it doesn't matter what operators are used. One reason for using OR might be that I split the SQL based on that to extract aliases or something like that, but I would have to look deeper to figure that out.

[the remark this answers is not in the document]
beikov: Actually, I think it doesn't make a difference what operator you are using, as the template query only seems to be used for parameter metadata generation.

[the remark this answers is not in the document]
JWGmeligMeyling: Thats true, unless the template query is ever used as fallback or whatsoever. What is the * logical query string* for?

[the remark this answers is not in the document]
beikov: The template query string is really just for parameter metadata, it has no other use.
The logical query string is currently only used for assertions in tests and for debugging purposes, but maybe at some point I'm going to use it to cache queries.

core/impl/src/main/java/com/blazebit/persistence/impl/JoinManager.java
AFbenHassine: Hi, could you explain to me why you prefer this?

beikov: In general, I prefer it because of performance. The if will be evaluated multiple times in the loop whereas a single end operation will be evaluated only once.

AFbenHassine: Isn't the loop ended at the same moment and in the same way when removing `sb.setLength(sb.lenth() - " OR ".length());` and changing `if ( i+1 < valueCount)` to `if ( i < valueCount)`, as when these two things are kept the same before the above change?

I assumed your first comment meant that that you would prefer keeping the `sb.setLength(sb.lenth() - " OR ".length());` so as to avoid ending with a condition.

beikov: Your code does
```
for (...) {
  if (condition) {
    append
  }
}
```
Whereas I prefer this
```
for (...) {
  append
}
if (!condition) {
  remove
}
```

When the loop executes 100 times, you will check the condition 100 times which I generally like to avoid if possible.

AFbenHassine: Alright, that makes sense. Thank you.

core/testsuite/src/test/java/com/blazebit/persistence/testsuite/IdClassValueQueryTest.java
```diff
@@ -0,0 +1,96 @@
...
+                em.persist(e2);
+                em.persist(e3);
+                em.persist(e4);
+                em.persist(e5);
+            }
+        });
+    }
+
+    @Test
+    public void testValueQuery() {
```
beikov: You're missing a `@Category({ NoDatanucleus.class, NoEclipseLink.class, NoOpenJPA.class })`

core/testsuite/src/test/java/com/blazebit/persistence/testsuite/IdClassValueQueryTest.java
```diff
@@ -0,0 +1,96 @@
...
+import javax.persistence.EntityManager;
+import javax.persistence.TypedQuery;
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ *
+ * @author Christian Beikov
```
beikov: I obviously wasn't the author of this ;)

beikov:
If you fix the last couple of issues, add an entry in the CHANGELOG.md and adapt the commit message to something like `[#399] Implement fromIdentifiableValues support for IdClass entities`, I am going to merge this after my testsuite succeeds.

core/impl/src/main/java/com/blazebit/persistence/impl/JoinManager.java
```diff
@@ -709,8 +704,10 @@ private static boolean isNoJoinValuesNode(JoinNode rootNode) {
                 } else {
                     if (type instanceof EntityType<?>) {
                         sb.append(((EntityType) type).getName());
-                        if (rootNode.getValuesIdName() != null) {
-                            sb.append('.').append(rootNode.getValuesIdName());
+                        Set<String> valuesIdNames = rootNode.getValuesIdNames();
+                        if (valuesIdNames != null && ! valuesIdNames.isEmpty()) {
+                            String next = valuesIdNames.iterator().next();
+                            sb.append(".").append(next);
```
JWGmeligMeyling: I still want to look into this method to see how it can be adapted for compound keys. After that, this PR is ready to be merged.

JWGmeligMeyling:
@beikov Basically the original PR is superseded by 637f19a8f3a8f8aa66e8734af5e0242ec107c6c1 . The test just passes without any required changes now.

If you want, here is still the test. But perhaps there is already an alternative test for it. I'll continue rebasing my other commits for `IdClass` support and see whether there is anything interesting in there.

beikov:
The test will be merged with https://github.com/Blazebit/blaze-persistence/pull/683

Issue #509, opened 2018-01-25
Make use of collection management for updatable entity views in flushing


beikov:
Currently we only use the collection management features for deletion but query flushing would greatly benefit from it as well since we could avoid loading collections altogether.

beikov:
Note that we currently flush the entity manager early in `ViewCollectionRemoveListener` because it is needed to implement orphan removal(https://github.com/Blazebit/blaze-persistence/issues/507) properly for collections. Updatable collections currently cause a fallback to entity flushing which will be fixed by the solution for this issue. Don't forget to remove the flushing when implementing this.

PR #655, opened 2018-09-04
Support binding embeddables and associations using compound or foreign keys in CTE’s, fixes #571

### LINKED ISSUE #571: Support binding embeddables and associations using compound or foreign keys in CTE’s
### Description

I am facing a couple of rarities with CTE's that have a `@ManyToOne` relationship. First of all, when I do something like this: `bind("ref").select("entity.someMapping")`, all columns of `someMapping` are bound in the CTE, even though they are unused. This can be worked around with by mapping to the id instead, like so: `bind("ref").select("entity.someMapping.id")`, which works, even without changing the type of the field, but has some other quirks: in the final select query, the reference will always be lazy loaded (i.e. only the id column). Specifically adding the fetch join to the criteria builder is ignored. The effective query remains the same.

### Expected behavior
The expected behaviour is that the CTE only selects the required columns, and that it is possible to optionally fetch join these fields in the final result set.

### Actual behavior
Either the fields have to be "fetch joined" in the CTE part already (leading to an unnecessarily complex CTE query), or it is not possible to fetch join these queries at all (adding fetch joins to a CTE root seems to be ignored).

### Workaround
A workaround is to map the id and referenced entity separately, but to a different column. Then in the CTE you bind the column through the id field, and in the final query you can fetch join the entity field.

### Environment

Version:            
JPA-Provider:       
DBMS:               
Application Server:

JWGmeligMeyling:
Fixes #571 

* Support for binding associations that are (part of) the entity id
* Support for binding nested id class attributes (i.e. multiple referenced id attributes in the referenced entity)
* Support for binding associations by foreign key
* Support for binding embeddable attributes to embeddable paths directly

~~Side note:~~

~~Because I now use the CTE's column names  as actual column names, rather than the attribute names (which won't work for multi column associations, as they reference the same attribute and would have the same name ), some older tests that expect the attribute as column name now fail. Please review my approach and I will alter these tests accordingly.~~

### DIFF
--- a/core/api/src/main/java/com/blazebit/persistence/spi/JpaMetamodelAccessor.java
+++ b/core/api/src/main/java/com/blazebit/persistence/spi/JpaMetamodelAccessor.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2014 - 2018 Blazebit.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.blazebit.persistence.spi;
+
+import javax.persistence.metamodel.Attribute;
+import javax.persistence.metamodel.EntityType;
+import javax.persistence.metamodel.ManagedType;
+import javax.persistence.metamodel.Metamodel;
+
+/**
+ * @author Jan-Willem Gmelig Meyling
+ * @since 1.3.0
+ */
+public interface JpaMetamodelAccessor {
+
+    /**
+     * Construct an {@code AttributePath} for a particular attribute in type.
+     * @param metamodel JPA metamodel
+     * @param type Owning type
+     * @param attributePath The attribute path
+     * @return The created attribute path
+     */
+    AttributePath getAttributePath(Metamodel metamodel, ManagedType<?> type, String attributePath);
+
+    /**
+     * Construct an {@code AttributePath} for a particular basic attribute in type.
+     * @param metamodel JPA metamodel
+     * @param type Owning type
+     * @param attributePath The attribute path
+     * @return The created attribute path
+     */
+    AttributePath getBasicAttributePath(Metamodel metamodel, ManagedType<?> type, String attributePath);
+
+    /**
+     * Construct an {@code AttributePath} for a particular collection attribute in type.
+     * @param metamodel JPA metamodel
+     * @param type Owning type
+     * @param attributePath The attribute path
+     * @param collectionName The name of the collection
+     * @return The created attribute path
+     */
+    AttributePath getJoinTableCollectionAttributePath(Metamodel metamodel, EntityType<?> type, String attributePath, String collectionName);
+
+    /**
+     * Returns true if the attribute is joinable (i.e. association).
+     * @param attr The attribute
+     * @return Whether the attribute is joinable
+     */
+    boolean isJoinable(Attribute<?, ?> attr);
+
+    /**
+     * Returns true if the attribute is composite (i.e. embeddable).
+     * @param attr The attribute
+     * @return Whether the attribute is composite
+     */
+    boolean isCompositeNode(Attribute<?, ?> attr);
+}
--- a/core/impl/src/main/java/com/blazebit/persistence/impl/AbstractCTECriteriaBuilder.java
+++ b/core/impl/src/main/java/com/blazebit/persistence/impl/AbstractCTECriteriaBuilder.java
@@ -18,29 +18,37 @@ package com.blazebit.persistence.impl;
 
 import com.blazebit.persistence.BaseCTECriteriaBuilder;
 import com.blazebit.persistence.SelectBuilder;
-import com.blazebit.persistence.parser.expression.Expression;
-import com.blazebit.persistence.parser.expression.NullExpression;
-import com.blazebit.persistence.parser.expression.PathExpression;
-import com.blazebit.persistence.parser.expression.PropertyExpression;
 import com.blazebit.persistence.impl.query.CTEQuerySpecification;
 import com.blazebit.persistence.impl.query.CustomSQLQuery;
 import com.blazebit.persistence.impl.query.EntityFunctionNode;
 import com.blazebit.persistence.impl.query.QuerySpecification;
+import com.blazebit.persistence.parser.expression.Expression;
+import com.blazebit.persistence.parser.expression.NullExpression;
+import com.blazebit.persistence.parser.expression.PathExpression;
+import com.blazebit.persistence.parser.expression.PropertyExpression;
 import com.blazebit.persistence.parser.util.JpaMetamodelUtils;
 import com.blazebit.persistence.spi.DbmsStatementType;
 import com.blazebit.persistence.spi.ExtendedAttribute;
 import com.blazebit.persistence.spi.ExtendedManagedType;
+import com.blazebit.persistence.spi.JpaMetamodelAccessor;
 import com.blazebit.persistence.spi.SetOperationType;
 
 import javax.persistence.Query;
 import javax.persistence.metamodel.Attribute;
+import javax.persistence.metamodel.EmbeddableType;
 import javax.persistence.metamodel.EntityType;
-import java.util.ArrayList;
+import javax.persistence.metamodel.SingularAttribute;
+import javax.persistence.metamodel.Type;
+import java.util.ArrayDeque;
+import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Queue;
 import java.util.Set;
+import java.util.TreeSet;
 
 /**
  *
@@ -201,30 +209,69 @@ public abstract class AbstractCTECriteriaBuilder<Y, X extends BaseCTECriteriaBui
     }
 
     protected List<String> prepareAndGetAttributes() {
-        List<String> attributes = new ArrayList<String>(bindingMap.size());
-        for (Map.Entry<String, Integer> bindingEntry : bindingMap.entrySet()) {
-            final String attributeName = bindingEntry.getKey();
+        final Queue<String> attributeQueue = new ArrayDeque<>(bindingMap.keySet());
+        while (!attributeQueue.isEmpty()) {
+            final String attributeName = attributeQueue.remove();
+            Integer tupleIndex = bindingMap.get(attributeName);
 
-            ExtendedAttribute attributeEntry = attributeEntries.get(attributeName);
-            List<Attribute<?, ?>> attributePath = attributeEntry.getAttributePath();
-            attributes.add(attributeName);
+            final ExtendedAttribute attributeEntry = attributeEntries.get(attributeName);
+            final List<Attribute<?, ?>> attributePath = attributeEntry.getAttributePath();
+            final JpaMetamodelAccessor jpaMetamodelAccessor = mainQuery.jpaProvider.getJpaMetamodelAccessor();
+            final Attribute<?, ?> lastAttribute = attributePath.get(attributePath.size() - 1);
 
-            if (JpaMetamodelUtils.isJoinable(attributePath.get(attributePath.size() - 1))) {
-                // We have to map *-to-one relationships to their ids
-                EntityType<?> type = mainQuery.metamodel.entity(attributeEntry.getElementClass());
-                Attribute<?, ?> idAttribute = JpaMetamodelUtils.getSingleIdAttribute(type);
+            if (jpaMetamodelAccessor.isJoinable(lastAttribute) || lastAttribute.getPersistentAttributeType() == Attribute.PersistentAttributeType.EMBEDDED) {
+                // We have to map *-to-one relationships to their id or unique props
                 // NOTE: Since we are talking about *-to-ones, the expression can only be a path to an object
                 // so it is safe to just append the id to the path
-                Expression selectExpression = selectManager.getSelectInfos().get(bindingEntry.getValue()).getExpression();
+                Expression selectExpression = selectManager.getSelectInfos().get(tupleIndex).getExpression();
 
                 // TODO: Maybe also allow Treat, Case-When, Array?
                 if (selectExpression instanceof NullExpression) {
                     // When binding null, we don't have to adapt anything
                 } else if (selectExpression instanceof PathExpression) {
-                    PathExpression pathExpression = (PathExpression) selectExpression;
-                    // Only append the id if it's not already there
-                    if (!idAttribute.getName().equals(pathExpression.getExpressions().get(pathExpression.getExpressions().size() - 1).toString())) {
-                        pathExpression.getExpressions().add(new PropertyExpression(idAttribute.getName()));
+                    boolean firstBinding = true;
+                    final Collection<String> embeddedPropertyNames;
+
+                    if (lastAttribute.getPersistentAttributeType() == Attribute.PersistentAttributeType.EMBEDDED) {
+                        embeddedPropertyNames = new TreeSet<>(JpaMetamodelUtils.getEmbeddedPropertyNames((EmbeddableType<?>) ((SingularAttribute<?, ?>) lastAttribute).getType()));
+                    } else {
+                        embeddedPropertyNames = new TreeSet<>(cbf.getJpaProvider().getIdentifierOrUniqueKeyEmbeddedPropertyNames(cteType, attributeName));
+                    }
+
+                    PathExpression baseExpression = embeddedPropertyNames.size() > 1 ?
+                            ((PathExpression) selectExpression).clone(false) : ((PathExpression) selectExpression);
+
+                    joinManager.implicitJoin(baseExpression, true, null, ClauseType.SELECT, null, false, false, false, false);
+
+                    if (baseExpression.getPathReference().getType().getPersistenceType().equals(Type.PersistenceType.BASIC)) {
+                        throw new IllegalStateException("An association should be bound to its association type and not its identifier type");
+                    }
+
+                    bindingMap.remove(attributeName);
+
+                    for (String embeddedPropertyName : embeddedPropertyNames) {
+                        PathExpression pathExpression = firstBinding ?
+                                ((PathExpression) selectExpression) : baseExpression.clone(false);
+
+                        pathExpression.getExpressions().add(new PropertyExpression(embeddedPropertyName));
+                        String nestedAttributePath = attributeName + "." + embeddedPropertyName;
+                        ExtendedAttribute<?, ?> nestedAttributeEntry = attributeEntries.get(nestedAttributePath);
+
+                        // Process the nested attribute path recursively
+                        attributeQueue.add(nestedAttributePath);
+
+                        // Replace this binding in the binding map, additional selects need an updated index
+                        bindingMap.put(nestedAttributePath, firstBinding ? tupleIndex : selectManager.getSelectInfos().size());
+
+                        if (!firstBinding) {
+                            selectManager.select(pathExpression, null);
+                        } else {
+                            firstBinding = false;
+                        }
+
+                        for (String column : nestedAttributeEntry.getColumnNames()) {
+                            columnBindingMap.put(column, nestedAttributePath);
+                        }
                     }
                 } else {
                     throw new IllegalArgumentException("Illegal expression '" + selectExpression.toString() + "' for binding relation '" + attributeName + "'!");
@@ -232,7 +279,11 @@ public abstract class AbstractCTECriteriaBuilder<Y, X extends BaseCTECriteriaBui
             }
         }
 
-        return attributes;
+        String[] attributes = new String[bindingMap.size()];
+        for (Map.Entry<String, Integer> entry : bindingMap.entrySet()) {
+            attributes[entry.getValue()] = entry.getKey();
+        }
+        return Arrays.asList(attributes);
     }
 
     protected List<String> prepareAndGetColumnNames() {
@@ -258,7 +309,11 @@ public abstract class AbstractCTECriteriaBuilder<Y, X extends BaseCTECriteriaBui
             throw new IllegalStateException(sb.toString());
         }
 
-        return new ArrayList<>(columnBindingMap.keySet());
+        String[] columns = new String[columnBindingMap.size()];
+        for (Map.Entry<String, String> columnBinding : columnBindingMap.entrySet()) {
+            columns[bindingMap.get(columnBinding.getValue())] = columnBinding.getKey();
+        }
+        return Arrays.asList(columns);
     }
     
     protected BaseFinalSetOperationCTECriteriaBuilderImpl<Object, ?> createFinalSetOperationBuilder(SetOperationType operator, boolean nested, boolean isSubquery) {
--- a/core/impl/src/main/java/com/blazebit/persistence/impl/EntityMetamodelImpl.java
+++ b/core/impl/src/main/java/com/blazebit/persistence/impl/EntityMetamodelImpl.java
@@ -124,15 +124,7 @@ public class EntityMetamodelImpl implements EntityMetamodel {
                     }
                     collectColumnNames(e, attributeMap, attribute.getName(), parents, embeddableType, temporaryExtendedManagedTypes);
                 } else if (isAssociation(attribute) && !attribute.isCollection() && !jpaProvider.isForeignJoinColumn(e, attribute.getName())) {
-                    List<String> identifierOrUniqueKeyEmbeddedPropertyNames = jpaProvider.getIdentifierOrUniqueKeyEmbeddedPropertyNames(e, attribute.getName());
-
-                    for (String name : identifierOrUniqueKeyEmbeddedPropertyNames) {
-                        EntityType<?> fieldEntityType = delegate.entity(fieldType);
-                        Attribute<?, ?> idAttribute = JpaMetamodelUtils.getAttribute(fieldEntityType, name);
-                        Class<?> idType = JpaMetamodelUtils.resolveFieldClass(fieldType, idAttribute);
-                        String idPath = attribute.getName() + "." + name;
-                        attributeMap.put(idPath, new AttributeEntry(jpaProvider, e, idAttribute, idPath, idType, Arrays.asList(attribute, idAttribute)));
-                    }
+                    collectIdColumns(attribute.getName(), e, e, attributeMap, attribute, fieldType);
                 }
 
                 attributeMap.put(attribute.getName(), new AttributeEntry(jpaProvider, e, attribute, attribute.getName(), fieldType, parents));
@@ -202,6 +194,22 @@ public class EntityMetamodelImpl implements EntityMetamodel {
         }
 
         return Collections.unmodifiableMap(attributes);
+    }
+
+    private void collectIdColumns(String prefix, EntityType<?> ownerType, EntityType<?> currentType, Map<String, AttributeEntry<?, ?>> attributeMap, Attribute<?, ?> attribute, Class<?> fieldType) {
+        List<String> identifierOrUniqueKeyEmbeddedPropertyNames = jpaProvider.getIdentifierOrUniqueKeyEmbeddedPropertyNames(currentType, attribute.getName());
+
+        for (String name : identifierOrUniqueKeyEmbeddedPropertyNames) {
+            EntityType<?> fieldEntityType = delegate.entity(fieldType);
+            Attribute<?, ?> idAttribute = JpaMetamodelUtils.getAttribute(fieldEntityType, name);
+            Class<?> idType = JpaMetamodelUtils.resolveFieldClass(fieldType, idAttribute);
+            String idPath = prefix + "." + name;
+            attributeMap.put(idPath, new AttributeEntry(jpaProvider, ownerType, idAttribute, idPath, idType, Arrays.asList(attribute, idAttribute)));
+
+            if (isAssociation(idAttribute) && !idAttribute.isCollection() && !jpaProvider.isForeignJoinColumn(ownerType, idAttribute.getName())) {
+                collectIdColumns(idPath, ownerType, fieldEntityType, attributeMap, idAttribute, idType);
+            }
+        }
     }
 
     private void detectCascadingDeleteCycles(TemporaryExtendedManagedType ownerManagedType, Map<String, TemporaryExtendedManagedType> extendedManagedTypes, Set<Class<?>> cascadingDeleteCycleSet, Class<?> ownerClass, AttributeEntry<?, ?> attributeEntry, Map<Class<?>, Type<?>> classToType) {
--- a/core/parser/src/main/java/com/blazebit/persistence/parser/PathTargetResolvingExpressionVisitor.java
+++ b/core/parser/src/main/java/com/blazebit/persistence/parser/PathTargetResolvingExpressionVisitor.java
@@ -170,11 +170,11 @@ public class PathTargetResolvingExpressionVisitor implements Expression.Visitor 
 
     private Type<?> getType(Type<?> baseType, Attribute<?, ?> attribute) {
         Class<?> baseClass = baseType.getJavaType();
+
         if (baseClass != null) {
-            if (attribute.getJavaMember() instanceof Field) {
-                return metamodel.type(ReflectionUtils.getResolvedFieldType(baseClass, (Field) attribute.getJavaMember()));
-            } else if (attribute.getJavaMember() instanceof Method) {
-                return metamodel.type(ReflectionUtils.getResolvedMethodReturnType(baseClass, (Method) attribute.getJavaMember()));
+            Class<?> clazz = JpaMetamodelUtils.resolveFieldClass(baseType.getJavaType(), attribute);
+            if (clazz != null) {
+                return metamodel.type(clazz);
             }
         }
         if (attribute instanceof PluralAttribute<?, ?, ?>) {
--- a/core/testsuite/src/test/java/com/blazebit/persistence/testsuite/Issue571Test.java
+++ b/core/testsuite/src/test/java/com/blazebit/persistence/testsuite/Issue571Test.java
@@ -1,0 +1,347 @@
+/*
+ * Copyright 2014 - 2018 Blazebit.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.blazebit.persistence.testsuite;
+
+import com.blazebit.persistence.CTE;
+import com.blazebit.persistence.CriteriaBuilder;
+import com.blazebit.persistence.impl.query.CustomSQLTypedQuery;
+import com.blazebit.persistence.testsuite.base.jpa.category.NoDatanucleus;
+import com.blazebit.persistence.testsuite.base.jpa.category.NoEclipselink;
+import com.blazebit.persistence.testsuite.base.jpa.category.NoHibernate42;
+import com.blazebit.persistence.testsuite.base.jpa.category.NoHibernate43;
+import com.blazebit.persistence.testsuite.base.jpa.category.NoMySQL;
+import com.blazebit.persistence.testsuite.base.jpa.category.NoOpenJPA;
+import com.blazebit.persistence.testsuite.base.jpa.category.NoOracle;
+import com.blazebit.persistence.testsuite.entity.IdClassEntity;
+import com.blazebit.persistence.testsuite.entity.IdClassEntityId;
+import com.blazebit.persistence.testsuite.tx.TxVoidWork;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.rules.ExpectedException;
+
+import javax.persistence.Entity;
+import javax.persistence.EntityManager;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.IdClass;
+import javax.persistence.ManyToOne;
+import java.io.Serializable;
+import java.util.Objects;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * @since 1.3.0
+ * @author Jan-Willem Gmelig Meyling
+ */
+public class Issue571Test extends AbstractCoreTest {
+
+    @Rule
+    public ExpectedException expectedException = ExpectedException.none();
+
+    @Override
+    protected Class<?>[] getEntityClasses() {
+        return new Class<?>[] { Cte.class, MyEntity.class, IdClassEntity.class, IdClassReferencingIdClass.class, EntityWithNestedIdClass.class };
+    }
+
+    @Before
+    public void setUp() throws Exception {
+        transactional(new TxVoidWork() {
+            @Override
+            public void work(EntityManager em) {
+                MyEntity myEntity = new MyEntity();
+                myEntity.setAttribute("attr");
+                em.persist(myEntity);
+            }
+        });
+    }
+
+    @Test
+    // Ignore MySQL / Oracle because of unsuppored use of CTE's, ignore Hibernate 4.3 and 4.3 metamodel bug
+    @Category({ NoDatanucleus.class, NoEclipselink.class, NoOpenJPA.class, NoMySQL.class, NoOracle.class, NoHibernate42.class, NoHibernate43.class })
+    public void testBindingCteAssociationToEntity() {
+        transactional(new TxVoidWork() {
+            @Override
+            public void work(EntityManager em) {
+                CriteriaBuilder<MyEntity> select = cbf.create(em, MyEntity.class)
+                        .withRecursive(Cte.class)
+                        .from(MyEntity.class, "a")
+                        .bind("myEntity").select("a")
+                        .bind("id").select("1")
+                        .unionAll()
+                        .from(MyEntity.class, "a")
+                        .where("a.attribute").eq("bogus")
+                        .bind("myEntity").select("a")
+                        .bind("id").select("1")
+                        .end()
+                        .from(Cte.class)
+                        .select("myEntity");
+
+                select.getResultList();
+
+                String sql = ((CustomSQLTypedQuery) select.getQuery()).getQuerySpecification().getSql();
+                String cteBase = sql.substring(sql.indexOf("\nselect") + 1, sql.lastIndexOf("UNION ALL") - 1);
+                String cteRecursivePart = sql.substring(sql.indexOf("UNION ALL")+10, sql.lastIndexOf("select")-3);
+                String queryPart = sql.substring(sql.lastIndexOf("select"));
+
+                assertEquals("Recursive base part should project 2 columns separated by a single comma",
+                        cteBase.indexOf(","), cteBase.lastIndexOf(","));
+
+                assertEquals("Recursive part should project 2 columns separated by a single comma",
+                        cteRecursivePart.indexOf(","), cteRecursivePart.lastIndexOf(","));
+
+                assertEquals("Final query should project 2 columns separated by a single comma",
+                        queryPart.indexOf(","), queryPart.lastIndexOf(","));
+            }
+        });
+    }
+
+    @Test
+    // Ignore MySQL / Oracle because of unsuppored use of CTE's, ignore Hibernate 4.3 and 4.3 metamodel bug
+    @Category({ NoDatanucleus.class, NoEclipselink.class, NoOpenJPA.class, NoMySQL.class, NoOracle.class, NoHibernate42.class, NoHibernate43.class  })
+    public void testBindingCteAssociationToEntityId() {
+        expectedException.expectMessage("An association should be bound to its association type and not its identifier type");
+
+        transactional(new TxVoidWork() {
+            @Override
+            public void work(EntityManager em) {
+                CriteriaBuilder<MyEntity> select = cbf.create(em, MyEntity.class)
+                        .withRecursive(Cte.class)
+                        .from(MyEntity.class, "a")
+                        .bind("myEntity").select("a.id")
+                        .bind("id").select("1")
+                        .unionAll()
+                        .from(MyEntity.class, "a")
+                        .where("a.attribute").eq("bogus")
+                        .bind("myEntity").select("a.id")
+                        .bind("id").select("1")
+                        .end()
+                        .from(Cte.class)
+                        .select("myEntity");
+
+                select.getResultList();
+
+                String sql = ((CustomSQLTypedQuery) select.getQuery()).getQuerySpecification().getSql();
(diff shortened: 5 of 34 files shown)


core/impl/src/main/java/com/blazebit/persistence/impl/AbstractCTECriteriaBuilder.java
```diff
@@ -221,18 +219,38 @@ public CTEInfo createCTEInfo() {
...
+                    for (String idOrUniqueKeyProp : idOrUniqueKeyProps) {
+                        // For the first id prop, alter the existing expression, for consecutive props
+                        // add additional selects
+                        PathExpression pathExpression = firstBinding ?
+                                ((PathExpression) selectExpression) : baseExpression.clone(false);
+
+                        if (!idOrUniqueKeyProp.equals(pathExpression.getExpressions().get(pathExpression.getExpressions().size() - 1).toString())) {
+                            pathExpression.getExpressions().add(new PropertyExpression(idOrUniqueKeyProp));
+
+                            String nestedAttributeName = attributeName + "_a" + idOrUniqueKeyProp;
```
JWGmeligMeyling: Typo, introduced this to check to ensure that my tests didn't incorrectly pass because I used a "logical" name as alias.

core/impl/src/main/java/com/blazebit/persistence/impl/AbstractCTECriteriaBuilder.java
```diff
@@ -201,18 +201,16 @@ public CTEInfo createCTEInfo() {
...
             ExtendedAttribute attributeEntry = attributeEntries.get(attributeName);
             List<Attribute<?, ?>> attributePath = attributeEntry.getAttributePath();
-            attributes.add(attributeName);
+            JpaMetamodelAccessor jpaMetamodelAccessor = mainQuery.jpaProvider.getJpaMetamodelAccessor();
 
-            if (JpaMetamodelUtils.isJoinable(attributePath.get(attributePath.size() - 1))) {
-                // We have to map *-to-one relationships to their ids
-                EntityType<?> type = mainQuery.metamodel.entity(attributeEntry.getElementClass());
-                Attribute<?, ?> idAttribute = JpaMetamodelUtils.getSingleIdAttribute(type);
+            if (jpaMetamodelAccessor.isJoinable(attributePath.get(attributePath.size() - 1))) {
```
JWGmeligMeyling: Now that we support collections in CTE's (do we? We dropped a few earlier constraints here) We may also want to include `!isCollection` here.

core/impl/src/main/java/com/blazebit/persistence/impl/CTEManager.java
```diff
@@ -108,12 +108,12 @@ void buildClause(StringBuilder sb) {
             sb.append(cte.name);
             sb.append('(');
 
-            final List<String> attributes = cte.attributes;
-            sb.append(attributes.get(0));
+            final List<String> columnNames = cte.columnNames;
```
JWGmeligMeyling: This change is what my question is about.

core/api/src/main/java/com/blazebit/persistence/spi/JpaMetamodelAccessor.java
```diff
@@ -0,0 +1,39 @@
...
+import javax.persistence.metamodel.ManagedType;
+import javax.persistence.metamodel.Metamodel;
+
+/**
+ * @author Jan-Willem Gmelig Meyling
+ * @since 1.3.0
+ */
+public interface JpaMetamodelAccessor {
+
+    AttributePath getAttributePath(Metamodel metamodel, ManagedType<?> type, String attributePath);
```
beikov: You will need java docs for the methods as well

core/impl/src/main/java/com/blazebit/persistence/impl/AbstractCTECriteriaBuilder.java
beikov: The thing is, you can't bind collections that are contained in CTEs, so this shouldn't be a problem here. IMO we should throw an exception during bind when encountering a plural attribute and not worry about this here.

core/impl/src/main/java/com/blazebit/persistence/impl/AbstractCTECriteriaBuilder.java
```diff
@@ -201,18 +206,16 @@ public CTEInfo createCTEInfo() {
...
+                List<String> idOrUniqueKeyProps = cbf.getJpaProvider().getIdentifierOrUniqueKeyEmbeddedPropertyNames(cteType, attributeName);
```
beikov: Move this into the `selectExpression instanceof PathExpression` branch

beikov: Move this into the `selectExpression instanceof PathExpression` branch

core/impl/src/main/java/com/blazebit/persistence/impl/AbstractCTECriteriaBuilder.java
```diff
@@ -221,18 +224,40 @@ public CTEInfo createCTEInfo() {
                 if (selectExpression instanceof NullExpression) {
                     // When binding null, we don't have to adapt anything
                 } else if (selectExpression instanceof PathExpression) {
-                    PathExpression pathExpression = (PathExpression) selectExpression;
-                    // Only append the id if it's not already there
-                    if (!idAttribute.getName().equals(pathExpression.getExpressions().get(pathExpression.getExpressions().size() - 1).toString())) {
-                        pathExpression.getExpressions().add(new PropertyExpression(idAttribute.getName()));
+                    boolean firstBinding = true;
+                    PathExpression baseExpression = ((PathExpression) selectExpression).clone(false);
```
beikov: Only clone when `idOrUniqueKeyProps.size() > 1`

core/impl/src/main/java/com/blazebit/persistence/impl/AbstractCTECriteriaBuilder.java
```diff
@@ -221,18 +224,40 @@ public CTEInfo createCTEInfo() {
...
+                    for (String idOrUniqueKeyProp : idOrUniqueKeyProps) {
+                        // For the first id prop, alter the existing expression, for consecutive props
+                        // add additional selects
+                        PathExpression pathExpression = firstBinding ?
+                                ((PathExpression) selectExpression) : baseExpression.clone(false);
+
+                        if (!idOrUniqueKeyProp.equals(pathExpression.getExpressions().get(pathExpression.getExpressions().size() - 1).toString())) {
```
beikov: Can you explain why this check is necessary? IMO this is always false.

core/impl/src/main/java/com/blazebit/persistence/impl/CTEManager.java
beikov: Looks good. It makes sense that this is required for multi-column associations.

core/parser/src/main/java/com/blazebit/persistence/parser/PathTargetResolvingExpressionVisitor.java
```diff
@@ -239,6 +239,8 @@ public void visit(PropertyExpression expression) {
                     keyType = metamodel.type(typeArguments[0]);
                 }
             }
+        } else if (attribute instanceof SingularAttribute<?,?>) {
+            valueType = type = ((SingularAttribute) attribute).getType();
```
beikov: Why do you think this is necessary? If there is some use of type variables, IMO this will break like this. `type` should already contain the same or in case of type variables the concrete type.

core/testsuite/src/test/java/com/blazebit/persistence/testsuite/Issue571Test.java
```diff
@@ -0,0 +1,325 @@
...
+import javax.persistence.Id;
+import javax.persistence.IdClass;
+import javax.persistence.ManyToOne;
+import java.io.Serializable;
+import java.util.List;
+import java.util.Objects;
+
+import static org.junit.Assert.assertEquals;
+
+public class Issue571Test extends AbstractCoreTest {
```
beikov: Class doc missing

JWGmeligMeyling:
I've updated the PR, as there were still some issues. One related to recursiveness (the attributes of introduced in a previous step may have to be expanded, so I have to use a queue), another related to 
attribute/column ordering in produced SQL, and another to go back to attribute mapping rather than column mapping which may be required for multi column types.

I'll comment to your remarks separately.

core/impl/src/main/java/com/blazebit/persistence/impl/AbstractCTECriteriaBuilder.java
JWGmeligMeyling: This is necessary for two things:
1) People that bind `association.id` directly
1) Recursive calls from the same method

core/parser/src/main/java/com/blazebit/persistence/parser/PathTargetResolvingExpressionVisitor.java
JWGmeligMeyling: This is necessary because `getJavaType()` may return the type of the identifier property within the IdClass rather than the return type of the association. We need the later, as we use this method for traversing property paths solely. We're not interested in IdClass properties. We can't even query them.

core/impl/src/main/java/com/blazebit/persistence/impl/CTEManager.java
JWGmeligMeyling: I've actually changed this behaviour back, because I foreseen some issues with multi column types (ie. java money). Didn't get a chance to test it, but the new approach uses more of the existing stuff, and didn't break existing tests.

core/api/src/main/java/com/blazebit/persistence/spi/JpaMetamodelAccessor.java
JWGmeligMeyling: Added javadocs here.

testsuite-base/hibernate/src/main/java/com/blazebit/persistence/testsuite/base/AbstractPersistenceTest.java
```diff
@@ -107,7 +107,7 @@ protected Properties applyProperties(Properties properties) {
         properties.put("hibernate.ejb.metamodel.population", "enabled");
         // We use the following only for debugging purposes
         // Normally these settings should be disabled since the output would be too big TravisCI
-//        properties.put("hibernate.show_sql", "true");
+        properties.put("hibernate.show_sql", "true");
```
JWGmeligMeyling: Oops, will undo in a consecutive commit.

integration/hibernate-base/src/main/java/com/blazebit/persistence/integration/hibernate/base/HibernateJpaProvider.java
```diff
@@ -866,8 +868,14 @@ public void setCacheable(Query query) {
         boolean supportsNaturalIdAccessOptimization = false;
         AbstractEntityPersister entityPersister = getEntityPersister(owner);
         Type propertyType = entityPersister.getPropertyType(attributeName);
-        org.hibernate.type.EntityType entityType = (org.hibernate.type.EntityType) propertyType;
+
+        // For nested embedded id's, the nested id is mapped as a component type
+        if (propertyType instanceof ComponentType) {
```
beikov: Shouldn't this be done in a recursive manner?

core/parser/src/main/java/com/blazebit/persistence/parser/PathTargetResolvingExpressionVisitor.java
beikov: Would you mind adding a comment about this here? Is there a test that fails if we remove this? It just seems strange because `getType` which is used to set the value on `type` should IMO produce the same result.

core/impl/src/main/java/com/blazebit/persistence/impl/EntityMetamodelImpl.java
```diff
@@ -595,7 +608,12 @@ public AttributeEntry(JpaProvider jpaProvider, ManagedType<X> ownerType, Attribu
             JpaProvider.ConstraintType[] requiresTreatFilter = new JpaProvider.ConstraintType[joinTypes.length];
             if (ownerType instanceof EntityType<?>) {
                 EntityType<?> entityType = (EntityType<?>) ownerType;
-                this.isForeignJoinColumn = jpaProvider.isForeignJoinColumn(entityType, attributeName);
+                try {
+                    this.isForeignJoinColumn = jpaProvider.isForeignJoinColumn(entityType, attributeName);
+                }
+                catch (Exception e) {
```
beikov: I suppose you will remove this later and this is just for testing?

core/impl/src/main/java/com/blazebit/persistence/impl/EntityMetamodelImpl.java
```diff
@@ -195,6 +187,27 @@ public EntityMetamodelImpl(EntityManagerFactory emf, JpaProviderFactory jpaProvi
...
+            try {
+                Class<?> idType = JpaMetamodelUtils.resolveFieldClass(fieldType, idAttribute);
+                String idPath = prefix + "." + name;
+                attributeMap.put(idPath, new AttributeEntry(jpaProvider, ownerType, idAttribute, idPath, idType, Arrays.asList(attribute, idAttribute)));
+
+                if (isAssociation(idAttribute) && !idAttribute.isCollection() && !jpaProvider.isForeignJoinColumn(ownerType, idAttribute.getName())) {
+                    collectIdColumns(idPath, ownerType, fieldEntityType, attributeMap, idAttribute, idType);
+                }
+            }
+            catch (NullPointerException e1) {
```
beikov: Again, I assume just for testing?

core/impl/src/main/java/com/blazebit/persistence/impl/AbstractCTECriteriaBuilder.java
beikov: Maybe I'm not understanding the code properly, but to me this seems like allowing to bind `someAssociation.sameNamedProperty` to an attribute like `id` which contains a property `sameNamedProperty`. Am I understanding this correctly?

beikov: It seems this happend due to the code that was used before which tried to handle a second call to prepareAndGetAttributes properly by skipping the add of the id property if it was already there.

core/parser/src/main/java/com/blazebit/persistence/parser/PathTargetResolvingExpressionVisitor.java
JWGmeligMeyling: It could be that this is a Hibernate specific bug, I haven't checked the spec just yet. However, I've introduced a test that fails without this change.

integration/hibernate-base/src/main/java/com/blazebit/persistence/integration/hibernate/base/HibernateJpaProvider.java
JWGmeligMeyling: The only way I have reached this path is with with nesting of `@Id` association properties using `@IdClass`, i.e. `someAlias.someIdAssocation.someOtherIdAssocation.compoundKeyPart`. The type for `someAlias.someIdAssocation` is `EntityType`, but for `someAlias.someIdAssocation.someOtherIdAssocation` it is `ComponentType` (even though I would somewhat expect an `EntityType here). I haven't reached this path in any other way. Because Embeddables are almost never Joinable I don't really see how this could ever happen. Do you have a concrete example in mind?

core/impl/src/main/java/com/blazebit/persistence/impl/EntityMetamodelImpl.java
JWGmeligMeyling: Yes, should not have been comitted.

core/impl/src/main/java/com/blazebit/persistence/impl/EntityMetamodelImpl.java
JWGmeligMeyling: Yes, should not have been comitted.

core/impl/src/main/java/com/blazebit/persistence/impl/AbstractCTECriteriaBuilder.java
JWGmeligMeyling: > Maybe I'm not understanding the code properly, but to me this seems like allowing to bind someAssociation.sameNamedProperty to an attribute like id which contains a property sameNamedProperty.

Eventually, it only matters that the columns of the CTE are mapped properly. Projecting entities in CTE queries will lead to additional columns in the selection that are not defined by the CTE and potentially even messes up attribute ordering. This fix traverses the `idOrUniqueKey` attributes so that every column is mapped to the underlying referenced attribute. Ie. `someAssocitionMappedByCompoundKey` will render `someAssocitionMappedByCompoundKey.keyA, someAssocitionMappedByCompoundKey.keyB` and adjust the mapping so that each is mapped to their respective column in the cte.

The code still does exactly the same if there is only one id attributre: it adds the id attribute to the path expression if its not already there. The latter can still happen even with the added changes.

 I removed the check that was already there before and saw some tests failing, I'll get back to you as to why this happens because it seems unlogical.

beikov: Actually, I think it's ok like this. Just re-add the comment that you removed.

core/impl/src/main/java/com/blazebit/persistence/impl/AbstractCTECriteriaBuilder.java
```diff
@@ -201,38 +203,68 @@ public CTEInfo createCTEInfo() {
...
                 // so it is safe to just append the id to the path
-                Expression selectExpression = selectManager.getSelectInfos().get(bindingEntry.getValue()).getExpression();
+                Expression selectExpression = selectManager.getSelectInfos().get(tupleIndex).getExpression();
 
                 // TODO: Maybe also allow Treat, Case-When, Array?
                 if (selectExpression instanceof NullExpression) {
                     // When binding null, we don't have to adapt anything
                 } else if (selectExpression instanceof PathExpression) {
-                    PathExpression pathExpression = (PathExpression) selectExpression;
-                    // Only append the id if it's not already there
```
beikov: Re-add this comment please

integration/hibernate-base/src/main/java/com/blazebit/persistence/integration/hibernate/base/HibernateJpaProvider.java
beikov: Now that I think of it, it's fine!

core/impl/src/main/java/com/blazebit/persistence/impl/AbstractCTECriteriaBuilder.java
JWGmeligMeyling: Allright so the idea behind this rewrite is that we simply can't bind association types to plain objects, and have to bind to the underlying identifier keys instead. It currently *is* however allowed, to bind an CTE association attribute directly to its identifier, i.e. `bind("someAttribute").to("someAssociation.id")` as opposed to `bind("someAttribute").to("someAssociation.id")`. There are tests that verify that both are currently allowed.

When in this method, we obtain the *identifier or unique properties* of the association attribute *on the CTE*, i.e. the inverse mapping from the join columns of the attribute. In the case a single attribute is used (most of the time) it is still possible that the binding goes to either `"someAssociation"` or `"someAssociation.id"`. In the case of the latter, we have to check whether we're not adding the property name twice.

For compound keys, it is unlikely that a user would attempt to bind to the id attribute directly, as this would involve two different bindings. I.e. there is no point in binding `bind("someAttribute").to("someAssociation.key1")`, so one would bind the keys seperately any way (`bind("someAttribute.key1").to("someAssociation.key1")). Because `"someAttribute.key1"` is not a *Joinable* attribute, the code won't even attempt to optimise this code (correctly).

So the code works, and the check has to remain in place. The only problem I have is that it is technically possible to have a collision between the association attribute name and the referenced identifier or unique key names. I.e. a case mapping `someAssoc.someId.someId` would be required, this would incorrectly be translated into `someAssoc.someId`. In order to fix this, we have to resolve the `Attribute` of the `PathExpression`, (is this possible through the `PathTargetResolvingExpressionVisitor`?), and verify that this `Attribute` is  not equal to the identifier or unique key `Attribute` on the referenced entity type.

core/parser/src/main/java/com/blazebit/persistence/parser/PathTargetResolvingExpressionVisitor.java
JWGmeligMeyling: After debugging this, I think this may be a Hibernate bug. Its just odd that `metamodel.entity(Entity.class). getSingularAttribute("key1").getMember()` returns `Entity.EntityId#key1` rather than `Entity#key1`. I can't find anything relevant in the JPA 2.1 spec as to why this would be useful. Nevertheless, this seems like a proper workaround. What is your worry with type variables?

It by the way is the `testBindingCteUsingNestedIdClass` test that fails without this change.

JWGmeligMeyling:
- Resolved some merge conflicts
- Resolved some compilation errors
- Ensured checkstyle / test passes on H2/Hibernate
- Responded to your comments
- Did some more thorough testing, I am pretty confident regarding the implementation right now


After the tests for DN, EclipseLink and OpenJPA have passed/failed I'll update the PR accordingly. Either I have to disable some tests under these profiles or add some additional changes. I also may have to add some changes for properly handling https://github.com/Blazebit/blaze-persistence/pull/655#discussion_r216375352 .

I am also considering to further test support for associations within embeddables and especially directly binding embedabbles to embeddables, do you foresee any problems? Is such behaviour already implemented or tested?

core/impl/src/main/java/com/blazebit/persistence/impl/AbstractCTECriteriaBuilder.java
beikov: IMO binding `someAssociation` to `someAssocation.id` shouldn't be allowed. The thing is, that the code alters the binding and select expressions i.e. it adds the id property to the path expression. To prevent adding the id multiple times in case the `prepare` is executed multiple times, it is checked whether the path already ends with that and if so, skip adding it again. Ideally, we would split embeddables and assocations to binds of the individual properties, but I'm not sure how this could work out. I'd be very much interested in such an approach though as I was in need for something like that recently for #509.

Splitting off to individual property binds might bring a few possible problems. For once, you might have to transform a parameter by extracting an attribute value. This shouldn't be too hard, the ParameterTransformer is there to help you.
The other possible issue that comes to my mind is that you might have to order the binds in a specific way so that values are bound to the correct CTE attributes in the SQL, but I haven't given that possible problem enough thought yet. Maybe try to add a test that does a fetch join on a multi-column assocation of a CTE in the main query to verify it works properly.

core/parser/src/main/java/com/blazebit/persistence/parser/PathTargetResolvingExpressionVisitor.java
beikov: If you look into `com.blazebit.persistence.parser.util.JpaMetamodelUtils#getIdAttributes` you can see special handling for Hibernate which behave strange when encountering a `@ManyToOne` id attribute. I think this is what might be biting you here.

beikov:
Apparently you missed a checkstyle failure: https://travis-ci.org/Blazebit/blaze-persistence/jobs/426740327#L824

Except for the parameter case which requires the use of a ParameterTransformer, I don't see any problems and very much welcome any efforts in that area. I was having trouble myself within the collection DML API with that recently where I tried to bind embeddables directly but had to bind individual attributes instead. Automatic expansion for that and assocations would be awesome!

core/parser/src/main/java/com/blazebit/persistence/parser/PathTargetResolvingExpressionVisitor.java
JWGmeligMeyling: That exception should be for Hibernate 4 solely afaik? It is presented as Id attribute, but the java member and java type values just don't make sense. I've posted a question at https://discourse.hibernate.org/t/bug-in-jpa-compliance/1372/2 and 
https://gitter.im/hibernate/hibernate-orm?at=5b96b3cd33ebb72e37f9c882 , I may open a Jira ticket if I do not get some proper explanation.

What are your worries regarding type variables for this work around? Could I test whether that still works?

beikov: I think if the resolver tries to resolve the path `e.association.key1` on a model like
```
@Entity class Base<T extends BaseAssoc> { @ManyToOne T assocation; }
@Entity class Sub1 extends Base<Assoc1> {}
@Entity class Sub2 extends Base<Assoc2> {}
@Entity class BaseAssoc {}
@Entity class Assoc1 extends BaseAssoc { Integer key1; }
@Entity class Assoc2 extends BaseAssoc { String key1; }
```

You might get weird results when using this code.

core/impl/src/main/java/com/blazebit/persistence/impl/AbstractCTECriteriaBuilder.java
beikov: > I can alter the code to do this, it's not going to be easy though, but doing so would allow for some additional checks that might be worthwhile. I would need to resolve the attribute of the path expression against the meta model though. Is there an expression visitor for that? One issue I see is that I have to pick a class, which will mean I have to figure out whether to resolve to the query root or a join alias in the alias manager. Is such code already out there somewhere?

The PathTargetResolvingExpressionVisitor should be capable of doing that. Note that you only need to analyze this further when the bind attribute is an assocation. If it is an assocation, the expression can only be a PathExpression, so it should be easy to check.

> Do you mean to split of these individual properties at bind time? Or do you think that a similar approach, namely post processing them in prepareAndGetAttributes, would suffice? why would or wouldn't that work? As far as I can see, it should be possible to solve this similarly to how the association type is handled now.

I would split off the properties to individual binds during after the type check of the target expression.

beikov:
I usually do a `mvn clean install -DskipTests` to also run checkstyle.

core/parser/src/main/java/com/blazebit/persistence/parser/PathTargetResolvingExpressionVisitor.java
JWGmeligMeyling: Ill have a try and see what happens. Is such a domain already present in our test code base?

beikov: I think `PolymorphicPropertySub1` and friends should fit that model.

beikov:
There are a few tests that require some adaptions for the new rendering: https://travis-ci.org/Blazebit/blaze-persistence/jobs/426848643

beikov:
Also, 4.2 and 4.3 seem to have problems with the entity model: https://travis-ci.org/Blazebit/blaze-persistence/jobs/426848714#L2041
You should use `NoHibernate42` and `NoHibernate 4.3` categories for ignoring the tests for these cases.

JWGmeligMeyling:
Hi Christian! I was planning to look into the last incompatibilities shown by the test suite in a bit, thanks for giving some pointers already!

JWGmeligMeyling:
Implementing embeddable was actually required to make an existing test work. It turned out to be very similar to what we do with associations, so it could be implemented rather smoothly. However, I had to spend quite a bit of time on a misplaced statement that caused nondeterministic results. This issue is however now resolved.

I've fixed most of the tests now, although I still may have to adjust my fix for EclipseLink (the `JpaMetamodelAccessor` related stuff). Furthermore, I want to look into using a `PathTargetREsolvingExpressionVisitor` to properly check for the duplicate id / incompatible type bindings of CTE attributes.

beikov:
I don't think it's necessary to use `PathTargetResolvingExpressionVisitor` as I wrote you on Slack. You should be able to get away with `PathExpression.getPathReference().getType()`

core/impl/src/main/java/com/blazebit/persistence/impl/AbstractCTECriteriaBuilder.java
```diff
@@ -201,38 +209,83 @@ public CTEInfo createCTEInfo() {
...
+
+                    for (String embeddedPropertyName : embeddedPropertyNames) {
+                        PathExpression pathExpression = firstBinding ?
+                                ((PathExpression) selectExpression) : baseExpression.clone(false);
+
+                        // TODO: this is supposed to check whether someAssociation is bound to someAlias.id
+                        // Actually, the id names can collide in the case of IdClass or EmbeddedIds, so a true check
+                        // should check whether the type of the path reference of the path expression is a basic type.
+                        // However, this seems not possible at this time because for some tests pathExpression.getPathReference()
+                        // returns null.
```
JWGmeligMeyling: @beikov Can you have a look at this comment?

core/impl/src/main/java/com/blazebit/persistence/impl/AbstractCTECriteriaBuilder.java
```diff
@@ -201,38 +209,83 @@ public CTEInfo createCTEInfo() {
...
-                    // Only append the id if it's not already there
-                    if (!idAttribute.getName().equals(pathExpression.getExpressions().get(pathExpression.getExpressions().size() - 1).toString())) {
-                        pathExpression.getExpressions().add(new PropertyExpression(idAttribute.getName()));
+                    boolean firstBinding = true;
+                    final Collection<String> embeddedPropertyNames;
+
+                    if (lastAttribute.getPersistentAttributeType() == Attribute.PersistentAttributeType.EMBEDDED) {
+                        embeddedPropertyNames = new TreeSet<>(JpaMetamodelUtils.getEmbeddedPropertyNames((EmbeddableType<?>) ((SingularAttribute<?, ?>) lastAttribute).getType()));
+                    } else {
+                        embeddedPropertyNames = new TreeSet<>(cbf.getJpaProvider().getIdentifierOrUniqueKeyEmbeddedPropertyNames(cteType, attributeName));
```
JWGmeligMeyling: I had to maintain a consistent ordering to ensure that tests pass between various build profiles, because some assertions are dependent on attribute order.

core/parser/src/main/java/com/blazebit/persistence/parser/PathTargetResolvingExpressionVisitor.java
```diff
@@ -170,11 +171,26 @@ public PathTargetResolvingExpressionVisitor(EntityMetamodel metamodel, Type<?> s
...
         Class<?> baseClass = baseType.getJavaType();
+        Member member = attribute.getJavaMember();
+
+        // Workaround HHH-12969: Hibernate returning the member of the identifier attribute of
+        // the IdClass as member of the association attribute on the entity.
+        if (attribute instanceof SingularAttribute) {
+            if (member instanceof Method) {
+                member = ReflectionUtils.getMethod(baseClass, member.getName());
+            } else if (member instanceof Field) {
+                member = ReflectionUtils.getField(baseClass, member.getName());
```
JWGmeligMeyling: Locally the test seemed to pass for Eclipselink, but maybe I messed something up in my profile configuration. I'll just wait for CI.

The code below makes me think that this code branch should work for Eclipselink too, but I could be mistaken. Do you want me to move this to the `JpaMetamodelAccessor` even if the existing tests would pass?

core/impl/src/main/java/com/blazebit/persistence/impl/AbstractCTECriteriaBuilder.java
```diff
@@ -201,38 +209,83 @@ public CTEInfo createCTEInfo() {
...
                 } else if (selectExpression instanceof PathExpression) {
-                    PathExpression pathExpression = (PathExpression) selectExpression;
-                    // Only append the id if it's not already there
-                    if (!idAttribute.getName().equals(pathExpression.getExpressions().get(pathExpression.getExpressions().size() - 1).toString())) {
-                        pathExpression.getExpressions().add(new PropertyExpression(idAttribute.getName()));
+                    boolean firstBinding = true;
+                    final Collection<String> embeddedPropertyNames;
+
+                    if (lastAttribute.getPersistentAttributeType() == Attribute.PersistentAttributeType.EMBEDDED) {
+                        embeddedPropertyNames = new TreeSet<>(JpaMetamodelUtils.getEmbeddedPropertyNames((EmbeddableType<?>) ((SingularAttribute<?, ?>) lastAttribute).getType()));
```
JWGmeligMeyling: Actually the implementation for Hibernate of `getIdentifierOrUniqueKeyEmbeddedPropertyNames` already is able to extract keys from component types, because nested id types are wrapped into virtual component types and it was mandatory to account for that. Needless to say, I think it is neater here to treat the embedded attribute separately here. Less room for errors across various ORMs.

Do you think the casts are safe here?

core/impl/src/main/java/com/blazebit/persistence/impl/AbstractCTECriteriaBuilder.java
```diff
@@ -201,38 +209,83 @@ public CTEInfo createCTEInfo() {
...
+        while (!attributeQueue.isEmpty()) {
+            final String attributeName = attributeQueue.remove();
+            Integer tupleIndex = bindingMap.get(attributeName);
+
+            final ExtendedAttribute attributeEntry = attributeEntries.get(attributeName);
+            final List<Attribute<?, ?>> attributePath = attributeEntry.getAttributePath();
+            final JpaMetamodelAccessor jpaMetamodelAccessor = mainQuery.jpaProvider.getJpaMetamodelAccessor();
+            final Attribute<?, ?> lastAttribute = attributePath.get(attributePath.size() - 1);
+
+            if (jpaMetamodelAccessor.isJoinable(lastAttribute) || lastAttribute.getPersistentAttributeType() == Attribute.PersistentAttributeType.EMBEDDED) {
```
JWGmeligMeyling: Perhaps I should use `isComponentType` here? Any suggestions as to how to improve this check across ORMs?

core/impl/src/main/java/com/blazebit/persistence/impl/AbstractCTECriteriaBuilder.java
```diff
@@ -201,38 +209,83 @@ public CTEInfo createCTEInfo() {
...
+                        PathExpression pathExpression = firstBinding ?
+                                ((PathExpression) selectExpression) : baseExpression.clone(false);
+
+                        // TODO: this is supposed to check whether someAssociation is bound to someAlias.id
+                        // Actually, the id names can collide in the case of IdClass or EmbeddedIds, so a true check
+                        // should check whether the type of the path reference of the path expression is a basic type.
+                        // However, this seems not possible at this time because for some tests pathExpression.getPathReference()
+                        // returns null.
+                        if (embeddedPropertyName.equals(pathExpression.getExpressions().get(pathExpression.getExpressions().size() - 1).toString())) {
+                            throw new IllegalStateException("An association should be bound to its association type and not its identifier type");
```
JWGmeligMeyling: I'm throwing an error here now because as we discussed, we don't want users to bind entity types directly to identifier values.

integration/jpa-base/pom.xml
```diff
@@ -114,6 +125,7 @@
                                         requires java.sql;
                                         requires java.persistence;
                                         requires com.blazebit.persistence.core;
+                                        exports com.blazebit.persistence.integration.jpa;
```
JWGmeligMeyling: Does this make the entry below superfluous?

JWGmeligMeyling:
Tested for various DB's and ORMs locally, the branch should be as good as finished. I have a couple of implementation questions though, as indicated by my questions. I'll progress these comments and any eventual CI failure later.

core/impl/src/main/java/com/blazebit/persistence/impl/AbstractCTECriteriaBuilder.java
beikov: Seems fine to me

core/impl/src/main/java/com/blazebit/persistence/impl/AbstractCTECriteriaBuilder.java
beikov: Yep, seems ok

core/impl/src/main/java/com/blazebit/persistence/impl/AbstractCTECriteriaBuilder.java
```diff
@@ -201,38 +208,83 @@ public CTEInfo createCTEInfo() {
...
+
+                        pathExpression.getExpressions().add(new PropertyExpression(embeddedPropertyName));
+                        String nestedAttributePath = attributeName + "." + embeddedPropertyName;
+                        ExtendedAttribute<?, ?> nestedAttributeEntry = attributeEntries.get(nestedAttributePath);
+
+                        // Process the nested attribute path recursively
+                        attributeQueue.add(nestedAttributePath);
+
+                        // Replace this binding in the binding map, additional selects need an updated index
+                        bindingMap.remove(attributeName);
```
beikov: Move this out of the loop

JWGmeligMeyling: I'm worried about the case where the embeddedProps list may be empty, in which case we do not optimize anything yet remove it from the attribute map. Could such a thing occur? I.e. if the inverse property is not mapped in an entity?

core/parser/src/main/java/com/blazebit/persistence/parser/PathTargetResolvingExpressionVisitor.java
beikov: Maybe it's easier if we use something like this

```
Class<?> clazz = com.blazebit.persistence.parser.util.JpaMetamodelUtils#resolveFieldClass(baseType.getJavaType(), attribute);
if (clazz != null) {
  return metamodel.type(clazz);
}

if (attribute instanceof PluralAttribute<?, ?, ?>) {
  return ((PluralAttribute<?, ?, ?>) attribute).getElementType();
}
return ((SingularAttribute<?, ?>) attribute).getType();
```

core/testsuite/src/test/java/com/blazebit/persistence/testsuite/Issue571Test.java
```diff
@@ -0,0 +1,342 @@
...
+            public void work(EntityManager em) {
+                MyEntity myEntity = new MyEntity();
+                myEntity.setAttribute("attr");
+                em.persist(myEntity);
+            }
+        });
+    }
+
+    @Test
+    @Category({ NoDatanucleus.class, NoEclipselink.class, NoOpenJPA.class, NoMySQL.class, NoOracle.class, NoHibernate42.class, NoHibernate43.class })
```
beikov: Please add a note about why MySQL, Oracle and Hibernate 4.2 and 4.3 aren't supported.

core/testsuite/src/test/java/com/blazebit/persistence/testsuite/Issue571Test.java
```diff
@@ -0,0 +1,342 @@
...
+                        .unionAll()
+                        .from(MyEntity.class, "a")
+                        .where("a.attribute").eq("bogus")
+                        .bind("myEntity").select("a.id")
+                        .bind("id").select("1")
+                        .end()
+                        .from(Cte.class)
+                        .select("myEntity");
+
+                List<MyEntity> resultList = select
```
beikov: Instead of ignoring the test, please catch the exception and verify that the right exception is thrown i.e. check for a known part of the exception message.

integration/jpa-base/pom.xml
beikov: No

entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/update/EntityViewUpdaterImpl.java
```diff
@@ -480,7 +480,7 @@ public static ViewToEntityMapper createViewIdMapper(EntityViewManagerImpl evm, M
     }
 
     @SuppressWarnings("unchecked")
-    private static void buildComponentFlushers(EntityMetamodel metamodel, Class<?> rootType, String attributePrefix, String mappingPrefix, String accessorPrefix, Set<Attribute<?, ?>> attributes, Map<AttributeAccessor, BasicAttributeFlusher> componentFlushers) {
+    private static void buildComponentFlushers(EntityViewManagerImpl evm, EntityMetamodel metamodel, Class<?> rootType, String attributePrefix, String mappingPrefix, String accessorPrefix, Set<Attribute<?, ?>> attributes, Map<AttributeAccessor, BasicAttributeFlusher> componentFlushers) {
```
beikov: Since you added the `EntityviewManagerImpl` to many methods now, could you please remove the `EntityMetamodel` parameter and instead access that in the methods via `evm.getMetamodel().getEntityMetamodel()`?

core/impl/src/main/java/com/blazebit/persistence/impl/AbstractCTECriteriaBuilder.java
beikov: I'm not sure this is possible. You shouldn't be able to bind an inverse property

core/impl/src/main/java/com/blazebit/persistence/impl/AbstractCTECriteriaBuilder.java
beikov: That's perfect!

core/testsuite/src/test/java/com/blazebit/persistence/testsuite/Issue571Test.java
JWGmeligMeyling: MySQL and Oracle because these issues require CTE's and all tests exclude them.
HH4.2 and 4.3 due to a metamodel bug.

Correct?

core/testsuite/src/test/java/com/blazebit/persistence/testsuite/Issue571Test.java
JWGmeligMeyling: Sure, just wasn't sure if we wanted the exception in the first place (i.e. because we don't exactly know how to properly detect it

entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/update/EntityViewUpdaterImpl.java
JWGmeligMeyling: Ok! Wasn't sure whether you'd rather remove `EntityMetamodel` or i.e. pass the `JpaProvider` rather than the `EVMI`.

core/testsuite/src/test/java/com/blazebit/persistence/testsuite/Issue571Test.java
beikov: That's ok, just add a comment noting this.

core/impl/src/main/java/com/blazebit/persistence/impl/AbstractCTECriteriaBuilder.java
JWGmeligMeyling: That uses the same logic unfortunately. I'll leave the check for what it is for now.

```
java.lang.NullPointerException
	at com.blazebit.persistence.impl.JpaUtils.getAttributeForJoining(JpaUtils.java:39)
	at com.blazebit.persistence.impl.AbstractCTECriteriaBuilder.prepareAndGetAttributes(AbstractCTECriteriaBuilder.java:247)
	at com.blazebit.persistence.impl.RecursiveCTECriteriaBuilderImpl.createCTEInfo(RecursiveCTECriteriaBuilderImpl.java:94)
	at com.blazebit.persistence.impl.CTEManager.onBuilderEnded(CTEManager.java:183)
	at com.blazebit.persistence.impl.RecursiveCTECriteriaBuilderImpl.onBuilderEnded(RecursiveCTECriteriaBuilderImpl.java:82)
	at com.blazebit.persistence.impl.SelectCTECriteriaBuilderImpl.end(SelectCTECriteriaBuilderImpl.java:47)
```

beikov: In that case you would have to call `implicitJoin(expression, true, null, ClauseType.SELECT, null, false, false, false, false)` before that.

core/parser/src/main/java/com/blazebit/persistence/parser/PathTargetResolvingExpressionVisitor.java
JWGmeligMeyling: I'm wondering whether its safe to assume that a field will exist with the attribute name, what if one uses method access rather than field access?

The current workaround looks at the member indicated by the attribute, and then looks up a similar typed member in the target class. Then for that member we resolve the actual type.

Now that I look into the code again, I do see though that the block has to be moved into the `if (baseClass != null) {` block.

core/impl/src/main/java/com/blazebit/persistence/impl/AbstractCTECriteriaBuilder.java
JWGmeligMeyling:
```java
                        joinManager.implicitJoin(pathExpression, true, null, ClauseType.SELECT, null, false, false, false, false);
                        AttributeHolder attributeForJoining = JpaUtils.getAttributeForJoining(getMetamodel(), pathExpression);

                        if (attributeForJoining.getAttributeType().getPersistenceType().equals(Type.PersistenceType.BASIC)) {
                            throw new IllegalStateException("An association should be bound to its association type and not its identifier type");
                        }
```

For example, fails on some occasions with:

```
java.lang.IllegalArgumentException: Attribute 'a' not found on type 'Issue571Test$MyEntity'
	at com.blazebit.persistence.parser.PathTargetResolvingExpressionVisitor.visit(PathTargetResolvingExpressionVisitor.java:224)
	at com.blazebit.persistence.parser.expression.PropertyExpression.accept(PropertyExpression.java:41)
	at com.blazebit.persistence.parser.PathTargetResolvingExpressionVisitor.visit(PathTargetResolvingExpressionVisitor.java:316)
	at com.blazebit.persistence.parser.expression.PathExpression.accept(PathExpression.java:74)
	at com.blazebit.persistence.impl.JpaUtils.getAttributeForJoining(JpaUtils.java:66)
	at com.blazebit.persistence.impl.JpaUtils.getAttributeForJoining(JpaUtils.java:57)
	at com.blazebit.persistence.impl.AbstractCTECriteriaBuilder.prepareAndGetAttributes(AbstractCTECriteriaBuilder.java:249)
	at com.blazebit.persistence.impl.RecursiveCTECriteriaBuilderImpl.createCTEInfo(RecursiveCTECriteriaBuilderImpl.java:94)
	at com.blazebit.persistence.impl.CTEManager.onBuilderEnded(CTEManager.java:183)
	at com.blazebit.persistence.impl.RecursiveCTECriteriaBuilderImpl.onBuilderEnded(RecursiveCTECriteriaBuilderImpl.java:82)
	at com.blazebit.persistence.impl.SelectCTECriteriaBuilderImpl.end(SelectCTECriteriaBuilderImpl.java:47)
	at com.blazebit.persistence.testsuite.Issue571Test$2.work(Issue571Test.java:94)
	at com.blazebit.persistence.testsuite.tx.TxSupport.transactional(TxSupport.java:33)
	at com.blazebit.persistence.testsuite.AbstractCoreTest.transactional(AbstractCoreTest.java:383)
	at com.blazebit.persistence.testsuite.Issue571Test.testBindingCteAssociationToEntity(Issue571Test.java:81)
```

I.e. it seems that I have to resolve some aliases manually? I'm honestly not sure whether we can *easily* strengthen the check, and I think the existing check will be sufficient 99% of the time. (I hardly imagine anyone every having a domain with a nested identifier with a poor naming scheme as `entity.id.id.id` where `id`'s are either associations or embeddables).

JWGmeligMeyling: Nevermind, 


```java
                       joinManager.implicitJoin(pathExpression, true, null, ClauseType.SELECT, null, false, false, false, false);
                        
                        if (pathExpression.getPathReference().getType().getPersistenceType().equals(Type.PersistenceType.BASIC)) {
                            throw new IllegalStateException("An association should be bound to its association type and not its identifier type");
                        }
```

Seems to do the trick.

core/parser/src/main/java/com/blazebit/persistence/parser/PathTargetResolvingExpressionVisitor.java
beikov: `com.blazebit.persistence.parser.util.JpaMetamodelUtils#resolveFieldClass` already does all the heavy lifting. The name is inappropriate now though since it also handles method members but the idea is that IMO the whole `getType` method can be reduced to the code I have shown.

PR #656, opened 2018-09-05
Make Blaze more Eclipse friendly

heruan:
## Description
This is to make Blaze code more Eclipse friendly, mainly:

- cast to raw types where the generic type isn't really needed (and Eclipse complains);
- add an m2eclipse execution entry for the `maven-antrun-plugin`.


beikov:
Moved commit to https://github.com/Blazebit/blaze-persistence/pull/651

PR #659, opened 2018-09-11
[#509, #669] Make use of Collection DML for collection flushers in updatable entity views

### LINKED ISSUE #434: Implement creatability validation
Currently, we don't validate if any entity could be successfully persisted based on non-nullable constraints, but we should. The validation should obviously consider excluded attributes as defined via `@CreatableEntityView`.
Also see #414 for details.

### LINKED ISSUE #509: Make use of collection management for updatable entity views in flushing

### LINKED ISSUE #517: Validate inverse remove strategy SET_NULL against column nullability
When having an updatable inverse collection, it might happen that the SET_NULL inverse remove strategy is used which won't work if the mapped by columns are non-nullable. This is especially problematic because SET_NULL is the default strategy. The current error a user sees is that a non-null constraint is violated which isn't very useful. This should be checked during metamodel building.

### LINKED ISSUE #658: Read-Only entity view subtypes are suddenly disallowed when a creatable entity view subtype exists
The problem is simple. When a creatable entity view subtype exists, and try to figure out allowed subtypes for an updatable attribute in an updatable entity view, we consider the attribute to also cascade persists for the creatable type. The problem now is, that when doing that, we also generate code that checks for allowed subtypes in the proxy. Unfortunately, we don't have a set of read-only subtypes that we also allow. This is missing and should be added.

### LINKED ISSUE #669: Ensure embeddable splitting is invoked for joined element collection aliases in group bys
@heruan reported that a query like `cb.groupBy("e.elementCollection")` doesn't work because the embeddable isn't split into it's component paths and Hibernate wrongly generates `alias.id` in the SQL.
We should ensure that embeddable splitting is properly invoked here as well. I guess the problem is just a wrong if condition that misses out on element collections.

### LINKED ISSUE #675: Split expression for group by into parts for DBMS that don't inline parameters
Currently, we assume when a DBMS supports complex group by expressions, that it also looks into the parameter values, but when we have an expression with a parameter, at least PostgreSQL can't see that the parameter value is the same as in the select clause resulting in an error. We have to introduce a DbmsDialect property and consider that during expression splitting for group by.

### LINKED ISSUE #679: Exception when saving a view which entity has domain events
### Description
When using Spring `@DomainEvents` annotated method in an entity, when saving an entity view for that entity via a Spring repository will fail since the annotated method is invoked with an instance of the view proxy, not the entity instance.

### Expected behavior
Domain events should be published or not, but not throw an exception.

### Actual behavior
An exception is thrown when trying to publish domain events.

### Steps to reproduce
1. Annotate an entity method with `@DomainEvents`, e.g.
```
@DomainEvents
public Collection<Object> domainEvents() {
    return Collections.emptyList();
}
```
2. Create an updatable view for that entity;
3. Create a repository for that view and try to save one.

### Environment

Version:            1.3.0
JPA-Provider:       Hibernate
DBMS:               PostgreSQL
Application Server: Spring Boot

beikov:
## Related Issue

Fixes #434, Fixes #509, Fixes #517, Fixes #669, Fixes #675, Fixes #679


integration/hibernate-base/src/main/java/com/blazebit/persistence/integration/hibernate/base/HibernateJpaProvider.java
```diff
@@ -213,6 +215,11 @@ public boolean supportsJoinElementCollectionsOnCorrelatedInverseAssociations() {
         return supportsJoinElementCollectionsOnCorrelatedInverseAssociations;
     }
 
+    @Override
+    public boolean supportsSingleValuedAssociationNaturalIdExpressions() {
+        return false;
```
JWGmeligMeyling: major >= 5, minor >= 4?

beikov: Was that merged already?

PR #665, opened 2018-09-25
[#667] Fix Spring Data 2.x compatibility and type variable related problems with entity views and Spring Data Repositories

### LINKED ISSUE #667: Work out what to do with embeddable types without a proper equals-hashCode implementation in updatable entity views
The problem is that updatable entity view objects are de-duplicated i.e. when joining collections, the result list tuples are flattened. The flattening can't happen unfortunately if there is an embeddable involved that doesn't implement equals-hashCode properly. There are probably some other collection duplication issues with that.
Since the symptom(exception that a parent was already set, duplicate elements in collections) says nothing about the actual cause, we have to check that somehow at validation time.
IMO we should require something explicit from the user to make sure he understands the implicaiton of using an embeddable type in an entity view.

Thoughts anyone?

beikov:
## Related Issue

Fixes #667


beikov:
Merged via 0e08aa81941a91b663fe1d8e2b72db6463ed47c5

PR #744, opened 2019-03-01
Kotlin QueryDSL extension methods [Work in progress]

jwgmeligmeyling:
Work in progress, feel free to review


querydsl/src/main/kotlin/com/blazebit/persistence/querydsl/Counter.kt
```diff
@@ -0,0 +1,30 @@
...
+package com.blazebit.persistence.querydsl
+
+import java.util.concurrent.atomic.AtomicInteger
+import kotlin.concurrent.getOrSet
+
+class Counter {
+    companion object {
+        val COUNTER = ThreadLocal<AtomicInteger>()
+
+        fun getCount() : Byte {
```
beikov: IMO we should somehow extend the parameter setting capabilities of the core-api to be able to register parameter values in an anonymous way also via e.g. RestrictionBuilder, so that we can avoid this. That way, we can also make sure parameter names are always the same.

JWGmeligMeyling: Yup this is a horrible hack :(

A problem that I do struggle with is that, even when I let the nested query builders like `WhereOrBuilder`, `RestrictionBuilder`, etc. extend `ParameterHolder`, I still don't have a mechanism to construct parameter names that I am binding.

That is because I seemingly can't set a parameter value before its actually used inside a parsed expression (this probably also determines the type of the parameter). So currently I do the following:

1. Convert the QueryDSL expression to a JPQL expression
2. Map the positional parameters into parameter names
3. Bind the parameters to the criteria builder

So actually I would need access to `ParameterManager#addPararmeterExpression(Object, ClauseType, AbstractCommonQueryBuilder)` (or at least the underlying mechanism of generating new parameter names), instead of just `ParameterHolder#setParameter(String, Object)`.

JWGmeligMeyling: I by the way thought to address that by introducing a `ParameterNameGenerator` or something. Not sure how to provide it though. I could make `ParameterHolder` extend it but it seems rather unuseful in the criteria builder API. So perhaps its more logical to provide it as a service. But then I need to make some additional interfaces extend `ServiceProvider`.

beikov: What you need is access to the `com.blazebit.persistence.impl.ParameterManager#addParameterExpression` method which is responsible for generating a parameter name for a parameter value.

JWGmeligMeyling: After constructing the JPQL fragment for the parsed expression I have collected the constant values for the parameters by index. However, because I haven't passed the expression over to the `CriteriaBuilder` yet, the parameters are not yet available. Binding the parameters just after setting the expression on the builder is not always sufficient too: not all expression setters actually end the builder, and only on builder end a listener is responsible for registering the expressions parameters.

So we have established that we need to add a parameter expression, or at least find a way to register a parameter in the `ParameterManager` and obtain the name under which the parameter is mapped. `ParameterManager#addParameterExpression` does roughly that.

In order to register the parameter though it seemed that I needed the `ClauseType` and `AbstractCommonQueryBuilder`. For some builders this seemed doable. I.e. a `RestrictionBuilderImpl` is aware of its `ClauseType`. But I don't seem to get a hold of the `AbstractCommonQueryBuilder`. After poking it for a while and investigating the internals of `ParameterImpl` I realised that I could simply provide a `null` value for `ClauseType` and `AbstractCommonQueryBuilder`. The two would be registered with the parameter as soon as the builder end listener would do its work anyway.

Now I needed a way to make `addParameterExpression(String, null, null)` available through the API somehow. My first intuition was to extend `ParameterHolder` and make the other builders extend this interface. I soon realised that this was probably not ideal:

- Extending `ParameterHolder` from builders would make other parameter related methods available in these builders. But after correctly implementing these based on the implementation from `AbstractQueryBuilder`, I realised that the parameters created within the builder cannot be set, because they haven't been registered by the listener yet. Unless that would be addressed this would lead to confusion, because the only parameters you'll likely want to set there, will throw an exception. I'm also not sure if this is even something that can be addressed, the code seems to quite deliberately want to finish the expression before registering all parameters and such.
- Extending `ParameterHolder` by adding a method that creates a parameter under a generated name, will also have to return that parameter name. As such it cannot return the current builder and will not follow the builder pattern. This seems ugly.

I considered that making a part of the `ParameterManager` available through the `ServiceProvider` (that I needed to obtain the metamodel through the `EntityManager` anyway), would be the better option. So I created a [`ConstantRegistry`](https://github.com/Blazebit/blaze-persistence/pull/744/files#diff-bda83653fe4dff46a59f0bea6553d2d3) interface (not particularly fond of the naming BTW), which `ParameterManager` implements and made it available as a service.

That all works quite fine, but now my extension methods have an even stronger dependency to types being a `ServiceProvider` (or at least being able to resolve one for a given builder). Effectively this means that my extension methods only work on the root query builder (which happens to implement `ServiceProvider`), and for some builders I provide delegates that keep track of a `ServiceProvider`. This however gets pretty tedious real quick for nested calls. It would be more ideal if the builders just all had the ability to lookup services.  

Therefore I started investigating how I could implement `ServiceProvider` on `BaseHavingBuilder`, `BaseWhereBuilder`, `RestrictionBuilderImpl`, etc. Actually, a provider for my `ConstantRegistry` should be pretty easy, as these builders all comprise a `ParameterHolder` reference anyway. However, obtaining the `EntityManager` is a whole other story. Besides that, I also think that it would be logical to provide all services for the current `CriteriaBuilder` rather than just 1 or 2 services supported by the sub-builder. So I investigated how I could propagate the service lookup to the parent builders.

This got ugly pretty fast. Currently builders do not have a reference to the root builder, usually only to their parent builder and even then that link is not obtainable through the API (for obvious reasons). I seemed to need a lot of signature changes and introduction of extra parameters. Very unpreferable stuff.

Am I overseeing something here? I've seen you use `subqueryInitFactory.getQueryBuilder()` in builders to obtain the `QueryBuilder` too. As the related statements seemingly have little to do with subqueries, it appears to me as if this is a more or less *accepted* workaround for the same issue? So would `subqueryInitFactory.getQueryBuilder().getService(x)` be a logical delegate for these builders?

I've updated my PR with the newest implementation.

beikov: I'm undecided about whether we really should expose the fact that everything extends `ServiceProvider`. I'm thinking about just letting the implementations implement the interface and do the casting in the extension functions to avoid _polluting_ the IDE suggestions. Wdyt?

JWGmeligMeyling: We could definitely hide it and do the casting internally. I am however afraid that this could lead to maintainability issues, such as extension methods being available but throwing a `CastClassException` upon use. I also think that having a mechanism to unwrap some dependencies on a particular builder could be a nice extension hook for any future extensions. This way we have at least limited the number of methods added to the bare minimum (one).Its mostly the name of the interface that I have trouble with, as a builder *is* not itself a `ServiceProvider` but instead is able to unwrap objects served through a service provider. That distinction may lead to unexpected API use. Not sure how things trade off, WDYT? I'd happily hide the interface or introduce a new one.

JWGmeligMeyling:
I have added extension methods for case builders too, something that seemed quite useful. I am at 75% method coverage now, main parts are missing are the tests for the more subtle cases of CTE queries, FROM clauses and manipulation queries. I'll keep extending the PR and I hope to wrap it up by the end of the week. That will also require me to look into some DBMS specific failures (or perhaps I simply decide not to actually execute the query and just look at the produced JPQL+ instead).

Anyways, as I am trying to wrap things up it would be nice if we decide on whether or not we want to expose the `ServiceProvider` (or any other common interface between builder types that is able to unwrap some underlying context), or whether I have to do some casts instead.

beikov:
I'd personally rather see the implementations implement the `ServiceProvider` interface directly and the extension methods doing unconditional casting. I'd rather not expose the fact that every builder implements the `ServiceProvider` interface through the public APIs.

JWGmeligMeyling:
I see the point of not exposing `ServiceProvider` through the API, but if we really discourage such use of the builder API, then whats the point of implementing `ServiceProvider` on the implementations and using that as backing mechanism for the extension methods? Is that primarily from a "experimental API" point-of-view? Anyways, I've been trying to wrap my head around it and I don't really like the idea (but perhaps you can change my mind). I am just worried about suggesting extension methods for `RestrictionBuilders` (for example) that do not actually implement `ServiceProvider` and as a result fail to bind parameters (either by a parameter not bound exception or a `CastClassException`, depending on where we let this fail).

Obviously the easiest change from here is to just move the `ServiceProvider` from the interfaces to the implementations and add the cast within the extension methods. But if we really want to discourage that API use, wouldn't it be better to create a wrapper criteria builder interface (similarly to our current jpa-criteria module) instead?

beikov:
I just don't think it's very nice when an end-user of the builder API sees methods suggested by the IDE that are only relevant for integrations like e.g. the extension methods. The exceptions you mention would be bugs as all implementations should implement the `ServiceProvider` interface if we do it like this.
I'm fine with exposing it if you think it's ok, but having a wrapper is IMO an overkill.

jwgmeligmeyling:
This PR is superseded by #1071

Issue #493, opened 2017-12-05
Add epoch based full temporal unit extraction functions


beikov:
Currently we have functions like e.g. `hour` that extract the _hour part_ of a temporal value i.e. produces 3 for the interval `2 days 3 hours`. Oftentimes it is necessary to extract the full hours i.e. 51 which can be done by extracting the epoch and then dividing it to the desired unit. Some DBMS(e.g. Oracle, DB2) might not have efficient epoch extraction functions so letting the user do the calculation based on the epoch might not be very nice performance-wise.

I propose we add the following functions, analogous to the _part extract_ function:

* epoch_microseconds - `epoch + microseconds` or native function
* epoch_milliseconds - `epoch + milliseconds` or native function
* epoch_seconds - alias for `epoch`
* epoch_days - `epoch / 86400‬` or native function

I'm not sure anymore if adding the following functions makes sense as I don't see use cases and AFAICS there is no native implementation possible that would be any better. In addition, I'm unsure if the definitions of week, month, quarter and year is very satisfying for users.

* epoch_minutes - `epoch / 60`
* epoch_hours - `epoch / 3600‬`
* epoch_weeks - `epoch / 604800` ?
* epoch_months - `epoch / 18144000‬` ?
* epoch_quarters - `epoch / 54432000` ?
* epoch_years - `epoch / 217 728 000‬` ?

beikov:
Related to #380

PR #746, opened 2019-03-03
[#380] [#524] Add missing date functions

### LINKED ISSUE #380: Add more temporal related functions
We are still lacking some date extract functions
* `microsecond`
* `millisecond`
* `week`
* `quarter`
* `dow` - day of week 0 (Sunday) - 6 (Saturday)
* `doy` - day of year (extract(day from ? - trunc_year(?)))

Also need some more diff functions
* `epoch_diff`
* `microsecond_diff`
* `week_diff`
* `quarter_diff`

Also add tests to make sure the diffing functions produce correct results i.e. a day_diff between _01-01-2001_ and _01-01-2002_ produces 365/366 days.

Something that we currently don't have but would be quite useful for grouping is a date truncation i.e. `trunc_hour(...)` should truncate off the minutes, seconds and fractional part.

Add support for the following functions
* `trunc_microseconds`
* `trunc_milliseconds`
* `trunc_second`
* `trunc_minute`
* `trunc_hour`
* `trunc_day`
* `trunc_week`
* `trunc_month`
* `trunc_quarter`
* `trunc_year`

PostgreSQL has these functions builtin and should serve as reference for how to implement the functions: https://www.postgresql.org/docs/9.4/static/functions-datetime.html

### LINKED ISSUE #494: Extend temporal extract/diff functions
We are still lacking some temporal functions that should be implemented.

Extraction functions
* microsecond
* millisecond
* week
* quarter
* dow - day of week 0 (Sunday) - 6 (Saturday)
* doy - day of year (extract(day from ? - trunc_year(?)))

Diffing functions
* microsecond_diff
* week_diff
* quarter_diff

Also add tests to make sure the diffing functions produce correct results i.e. a day_diff between _01-01-2001_ and _01-01-2002_ produces 365/366 days.

### LINKED ISSUE #524: Add temporal functions that allow adding a time unit to a timestamp
It is sometimes a lot better to do something like `timestamp + 3 hours > ?` rather than `? - timestamp > 3` but currently we only have _diff_ functions which allow to get the difference in a unit between two timestamps. We should add functions that allow adding units to a timestamp as well.
I propose we add the following

* add_microseconds
* add_milliseconds
* add_second
* add_minute
* add_hour
* add_day
* add_week
* add_month
* add_quarter
* add_year

The functions take a timestamp and an integer value representing the value to be added as unit as defined by the function name. The return type is a timestamp. The value obviously can be negative which allows to subtract a value.

### LINKED ISSUE #745: Add missing date extraction functions
During testing I noticed 
* milliSecond
* week
* day of week, day of year

Are missing as date extraction functions.

We also seem to miss the date trunc functions. Would be nice to add these as well.

Week diff seems to miss as well.

### Environment

Version:            
JPA-Provider:       
DBMS:               
Application Server:

JWGmeligMeyling:
## Description

Mostly done. The missing parts are

- [x] Date extract `WEEK_IN_YEAR` + documentation that this might be more natural
- [x] Date extract `YEAR_OF_WEEK` + documentation. Possibly gives previous year `2000-01-01` is 1999
- [x] Date extract documentation regarding ISO standard and possibly surprising results of `WEEK`
- [x] Date truncation documentation
- [x] `YEAR_WEEK` function that produces the string `YYYY-WW` from a timestamp
- [x] epoch_diff
- [x] microsecond_diff
- [x] week_diff
- [x] quarter_diff
- [x] Date diff documentation update
- [x] add_microseconds
- [x] add_milliseconds
- [x] add_second
- [x] add_minute
- [x] add_hour
- [x]  add_day
- [x] add_week
- [x] add_month
- [x] add_quarter
- [x] add_year
- [x] add_xxx documentation update
- [x] epoch_microseconds - `epoch + microseconds` or native function (#493)
- [x] epoch_milliseconds - `epoch + milliseconds` or native function (#493)
- [x] epoch_seconds - alias for `epoch` (#493)
- [x] epoch_days - `epoch / 86400‬` or native function (#493)

## Related Issue

Fixes #380 
Fixes #494
Fixes #524
FIxes #745


core/impl/src/main/java/com/blazebit/persistence/impl/function/datetime/millisecond/DB2MillisecondFunction.java
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ *
+ * @author Jan-Willem Gmelig Meyling
+ * @since 1.4.0
+ */
+public class DB2MillisecondFunction extends MillisecondFunction {
+
+    public DB2MillisecondFunction() {
+        super("MICROSECOND(?1)");
```
beikov: Shouldn't you also divide the result by 1000 to get milliseconds?

JWGmeligMeyling: Ah yes, I got confused by the DB2 doc. 😄 Now I am reading it again, I see that it indeed has to be divided by 1000.

I'll finish up the PR by adding some tests. I also intend to add date_trunc functions which I quite heavily use.

beikov: Sounds great, I'll wait for these changes then 👍

core/impl/src/main/java/com/blazebit/persistence/impl/function/datetime/dayofyear/DB2DayOfYearFunction.java
```diff
@@ -0,0 +1,30 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.blazebit.persistence.impl.function.datetime.dayofyear;
+
+import com.blazebit.persistence.impl.function.datetime.week.WeekFunction;
```
beikov: Unused import

JWGmeligMeyling:
> Do you want to work on the diff and trunc function implementations as well or just these?

Yes, I'll probably keep extending the PR but just want to make sure I've covered every dialect properly for these changes first.

beikov:
Ok, you will probably have to apply the following as well to fix the build: https://github.com/Blazebit/blaze-persistence/pull/829/commits/1e99c925ed6c04a61e598972af834c1c99adbc5f

JWGmeligMeyling:
I'll rebase on 1e99c92 then 😄

JWGmeligMeyling:
I added `date_trunc` functions to the PR. All tests pass but the DB2 ones that halt : https://travis-ci.org/Blazebit/blaze-persistence/jobs/575628343 . No idea what is going on there :/

beikov:
I think the problem was that the DB2 connection acquisition takes so long that we ran into a Travis CI timeout because we had no output for more than 10 minutes. I fixed that now by improving the test performance.

JWGmeligMeyling:
Let's see what it does 😄

JWGmeligMeyling:
There is something not entirely right with the timezones it seems:

(under PostgreSQL)

```
Failed tests: 
1697  DateDiffTest.testTimestampWithDateDiff:235 expected:<-17544> but was:<-17550>
1698  DateDiffTest.testTimestampWithParameterDiff:285 expected:<397> but was:<396>
1699  ValuesClauseTest.testValuesEntityFunctionLikeCalendar:208 expected:<java.util.GregorianCalendar[time=?,areFieldsSet=false,areAllFieldsSet=true,lenient=true,zone=sun.util.calendar.ZoneInfo[id="UTC",offset=0,dstSavings=0,useDaylight=false,transitions=0,lastRule=null],firstDayOfWeek=1,minimalDaysInFirstWeek=1,ERA=1,YEAR=2019,MONTH=7,WEEK_OF_YEAR=34,WEEK_OF_MONTH=4,DAY_OF_MONTH=23,DAY_OF_YEAR=235,DAY_OF_WEEK=6,DAY_OF_WEEK_IN_MONTH=4,AM_PM=0,HOUR=9,HOUR_OF_DAY=0,MINUTE=0,SECOND=0,MILLISECOND=0,ZONE_OFFSET=0,DST_OFFSET=0]> but was:<java.util.GregorianCalendar[time=1566496800000,areFieldsSet=true,areAllFieldsSet=true,lenient=true,zone=sun.util.calendar.ZoneInfo[id="UTC",offset=0,dstSavings=0,useDaylight=false,transitions=0,lastRule=null],firstDayOfWeek=1,minimalDaysInFirstWeek=1,ERA=1,YEAR=2019,MONTH=7,WEEK_OF_YEAR=34,WEEK_OF_MONTH=4,DAY_OF_MONTH=22,DAY_OF_YEAR=234,DAY_OF_WEEK=5,DAY_OF_WEEK_IN_MONTH=4,AM_PM=1,HOUR=6,HOUR_OF_DAY=18,MINUTE=0,SECOND=0,MILLISECOND=0,ZONE_OFFSET=0,DST_OFFSET=0]>
```

https://travis-ci.org/JWGmeligMeyling/blaze-persistence/jobs/575740890

beikov:
Let's see if it works now. I think I found the issue.

JWGmeligMeyling:
> * [ ]  Date extract `WEEK_IN_YEAR` + documentation that this might be more natural

How do you want to implement this one? `DAY_IN_YEAR % 7 + 1` ?

> * [ ]  Date extract `YEAR_OF_WEEK` + documentation. Possibly gives previous year `2000-01-01` is 1999

This one is going to be tricky, so far I've found an approach for one database.

> * [ ]  Date extract documentation regarding ISO standard and possibly surprising results of `WEEK`

I can add this.

> * [ ]  Date truncation documentation

I can add this.

> * [ ]  `YEAR_WEEK` function that produces the string `YYYY-WW` from a timestamp

I can add this.

> * [ ]  epoch_diff
> * [ ]  microsecond_diff
> * [ ]  week_diff
> * [ ]  quarter_diff
> * [ ]  Date diff documentation update

I'm personally not so invested in having diff functions, because I just subtract timestamps from each other and have a custom type that maps a `PGInterval` to a java `Interval` type which gives me a lot more clarity about the time difference (i.e. is it millis? seconds?).

beikov:
> How do you want to implement this one? DAY_IN_YEAR % 7 + 1 ?

beikov:
> Date extract WEEK_IN_YEAR + documentation that this might be more natural
> How do you want to implement this one? DAY_IN_YEAR % 7 + 1 ?

That's one possibility, but there are database functions that support that as well for some DBMS.

> Date extract YEAR_OF_WEEK + documentation. Possibly gives previous year 2000-01-01 is 1999
> This one is going to be tricky, so far I've found an approach for one database.

If the DBMS doesn't support it, we could do `SUBSTRING(YEAR_WEEK(...), 1, 4)`

> I'm personally not so invested in having diff functions, because I just subtract timestamps from each other and have a custom type that maps a PGInterval to a java Interval type which gives me a lot more clarity about the time difference (i.e. is it millis? seconds?).

I understand that, but blaze-persistence currently has no way to abstract intervals yet. Also, I'm not sure if we can do it for all DBMS, so having this as fallback is nice. On top of that, I'm not sure if a plain interval can provide all the same details. Just ask yourself if a 30 day interval is a month? That depends, but can be answered clearly when considering two timestamps which is why these functions matter.

JWGmeligMeyling:
MySQL has various modes for `WEEK()`, but none seems to match the idea behind `WEEK_IN_YEAR`: https://dev.mysql.com/doc/refman/5.5/en/date-and-time-functions.html#function_week . I'm not at all sure whether there is any other kind of week we could support besides ISOWEEK that would make sense as a standard, unless we add an option to specify the mode.  Porting all the various modes to different dialects though may require us to create quite extensive custom functions.  We could use `ROUND(DAY_IN_YEAR / 7, 0)` as implementation for `WEEK_IN_YEAR`, but I don't think this has any actual sensible meaning.

I haven't found many implementations for `YEAR_OF_WEEK` or `YEAR_WEEK` in databases, but I think we should just truncate the date by week and then extract / format the result.

`epoch_diff` and `microsecond_diff` are trivial.

`week_diff`: would you expect `ROUND(days_diff / 7, 0)` or the number of iso weeks in between (in which case we probably end up having to to account for the difference in years separately)?

`quarter_diff`: I think we agree that this should be `ROUND(month_diff / 3, 0)` ? (3 months in every of the 4 quarters).

Equivalently, should `add_weeks` (#524) add 7 days?  And should `add_quarter` then add 3 months? Or should these methods truncate to the start of the iso week or quarter, and then add the interval?

beikov:
> ROUND(DAY_IN_YEAR / 7, 0)

Sounds good, but does `ROUND(1 / 7, 0)` produce 1?

> YEAR_OF_WEEK

Maybe do something like `YEAR(TIMESTAMP - 6 days)`?

> YEAR_WEEK

Maybe do something like `YEAR(TIMESTAMP - 6 days) || WEEK(TIMESTAMP)`?

> week_diff: would you expect ROUND(days_diff / 7, 0) or the number of iso weeks in between (in which case we probably end up having to to account for the difference in years separately)?

IMO we should follow the standard and return the full weeks between timestamps. Not sure if it makes sense to have a separate function that returns +1 when crossing a week boundary.

> quarter_diff: I think we agree that this should be ROUND(month_diff / 3, 0) ? (3 months in every of the 4 quarters).

Again, yes. Only full quarters are counted.

> Equivalently, should add_weeks (#524) add 7 days? And should add_quarter then add 3 months? Or should these methods truncate to the start of the iso week or quarter, and then add the interval?

Good question. PostgreSQL doesn't allow quarter in intervals, so I think we might better stay away from offering that functionality. IMO we shouldn't add support for add_quater. Adding a week should add 7 days. For all other arithmetic, one can first truncate and then add the amount that is needed.

beikov:
On second thought, maybe let's define `add_quarter` as adding 3 months i.e. `add_quarter('2000-02-28', 1)` will produce `'2000-05-28'`

JWGmeligMeyling:
I don't think `YEAR(TIMESTAMP - 6 days)` is reliable if the year starts on the start of a week, and that day is the value used  for timestamp.

beikov:
Maybe `YEAR(TRUNC_WEEK(TIMESTAMP))`?

JWGmeligMeyling:
I think thats the way to go yes👍🏻

JWGmeligMeyling:
Added some more methods, made some more tests pass. Still struggling with DB2, but going stronger now. Lets see what CI says tomorrow morning. It seems there's only date diff to go.

beikov:
MySQL only allows the MICROSECONDS unit, no MILLISECONDS: https://travis-ci.org/Blazebit/blaze-persistence/jobs/582965103#L1543 also see https://dev.mysql.com/doc/refman/5.5/en/date-and-time-functions.html#function_timestampadd

PostgreSQL invocations are failing because you are missing single quotes: https://travis-ci.org/Blazebit/blaze-persistence/jobs/582965105#L1576 Adding units should look like `?1 + interval ?2 || ' day'`

SQLServer doesn't allow plurals for the units https://travis-ci.org/Blazebit/blaze-persistence/jobs/582965106#L1555 also see https://docs.microsoft.com/de-de/sql/t-sql/functions/dateadd-transact-sql?view=sql-server-2017

AFAIU DB2 always requires pluralized units: https://www.ibm.com/developerworks/community/blogs/SQLTips4DB2LUW/entry/dateadd?lang=en

The `DB2YearWeekFunction` should use the concat operator `||` instead to concat the elements.

Oracle tests are failing because you are missing single quotes around the second argument: https://travis-ci.org/Blazebit/blaze-persistence/jobs/582965108#L1945 Adding units should look like `?1 + interval ?2 || '' day`

Not sure why EclipseLink is failing, but I'll look into that.

More comments inline

JWGmeligMeyling:
Hmm the postgres tests are too bad. Newer versions do support this syntax. The str concatenation requires casting for some parameter types. Anyways, lets deal with that now 🙈

Thanks for looking into all the test failures!

core/impl/src/main/java/com/blazebit/persistence/impl/function/datetime/yearweek/H2YearWeekFunction.java
```diff
@@ -0,0 +1,29 @@
...
+package com.blazebit.persistence.impl.function.datetime.yearweek;
+
+/**
+ * @author Jan-Willem Gmelig Meyling
+ * @since 1.4.0
+ */
+public class H2YearWeekFunction extends YearWeekFunction {
+
+    public H2YearWeekFunction() {
+        super("CONCAT(extract(year from DATE_TRUNC('week', ?1)), '-', iso_week(?1))");
```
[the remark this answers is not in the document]
JWGmeligMeyling: Its a workaroud commonly found on StackOverflow for missing date trunc in MSSQL and MySQL, that I reused here. Ill add a reference to date trunc.

core/impl/src/main/java/com/blazebit/persistence/impl/function/datetime/yearweek/DB2YearWeekFunction.java
```diff
@@ -0,0 +1,29 @@
...
+package com.blazebit.persistence.impl.function.datetime.yearweek;
+
+/**
+ * @author Jan-Willem Gmelig Meyling
+ * @since 1.4.0
+ */
+public class DB2YearWeekFunction extends YearWeekFunction {
+
+    public DB2YearWeekFunction() {
+        super("CONCAT(YEAR(DATE_TRUNC('week', ?1)), '-', WEEK_ISO(?1))");
```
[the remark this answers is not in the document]
JWGmeligMeyling: Ah yes, thats much  better

core/impl/src/main/java/com/blazebit/persistence/impl/function/dateadd/DateAddFunction.java
```diff
@@ -0,0 +1,65 @@
...
+        return firstArgumentType;
+    }
+
+    @Override
+    public void render(FunctionRenderContext context) {
+        if (context.getArgumentsSize() != 2) {
+            throw new RuntimeException("The " + functionName + " function needs exactly two arguments <start_datetime> and <interval>! args=" + context);
+        }
+
+        renderDiff(context);
```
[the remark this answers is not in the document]
JWGmeligMeyling: Some dialects actually require the interval to come first, others need the timestamp as first argument, so the swapped arguments is intentional here. Or do you mean something else?

Why are arguments with parameters not allowed? Do you have an example?

core/impl/src/main/java/com/blazebit/persistence/impl/function/datetime/yearweek/SQLServerYearWeekFunction.java
```diff
@@ -23,7 +23,7 @@
 public class SQLServerYearWeekFunction extends YearWeekFunction {
 
     public SQLServerYearWeekFunction() {
-        super("CONCAT(datepart(yy, CONVERT(date, DATEADD(WEEK, DATEDIFF(WEEK, 0, ?1), 0)), '-', DATEPART(isowk, CONVERT(date, ?1)))");
+        super("CONCAT(datepart(yy, CONVERT(date, DATEADD(WEEK, DATEDIFF(WEEK, 0, ?1), 0))), '-', DATEPART(isowk, CONVERT(date, ?1)))");
```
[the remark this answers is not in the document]
beikov: Also see com.blazebit.persistence.impl.function.datetime.epoch.OracleEpochFunction

beikov:
IMO we should also look into adapting the epoch functions of Oracle.

A new Oracle implementation could look like `cast(cast(?1 as timestamp) - to_date('1970-01-01', 'yyyy-mm-dd') as int) * 86400`

Although I am not sure how this works regarding TZs, it might be worth looking into as the current implementation is pretty complex.

core/impl/src/main/java/com/blazebit/persistence/impl/function/datetime/yearweek/DB2YearWeekFunction.java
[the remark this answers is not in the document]
JWGmeligMeyling:
```suggestion
        super("VARCHAR_FORMAT('IYYYY-IW', ?1)'");
```

core/impl/src/main/java/com/blazebit/persistence/impl/function/datetime/yearweek/H2YearWeekFunction.java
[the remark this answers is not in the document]
JWGmeligMeyling:
```suggestion
        super("(select CONCAT(extract(year from DATE_TRUNC('week', x.y)), '-', iso_week(x.y)) from (values (?1)) x(y))");
```

core/impl/src/main/java/com/blazebit/persistence/impl/function/datetime/yearweek/SQLServerYearWeekFunction.java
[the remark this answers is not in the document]
JWGmeligMeyling:
```suggestion
        super("(select CONCAT(datepart(yy, CONVERT(date, DATEADD(WEEK, DATEDIFF(WEEK, 0, x.y), 0))), '-', DATEPART(isowk, CONVERT(date, x.y))) from (values (?1)) x(y))");
```

core/impl/src/main/java/com/blazebit/persistence/impl/function/trunc/week/MySQLTruncWeekFunction.java
```diff
@@ -0,0 +1,29 @@
...
+package com.blazebit.persistence.impl.function.trunc.week;
+
+/**
+ * @author Jan-Willem Gmelig Meyling
+ * @since 1.4.0
+ */
+public class MySQLTruncWeekFunction extends TruncWeekFunction {
+
+    public MySQLTruncWeekFunction() {
+        super("date_add('1900-01-01', interval TIMESTAMPDIFF(WEEK, '1900-01-01', ?1) WEEK)");
```
JWGmeligMeyling:
```suggestion
        // Implementation from https://stackoverflow.com/a/32955740/2104280
        super("date_add('1900-01-01', interval TIMESTAMPDIFF(WEEK, '1900-01-01', ?1) WEEK)");
```

core/impl/src/main/java/com/blazebit/persistence/impl/function/trunc/week/MSSQLTruncWeekFunction.java
```diff
@@ -0,0 +1,29 @@
...
+package com.blazebit.persistence.impl.function.trunc.week;
+
+/**
+ * @author Jan-Willem Gmelig Meyling
+ * @since 1.4.0
+ */
+public class MSSQLTruncWeekFunction extends TruncWeekFunction {
+
+    public MSSQLTruncWeekFunction() {
+        super("DATEADD(WEEK, DATEDIFF(WEEK, 0, ?1), 0)");
```
JWGmeligMeyling:
```suggestion
        // Implementation from https://stackoverflow.com/a/32955740/2104280
        super("DATEADD(WEEK, DATEDIFF(WEEK, 0, ?1), 0)");
```

core/impl/src/main/java/com/blazebit/persistence/impl/function/datetime/yearofweek/MySQLYearOfWeekFunction.java
```diff
@@ -0,0 +1,29 @@
...
+package com.blazebit.persistence.impl.function.datetime.yearofweek;
+
+/**
+ * @author Jan-Willem Gmelig Meyling
+ * @since 1.4.0
+ */
+public class MySQLYearOfWeekFunction extends YearOfWeekFunction {
+
+    public MySQLYearOfWeekFunction() {
+        super("extract(YEAR FROM date_add('1900-01-01', interval TIMESTAMPDIFF(WEEK, '1900-01-01', ?1) WEEK))");
```
JWGmeligMeyling:
```suggestion
        // Extract year from truncated time, derived from https://stackoverflow.com/questions/5541326/postgresqls-date-trunc-in-mysql
        super("extract(YEAR FROM date_add('1900-01-01', interval TIMESTAMPDIFF(WEEK, '1900-01-01', ?1) WEEK))");
```

JWGmeligMeyling:
> IMO we should also look into adapting the epoch functions of Oracle.
> 
> A new Oracle implementation could look like `cast(cast(?1 as timestamp) - to_date('1970-01-01', 'yyyy-mm-dd') as int) * 86400`

This would only consider days and disregard all time-based units right?

> 
> Although I am not sure how this works regarding TZs, it might be worth looking into as the current implementation is pretty complex.

I think it works if Oracle converts the timestamp to ISO on a cast to any timestamp without timezone, which I think it does. If not, then the implementation we're trying to replace also does not consider different TZs either.

beikov:
> This would only consider days and disregard all time-based units right?

The cast should be the outermost part `cast((cast(?1 as timestamp) - to_date('1970-01-01', 'yyyy-mm-dd')) * 86400 as int)`

documentation/src/main/asciidoc/core/manual/en_US/18_jpql_functions.adoc
```diff
@@ -240,14 +240,64 @@ Extracts the requested field of temporal argument like specified by the ANSI SQL
...
+* *Microsecond* - `MICROSECOND`
+* *Millisecond* - `MILLISECOND`
 * *Second* - `SECOND`
 * *Minute* - `MINUTE`
 * *Hour* - `HOUR`
 * *Day* - `DAY`
+* *Day of week* - `DAYOFWEEK` -  an integer between 1 (Monday) and 7 (Sunday)
+* *Day of year* - `DAYOFYEAR`
+* *Week* - `WEEK` or `ISO_WEEK` - an integer between 1 and 53
+* *Year week* - `YEAR_WEEK` - a string with the forma `IYYY-IW`
```
beikov:
```suggestion
* *Year week* - `YEAR_WEEK` - a string with the format `IYYY-IW`
```

core/impl/src/main/java/com/blazebit/persistence/impl/function/datetime/yearweek/H2YearWeekFunction.java
```diff
@@ -0,0 +1,29 @@
...
+package com.blazebit.persistence.impl.function.datetime.yearweek;
+
+/**
+ * @author Jan-Willem Gmelig Meyling
+ * @since 1.4.0
+ */
+public class H2YearWeekFunction extends YearWeekFunction {
+
+    public H2YearWeekFunction() {
+        super("(select CONCAT(extract(year from DATE_TRUNC('week', x.y)), '-', iso_week(x.y)) from (values (?1)) x(y))");
```
beikov: Unfortunately, H2 doesn't support that syntax. In DB2 this works by using `from lateral (values (?1))` but I doubt this will work here. Can you think of any other SQL that might be able to produce the same result while rendering the template argument only once? If not, we will have to bail if encountering a parameter.

beikov:
Also, please remove unnecessary imports in `EclipseLinkFunctionRenderContext` to resolve the checkstyle issue

JWGmeligMeyling:
I should have addressed the test failures and your review comments now.

What remains is a couple of diff methods:

- [ ] epoch_diff
- [ ] microsecond_diff
- [ ] week_diff
- [ ] quarter_diff
- [ ] Date diff documentation update

And if we decide to do the epoch functions too:

- [ ] epoch_microseconds - `epoch + microseconds` or native function (#493)
- [ ] epoch_milliseconds - `epoch + milliseconds` or native function (#493)
- [ ] epoch_seconds - alias for `epoch` (#493)
- [ ] epoch_days - `epoch / 86400‬` or native function (#493)

I'll first resolve any pending issues in the existing tree. It would be nice to finally see the suite pass on commit 0df78d2. I am a bit occupied for the rest of the day, so it will probably be tomorrow before I can look into the other methods.

JWGmeligMeyling:
Ah damnit, the CI PostgreSQL is too old for the alternative syntax too... That sucks

beikov:
> I'll first resolve any pending issues in the existing tree. It would be nice to finally see the suite pass on commit 0df78d2. I am a bit occupied for the rest of the day, so it will probably be tomorrow before I can look into the other methods.
I think the main issues are fixed now, so we can focus on the remaining functions.

> Are you sure that timestamp subtraction yields a double that is precise to the millisecond?
According to Oracle docs it does and is apparently the only viable way AFAICT.

> For example in PostgreSQL the statement would return the difference as days in an integer value. And do we trust that no rounding errors occur here that affect the outcome?
Not sure, I guess we have to try, but the current implementation extracts all fields separately and adds them up which is very inefficient.

JWGmeligMeyling:
## `epoch_diff`:
I am pretty sure that this should be an alias for `second_diff`. In fact, we depend on epoch extraction in the implementation of second diff for a couple of implementations, i.e. : `(select cast(trunc(EXTRACT(EPOCH FROM t2::timestamp)) - trunc(EXTRACT(EPOCH FROM t1::timestamp)) as integer) from (values (?1, ?2)) as temp(t1,t2))`.

## `microsecond_diff`
For databases that do not support `DATE_DIFF`, we can multiply the epoch diff by 1M, as long as the epoch diff is returned in a floating point value rather than a rounded integer. Otherwise we might be able to extract the microseconds part from the DBMS native interval type and add this to the result of millisecond_diff. It seems that the Microsecond diff function for Oracle will  become particularly nasty, as the ones for second and millisecond are already quite sophisticated.

## `week_diff`
This is pretty well defined as `floor(day_diff/7)`. I think we should not even bother with DBMS native implementations. For example, when extracting the weeks from a PostgreSQL interval, I had to compensate for the year_diff part again. Knowing our day_diff implementation is sane, the division and floor operation will be a relatively easy modification.

## `quarter_diff`
Same as week_diff, but then `floor(month_diff/3)`

##  `epoch_microseconds`
The current epoch function returns an `Integer`, but for timestamps with high enough precision, most DBMSes will return a float. Therefore, this should be equivalent to our implementation for `epoch * 1M`.

# `epoch_milliseconds`
The current epoch function returns an `Integer`, but for timestamps with high enough precision, most DBMSes will return a float. Therefore, this should be equivalent to our implementation for `epoch * 1M`.

beikov:
> It seems that the Microsecond diff function for Oracle will become particularly nasty, as the ones for second and millisecond are already quite sophisticated.

That's why I proposed to adapt the Oracle implementation. I think the following should do the same as the previous implementation: `cast((cast(?1 as timestamp) - to_date('1970-01-01', 'yyyy-mm-dd')) * 86400 as int)` although we might want to cast to bigint.

I agree with the rest regarding diff functions.

The epoch stuff should use bigint and Long in Java.

core/impl/src/main/java/com/blazebit/persistence/impl/function/datediff/microsecond/MicrosecondDiffFunction.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+public abstract class MicrosecondDiffFunction extends DateDiffFunction {
+
+    public MicrosecondDiffFunction(String template) {
+        super("millisecond_diff", template);
+    }
+
+    @Override
+    public Class<?> getReturnType(Class<?> firstArgumentType) {
+        return BigInteger.class;
```
beikov: Should use `Long.class`

core/impl/src/main/java/com/blazebit/persistence/impl/function/datediff/week/WeekDiffFunction.java
```diff
@@ -0,0 +1,30 @@
...
+import com.blazebit.persistence.impl.function.datediff.DateDiffFunction;
+
+/**
+ * @author Jan-Willem Gmelig Meyling
+ * @since 1.4.0
+ */
+public abstract class WeekDiffFunction extends DateDiffFunction {
+
+    public WeekDiffFunction(String template) {
+        super("day_diff", template);
```
beikov: day_diff?

core/impl/src/main/java/com/blazebit/persistence/impl/function/datediff/microsecond/MySQLMicrosecondDiffFunction.java
```diff
@@ -0,0 +1,29 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.blazebit.persistence.impl.function.datediff.microsecond;
+
+/**
+ *
+ * @author Moritz Becker
+ * @since 1.2.0
```
beikov: Author & version wrong

core/testsuite/src/test/java/com/blazebit/persistence/testsuite/DateDiffTest.java
```diff
@@ -398,11 +444,19 @@ public void testDateWithParameterDiff() {
...
+        assertEquals(weeksBetween(c2, c1), actual.get(23));
+        assertEquals(quartersBetween(c2, c1), actual.get(24));
     }
 
     public static long millisecondsBetween(Calendar day1, Calendar day2) {
         return day2.getTimeInMillis() - day1.getTimeInMillis();
     }
+
+    public static long microsecondsBetween(Calendar day1, Calendar day2) {
+        return (day2.getTimeInMillis() - day1.getTimeInMillis()) * 1000;
```
beikov: Since the best precision we have with Calendar is milliseconds, we should add a test that uses DBMS literals/functions to produce values with microseconds which we can check against. Currently we don't know for sure if we are correctly extracting microsecond parts

core/testsuite/src/test/java/com/blazebit/persistence/testsuite/DateExtractTest.java
```diff
@@ -94,12 +89,7 @@ public DateExtractTest(TimeZone producerTimeZone, TimeZone clientTimeZone) {
 
     @Override
     protected boolean recreateDataSource() {
-        // Some drivers have timezone information bound to the connection
-        // So we have to recreate the data source to get the newly configured time zones for connections
-        boolean recreate = previousProducerTimeZone != null && previousClientTimeZone != null && (!producerTimeZone.equals(previousProducerTimeZone) || !clientTimeZone.equals(previousClientTimeZone));
-        previousProducerTimeZone = producerTimeZone;
-        previousClientTimeZone = clientTimeZone;
-        return recreate;
+        return true;
```
beikov: Shouldn't this be reverted?

core/impl/src/main/java/com/blazebit/persistence/impl/function/datediff/quarter/DB2QuarterDiffFunction.java
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * @author Jan-Willem Gmelig Meyling
+ * @since 1.4.0
+ */
+public class DB2QuarterDiffFunction extends QuarterDiffFunction {
+
+    public DB2QuarterDiffFunction() {
+        // NOTE: we need lateral, otherwise the alias will be lost in the subquery
+        super("(select sign(i1) * floor(abs(i1)) from lateral(values (trunc(-months_between(?1, ?2))/3)) as temp(i1))");
```
beikov: Shouldn't a cast to int be enough?

core/impl/src/main/java/com/blazebit/persistence/impl/function/datediff/quarter/MSSQLQuarterDiffFunction.java
```diff
@@ -0,0 +1,29 @@
...
+package com.blazebit.persistence.impl.function.datediff.quarter;
+
+/**
+ * @author Jan-Willem Gmelig Meyling
+ * @since 1.4.0
+ */
+public class MSSQLQuarterDiffFunction extends QuarterDiffFunction {
+
+    public MSSQLQuarterDiffFunction() {
+        super("(select sign(i1) * floor(abs(i1)) from (values (datediff(mm, ?1, ?2)/3)) as temp(i1))");
```
beikov: Shouldn't a cast to int be enough?

core/impl/src/main/java/com/blazebit/persistence/impl/function/datediff/quarter/OracleQuarterDiffFunction.java
```diff
@@ -0,0 +1,28 @@
...
+package com.blazebit.persistence.impl.function.datediff.quarter;
+
+/**
+ * @author Jan-Willem Gmelig Meyling
+ * @since 1.4.0
+ */
+public class OracleQuarterDiffFunction extends QuarterDiffFunction {
+
+    public OracleQuarterDiffFunction() {
+        super("(select sign(i1) * floor(abs(i1)) from (values (trunc(-months_between(?1, ?2))/3)) as temp(i1))");
```
beikov: Shouldn't a cast to int be enough?

core/impl/src/main/java/com/blazebit/persistence/impl/function/datediff/quarter/PostgreSQLQuarterDiffFunction.java
```diff
@@ -0,0 +1,29 @@
...
+package com.blazebit.persistence.impl.function.datediff.quarter;
+
+/**
+ * @author Jan-Willem Gmelig Meyling
+ * @since 1.4.0
+ */
+public class PostgreSQLQuarterDiffFunction extends QuarterDiffFunction {
+
+    public PostgreSQLQuarterDiffFunction() {
+        super("(select sign(i2) * floor(abs(i2)) from ( select (date_part('year', t1) * 12 + date_part('month', t1))/-3 from (values (age(?1,?2))) as temp(t1)) as temp2(i2))");
```
beikov: Shouldn't a cast to int be enough?

core/impl/src/main/java/com/blazebit/persistence/impl/function/datediff/week/DB2WeekDiffFunction.java
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * @author Jan-Willem Gmelig Meyling
+ * @since 1.4.0
+ */
+public class DB2WeekDiffFunction extends WeekDiffFunction {
+
+    public DB2WeekDiffFunction() {
+        // NOTE: we need lateral, otherwise the alias will be lost in the subquery
+        super("(select sign(i1) * floor(abs(i1)) from lateral(values (trunc(-(cast(?1 as date) - cast(?2 as date))) / 7)) as temp(i1))");
```
beikov: Shouldn't a cast to int be enough?

core/impl/src/main/java/com/blazebit/persistence/impl/function/datediff/week/OracleWeekDiffFunction.java
```diff
@@ -0,0 +1,28 @@
...
+package com.blazebit.persistence.impl.function.datediff.week;
+
+/**
+ * @author Jan-Willem Gmelig Meyling
+ * @since 1.4.0
+ */
+public class OracleWeekDiffFunction extends WeekDiffFunction {
+
+    public OracleWeekDiffFunction() {
+        super("(select sign(i1) * floor(abs(i1)) from (values (trunc(-(cast(?1 as date) - cast(?2 as date))) / 7)) as temp(i1))");
```
beikov: Shouldn't a cast to int be enough?

core/impl/src/main/java/com/blazebit/persistence/impl/function/datediff/week/PostgreSQLWeekDiffFunction.java
```diff
@@ -0,0 +1,29 @@
...
+package com.blazebit.persistence.impl.function.datediff.week;
+
+/**
+ * @author Jan-Willem Gmelig Meyling
+ * @since 1.4.0
+ */
+public class PostgreSQLWeekDiffFunction extends WeekDiffFunction {
+
+    public PostgreSQLWeekDiffFunction() {
+        super("(select sign(i1) * floor(abs(i1)) from (values (extract('day' from (?1::timestamp - ?2::timestamp)) / -7)) as temp(i1))");
```
beikov: Shouldn't a cast to int be enough?

core/impl/src/main/java/com/blazebit/persistence/impl/function/datediff/week/OracleWeekDiffFunction.java
beikov: Also, casting to date here will result in loss of precision

core/impl/src/main/java/com/blazebit/persistence/impl/function/datediff/week/DB2WeekDiffFunction.java
beikov: Also, casting to date here will result in loss of precision

beikov:
- [ ] epoch_microseconds - epoch + microseconds or native function (#493)
- [ ] epoch_milliseconds - epoch + milliseconds or native function (#493)
- [ ] epoch_seconds - alias for epoch (#493)
- [ ] epoch_days - epoch / 86400‬ or native function (#493)
- [ ] Oracle timetstamp arithmetic optimization tests
- [ ] extract documentation update

core/testsuite/src/test/java/com/blazebit/persistence/testsuite/DateExtractTest.java
JWGmeligMeyling: No, the date extract tests are not reliable without this. Tests started failing after adding the epoch_seconds alias for epoch. Even though they use the same function implementation and test logic, the tests fail if both tests for epoch snd epich seconds are ran without this change.

core/impl/src/main/java/com/blazebit/persistence/impl/function/datediff/quarter/DB2QuarterDiffFunction.java
JWGmeligMeyling: Db2, psql and oracle round to the closest int when casting, which is not appropiate here.

core/impl/src/main/java/com/blazebit/persistence/impl/function/datediff/quarter/MSSQLQuarterDiffFunction.java
JWGmeligMeyling: Db2, psql and oracle round to the closest int when casting, which is not appropiate here

core/impl/src/main/java/com/blazebit/persistence/impl/function/datediff/quarter/OracleQuarterDiffFunction.java
JWGmeligMeyling: Db2, psql and oracle round to the closest int when casting, which is not appropiate here

core/impl/src/main/java/com/blazebit/persistence/impl/function/datediff/quarter/PostgreSQLQuarterDiffFunction.java
JWGmeligMeyling: Db2, psql and oracle round to the closest int when casting, which is not appropiate here

core/impl/src/main/java/com/blazebit/persistence/impl/function/datediff/week/DB2WeekDiffFunction.java
JWGmeligMeyling: Db2, psql and oracle round to the closest int when casting, which is not appropiate here. Ill try to find another impl for this, i copied the days extraction from the db2 day diff function. Anyways, the db2 tests currently struggle anyway.

core/impl/src/main/java/com/blazebit/persistence/impl/function/datediff/week/OracleWeekDiffFunction.java
JWGmeligMeyling: Db2, psql and oracle round to the closest int when casting, which is not appropiate here. Ill try to find another impl for this, i copied the days extraction from the oracle day diff function. Anyways, the oracle tests currently struggle anyway

core/impl/src/main/java/com/blazebit/persistence/impl/function/datediff/week/PostgreSQLWeekDiffFunction.java
JWGmeligMeyling: Db2, psql and oracle round to the closest int when casting, which is not appropiate here. L

PR #749, opened 2019-03-13
[#748] fix parser error with subview treat mapping, restructure gramm…

### LINKED ISSUE #748: Subview mapping with TREAT fails to parse
The following subview mapping fails to parse:
```java
@Mapping("TREAT(parent AS BaseProduct)")
BaseProductView getBaseProduct();
```
The error is `no viable alternative at input`.

Mobe91:
fixes #748


beikov:
You forgot to give explicit table and column names for the oracle identifier length restriction: https://travis-ci.org/Blazebit/blaze-persistence/jobs/506713184#L2734

beikov:
Also, your _putting entity classes into the test classes_ backfires when using datanucleus which requires enhancement for all entity classes. Entity classes are only enhanced if you put a reference into the orm.xml: https://travis-ci.org/Blazebit/blaze-persistence/jobs/506713187#L2147

Issue #753, opened 2019-03-14
Map key subview mapping not working


Mobe91:

beikov:
Merged by 21836fd

PR #754, opened 2019-03-15
[#753] map key subview mapping not working

Mobe91:


Mobe91:
Added test case to reproduce the issue.

beikov:
You forgot respect the oracle identifier length limitation: https://travis-ci.org/Blazebit/blaze-persistence/jobs/506712373#L2740

beikov:
Instead of excluding Oracle from the tests, please name the tables and other objects accordingly to be able to run these tests there too. Also, please add a comment on class level about why the excludes are necessary.

PR #760, opened 2019-03-16
[#759, #761] Test and fix for absolute paths referring to dependent joins in ON clauses

### LINKED ISSUE #759: Implicit join in ON clause rewrite mechanism fails to detect the proper use the base node
### Description

A query like `FROM Order o LEFT JOIN o.orderPositions p LEFT JOIN SomeEntity e ON o.orderPositions.id = e.id` fails to recognize that `o.orderPositions.id` is a path that doesn't need implicit joining because it is part of the JOIN base and thus already available for use within the ON clause. This triggers an EXISTS query rewrite that is wrong because the subquery refers to a collection role that is different from the one in the outer query.

### Environment

Version:            1.3.0
JPA-Provider:       -
DBMS:               -
Application Server: -

### LINKED ISSUE #761: Attributes having the same name as a correlation basis alias aren't prefixed in the correlation expression
### Description

When having e.g. a query root alias named `workflow` and using the fetch strategy SUBSELECT for fetching `workflowSteps`, which is _mappedBy_ an attribute called `workflow`, the resulting correlation expression will be `workflow IN __correlationAlias`. This expression is processed to `workflow = workflow` because we don't prefix `workflow` as it is the view root alias.
Either we defer the view root macro rendering or we figure out something else to properly fix this. 

### Environment

Version:            1.3.0
JPA-Provider:       -
DBMS:               -
Application Server: -

beikov:
## Related Issue

Fixes #759, Fixes #761

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ None
 ### Bug fixes
 
 * Fix support for de-referencing map key entity type expressions
+* Fix for absolute paths referring to dependent joins in ON clauses
 
 ### Backwards-incompatible changes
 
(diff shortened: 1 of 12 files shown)


beikov:
Merged by 978ca56a089925fb923e4cdd69a93f1eb9b1b821 and d927a1dc27a1a07367a32fa2db4fe80ed1e9d44e

Issue #765, opened 2019-03-20
Add support for composite user types in Hibernate


AFbenHassine:
Currently it is not possible to query composite properties of an entity. However, Hibernate does support such constructions through the `CompositeUserType` interface and `@Type` annotation. 
In our implementation of Java Money using Jadira we encountered such a composite type; `Money`, which consists of the properties `amount` and `currency`. We would like to be able to query it correctly using Blaze. 

Currently, attempting to query a composite type throws 
```
java.lang.IllegalArgumentException: The join path [product.price.amount] has a non joinable part [price.amount]
```

jwgmeligmeyling:
I recently started working on this issue (https://github.com/jwgmeligmeyling/blaze-persistence/commit/6c2ca771c34c5c3f43cb903921d8f861906a1f03). Basically we need to treat Composite types as embeddables. Making this change in our metamodel is relatively easy. However, a lot of code still uses the plain JPA metamodel. In order to support composite user types fully, we need to change these occurrences of JPA metamodel access to that they use BP's internal metamodel instead. This is a change we're willing to make, because it's not only required for dereferencing composite user types but its also relevant for dereferencing JSON properties.

jwgmeligmeyling:
Since Hibernate 5.4.22 Composite User Types are expressed in the JPA Metamodel as Embeddables. As a result, accessing composite user type properties works in Blaze-Persistence when using Hibernate 5.4.22.

Adding support for it in Blaze-Persistence would make it possible to dereference composite user type properties with Hibernate versions prior to 5.4.22 as well.

For now the workaround around this issue is simply upgrading to the latest Hibernate version.

That said, there are still various other enhancements that would benefit from a custom JPA metamodel wrapper in Blaze-Persistence.

beikov:
What do you think about merging the parser into core-impl? I don't think there is a real need to have the parser separately anyway. Merging would make all of this a lot easier I guess? The parser is not something that we particularly mention as API so I would say it is fine to do the move for a minor release i.e. for 1.7

jwgmeligmeyling:
That's certainly possible. Didn't really consider it an option as parser was recently separated from core as far as I remember 😄

PR #766, opened 2019-03-20
[#765]  support Hibernate composite user types

AFbenHassine:
Based on issue: https://github.com/Blazebit/blaze-persistence/issues/765

### DIFF
--- a/core/api/src/main/java/com/blazebit/persistence/spi/JpaMetamodelAccessor.java
+++ b/core/api/src/main/java/com/blazebit/persistence/spi/JpaMetamodelAccessor.java
@@ -20,6 +20,7 @@ import javax.persistence.metamodel.Attribute;
 import javax.persistence.metamodel.EntityType;
 import javax.persistence.metamodel.ManagedType;
 import javax.persistence.metamodel.Metamodel;
+import java.util.Collection;
 
 /**
  * @author Jan-Willem Gmelig Meyling
@@ -76,4 +77,8 @@ public interface JpaMetamodelAccessor {
      * @return true if the attribute is an element collection, false otherwise
      */
     boolean isElementCollection(Attribute<?, ?> attribute);
+
+    boolean isCompositeUserType(Metamodel metamodel, Attribute<?,?> attribute);
+
+    Collection<String> getCompositePropertyNames(Metamodel metamodel, Attribute<?,?> attribute);
 }
--- a/core/impl/src/main/java/com/blazebit/persistence/impl/EntityMetamodelImpl.java
+++ b/core/impl/src/main/java/com/blazebit/persistence/impl/EntityMetamodelImpl.java
@@ -40,6 +40,7 @@ import javax.persistence.metamodel.Metamodel;
 import javax.persistence.metamodel.PluralAttribute;
 import javax.persistence.metamodel.SingularAttribute;
 import javax.persistence.metamodel.Type;
+import java.math.BigDecimal;
 import java.util.AbstractMap;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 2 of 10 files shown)


core/impl/src/main/java/com/blazebit/persistence/impl/EntityMetamodelImpl.java
```diff
@@ -337,6 +339,20 @@ private TemporaryExtendedManagedType collectColumnNames(EntityType<?> e, Map<Str
...
+                else if (jpaProvider.getJpaMetamodelAccessor().isCompositeUserType(delegate, attribute)) {
+                    Collection<String> compositePropertyNames = jpaProvider.getJpaMetamodelAccessor().getCompositePropertyNames(delegate, attribute);
+
+
+
+                    for (String compositePropertyName : compositePropertyNames) {
+                        String idPath = attributeName + "." + compositePropertyName;
+                        List<Attribute<?,?>> idParents = new ArrayList<>();
+                        idParents.add(attribute);
+                        AttributeEntry attributeEntry = new AttributeEntry(jpaProvider, type, attribute, idPath, BigDecimal.class, idParents, null);
```
beikov: We need a way to access the property information like the java type, column type, nullability etc. Here you are always registering `BigDecimal` as type. I know it's just a PoC, I'm just mentioning it.

core/api/src/main/java/com/blazebit/persistence/spi/JpaMetamodelAccessor.java
```diff
@@ -76,4 +77,8 @@
      * @return true if the attribute is an element collection, false otherwise
      */
     boolean isElementCollection(Attribute<?, ?> attribute);
+
+    boolean isCompositeUserType(Metamodel metamodel, Attribute<?,?> attribute);
```
beikov: `Metamodel` isn't necessary

core/api/src/main/java/com/blazebit/persistence/spi/JpaMetamodelAccessor.java
```diff
@@ -76,4 +77,8 @@
...
+
+    Collection<String> getCompositePropertyNames(Metamodel metamodel, Attribute<?,?> attribute);
```
beikov: `Metamodel` isn't necessary

JWGmeligMeyling:
> Overall it looks interesting, although I think we could try to treat composite user types like embeddables which would handle many possibly problematic cases like joins, property expansion etc. out of the box.

Yes I think that is definitely the way to go, but I wasn't exactly sure how to achieve it. Are nested properties from embeddables currently reflected in the metamodel?

beikov:
> Yes I think that is definitely the way to go, but I wasn't exactly sure how to achieve it. Are nested properties from embeddables currently reflected in the metamodel?

Yes, an embeddable is completely covered by the metamodel. Maybe we could create custom EmbeddableType instances for composite types?

JWGmeligMeyling:
Interesting hook, I'll do some tests.

JWGmeligMeyling:
Where do you think we should represent this in the metamodel? I believe we quite heavily rely on the standard JPA metamodel over our extended metamodel? Where do you think this could be implemented the easiest? Any specific pointers to occurences of a method to look out for?

beikov:
It would be great if we could completely rely on the extendend metamodel we defined. I'm not sure how much work that is going to be though. I'd be ok with handling composite types specially, like you do now in the PR, given that you handle all the cases where embeddables are treated specially.

JWGmeligMeyling:
I'm closing this as @AFbenHassine won't finish it (he's heading to London for a 2nd masters). I am planning to finish it someday, but it will require quite some work, especially since we want to try the metamodel route. I want to concentrate on my other PRs first (#744 and #746). 

I remembered looking into this briefly when you mentioned it a while back. It seemed going to be quite involved, because a lot of places ([`JpaUtils`](https://github.com/Blazebit/blaze-persistence/blob/master/core/impl/src/main/java/com/blazebit/persistence/impl/JpaUtils.java) in particular) we still use the plain JPA metamodel rather than our own extended metamodel. So everything related to embeddables will have to be changed so that it uses our own metamodel.

PR #770, opened 2019-04-04
[#769] add support for treated join bases

### LINKED ISSUE #769: joinOn with treated join base not working
Something like `.leftJoinOn("TREAT(parent AS PolymorphicSub2)"` gives a parser error. This is also relevant for entity view correlations with join strategy.

### LINKED ISSUE #771: Support correlating with TREAT expression properly
There are some cases we didn't handle yet when correlating assocations in subqueries. A treat expression containing a deep path isn't properly handled i.e. it fails to generate a query because it tries to resolve the final assocation on the inner-most join node type rather than the treat type.

See `TreatedCorrelatedSubqueryTest` for a test.

Mobe91:
Fixes #769, Fixes #771


beikov:
Merged by 73376d90ad3f5894ee6899d0459e03dc61f7957b

PR #807, opened 2019-05-20
[#804] CTE support in entity views

### LINKED ISSUE #804: CTE support in entity views
Currently to use a CTE in an entity view one needs to map a bogus `SubqueryProvider` to be able to issue the `with`/`withRecursive` methods. It would be easier and more straightforward to have a declarative way to do this, for example with a `@With` annotation on the entity view:

```java
@EntityView(MyEntity.class)
@With(cte = MyCte.class, provider = MyCteProvider.class)
interface MyEntityView {

    @MappingCorrelatedSimple(correlated = MyCte.class, ...)
    long someMapping();
	
    @CTE
    @Entity
    class MyCte {
        ...
    }
    
    class MyCteProvider implements CteProvider {
        
        public void bindCte(FullSelectCTECriteriaBuilder builder) {
            builder.from(...).bind("id").select(...).end();
        } 
    }
}
```

The `@With` annotation would have a `recursive` boolean parameter and would be repeatable, e.g.

```java
@EntityView(MyEntity.class)
@With(cte = MyCte.class, provider = MyCteProvider.class)
@With(cte = MyRecursiveCte.class, recursive = true, provider = ...)
interface MyEntityView { ... }
```

heruan:
## Description

- [x] Define annotation and provider interface
- [x] Extend metamodel, annotation reader etc.
- [x] Implement the provider collecting and calling logic in `ViewTypeObjectBuilderTemplate`
- [x] Tests
- [x] Documentation

## Related Issue
Fixes #804

### DIFF
--- a/entity-view/api/src/main/java/com/blazebit/persistence/view/With.java
+++ b/entity-view/api/src/main/java/com/blazebit/persistence/view/With.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2014 - 2019 Blazebit.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.blazebit.persistence.view;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Adds {@link CTEProvider}s that should be invoked when applying the annotated entity view.
+ *
+ * @author Giovanni Lovato
+ * @since 1.4.0
+ */
+@Target({ ElementType.TYPE })
+@Retention(RetentionPolicy.RUNTIME)
+public @interface With {
+
+    /**
+     * The CTE providers.
+     *
+     * @return the CTE providers
+     */
+    Class<? extends CTEProvider>[] value();
+
+}
--- a/entity-view/api/src/main/java/com/blazebit/persistence/view/metamodel/ManagedViewType.java
+++ b/entity-view/api/src/main/java/com/blazebit/persistence/view/metamodel/ManagedViewType.java
@@ -16,8 +16,10 @@
 
 package com.blazebit.persistence.view.metamodel;
 
+import com.blazebit.persistence.view.CTEProvider;
 import com.blazebit.persistence.view.FlushMode;
 import com.blazebit.persistence.view.FlushStrategy;
+import com.blazebit.persistence.view.With;
 
 import java.lang.reflect.Method;
 import java.util.Set;
@@ -161,5 +163,13 @@ public interface ManagedViewType<X> extends Type<X> {
      * @since 1.2.0
      */
     public Set<String> getPersistabilityValidationExcludedEntityAttributes();
+
+    /**
+     * Returns the set of {@link CTEProvider} defined in the view using the {@link With} annotation.
+     *
+     * @return the set of {@link CTEProvider}
+     * @since 1.4.0
+     */
+    public Set<CTEProvider> getCteProviders();
 
 }
--- a/entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/CTEProviderFactory.java
+++ b/entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/CTEProviderFactory.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2014 - 2019 Blazebit.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.blazebit.persistence.view.impl;
+
+import com.blazebit.persistence.view.CTEProvider;
+
+/**
+ *
+ * @author Giovanni Lovato
+ * @since 1.4.0
+ */
+public interface CTEProviderFactory {
+
+    public CTEProvider create();
+}
--- a/entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/metamodel/AnnotationViewMappingReader.java
+++ b/entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/metamodel/AnnotationViewMappingReader.java
@@ -18,6 +18,7 @@ package com.blazebit.persistence.view.impl.metamodel;
 
 import com.blazebit.annotation.AnnotationUtils;
 import com.blazebit.persistence.view.BatchFetch;
+import com.blazebit.persistence.view.CTEProvider;
 import com.blazebit.persistence.view.CreatableEntityView;
 import com.blazebit.persistence.view.EntityView;
 import com.blazebit.persistence.view.EntityViewInheritance;
@@ -28,6 +29,7 @@ import com.blazebit.persistence.view.LockOwner;
 import com.blazebit.persistence.view.PostCreate;
 import com.blazebit.persistence.view.UpdatableEntityView;
 import com.blazebit.persistence.view.ViewConstructor;
+import com.blazebit.persistence.view.With;
 import com.blazebit.reflection.ReflectionUtils;
 
 import java.lang.annotation.Annotation;
@@ -37,6 +39,7 @@ import java.lang.reflect.Modifier;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -83,6 +86,14 @@ public class AnnotationViewMappingReader implements ViewMappingReader {
         if (batchFetch != null) {
             viewMapping.setDefaultBatchSize(batchFetch.size());
         }
+
+        Set<Class<? extends CTEProvider>> cteProviders = new LinkedHashSet<>();
+        for (Annotation a : AnnotationUtils.getAllAnnotations(entityViewClass)) {
+            if (a instanceof With) {
+                cteProviders.addAll(Arrays.asList(((With) a).value()));
+            }
+        }
+        viewMapping.setCteProviders(cteProviders);
 
         UpdatableEntityView updatableEntityView = AnnotationUtils.findAnnotation(entityViewClass, UpdatableEntityView.class);
 
--- a/entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/metamodel/ManagedViewTypeImpl.java
+++ b/entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/metamodel/ManagedViewTypeImpl.java
@@ -19,10 +19,13 @@ package com.blazebit.persistence.view.impl.metamodel;
 import com.blazebit.persistence.parser.util.JpaMetamodelUtils;
 import com.blazebit.persistence.spi.ExtendedAttribute;
 import com.blazebit.persistence.spi.ExtendedManagedType;
+import com.blazebit.persistence.view.CTEProvider;
 import com.blazebit.persistence.view.EntityViewManager;
 import com.blazebit.persistence.view.FlushMode;
 import com.blazebit.persistence.view.FlushStrategy;
 import com.blazebit.persistence.view.LockMode;
+import com.blazebit.persistence.view.impl.PrefixingQueryGenerator;
+import com.blazebit.persistence.view.impl.SimpleCTEProviderFactory;
 import com.blazebit.persistence.view.metamodel.ManagedViewType;
 import com.blazebit.persistence.view.metamodel.MappingConstructor;
 import com.blazebit.persistence.view.metamodel.MethodAttribute;
@@ -55,7 +58,6 @@ import java.util.Set;
 import java.util.SortedMap;
 import java.util.TreeMap;
 import java.util.TreeSet;
-import java.util.concurrent.ConcurrentSkipListMap;
 
 /**
  *
@@ -78,6 +80,7 @@ public abstract class ManagedViewTypeImpl<X> implements ManagedViewTypeImplement
     private final int defaultBatchSize;
     private final Map<String, AbstractMethodAttribute<? super X, ?>> attributes;
     private final NavigableMap<String, AbstractMethodAttribute<? super X, ?>> recursiveAttributes;
+    private final NavigableMap<String, AbstractMethodAttribute<? super X, ?>> recursiveSubviewAttributes;
     private final Set<AbstractMethodAttribute<? super X, ?>> updateMappableAttributes;
     private final AbstractMethodAttribute<? super X, ?>[] mutableAttributes;
     private final Map<ParametersKey, MappingConstructorImpl<X>> constructors;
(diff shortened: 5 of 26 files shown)


heruan:
So far I was able to collect the `With` annotations in the `ViewMapping`, but I couldn't find the right spot where to call `CTEBuilder::with` to apply the binding.

entity-view/api/src/main/java/com/blazebit/persistence/view/With.java
```diff
@@ -0,0 +1,74 @@
...
+     * @return The CTE class
+     */
+    Class<?> cte();
+
+    /**
+     * A flag set to {@code true} for recursive CTEs.
+     *
+     * @return true if recursive, false otherwise
+     */
+    boolean recursive() default false;
```
beikov: I don't think this is necessary, the provider will call with or withRecursive

entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/metamodel/AnnotationViewMappingReader.java
```diff
@@ -84,6 +85,11 @@ public ViewMapping readViewMapping(Class<?> entityViewClass) {
             viewMapping.setDefaultBatchSize(batchFetch.size());
         }
 
+        With[] with = entityViewClass.getAnnotationsByType(With.class);
```
beikov: Use something like
```
Set<CteProvider> cteProviders = new LinkedHashSet<>();
for (With w : AnnotationUtils.getAllAnnotations(With.class)) {
  cteProviders.add(w.value());
}
Set<Withs> withs = AnnotationUtils.getAllAnnotations(Withs.class);
for (Withs w : withs) {
  for (With with : w.value()) {
  cteProviders.add(with.value());
  }
}
viewMapping.setCteProviders(cteProviders);
```

entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/metamodel/ViewMapping.java
```diff
@@ -81,4 +82,8 @@
     List<Method> getSpecialMethods();
 
     void setSpecialMethods(List<Method> specialMethods);
+
+    List<With> getWithAnnotations();
```
beikov: `Set<CteProvider> getCteProviders()`

entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/metamodel/ViewMappingImpl.java
```diff
@@ -94,6 +95,8 @@
     private ManagedViewTypeImplementor<?> viewType;
     private Map<EmbeddableOwner, ManagedViewTypeImplementor<?>> embeddableViewTypeMap;
 
+    private List<With> withAnnotations = new ArrayList<>();
```
beikov: Null by default to avoid unnecessary allocations. Also, rename to `Set<CteProvider> cteProviders`

entity-view/api/src/main/java/com/blazebit/persistence/view/CteProvider.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+public interface CteProvider {
+
+    /**
+     * Binds a CTE.
+     *
+     * @param <X> the result type which is returned after the CTE builder
+     * @param builder the builder
+     */
+    <X> void bindCte(FullSelectCTECriteriaBuilder<X> builder);
```
beikov: Maybe use `void applyCtes(CTEBuilder<?> builder);` instead?

entity-view/api/src/main/java/com/blazebit/persistence/view/With.java
```diff
@@ -0,0 +1,74 @@
...
+@Retention(RetentionPolicy.RUNTIME)
+@Repeatable(RepeatableWith.class)
+public @interface With {
+
+    /**
+     * The CTE class.
+     *
+     * @return The CTE class
+     */
+    Class<?> cte();
```
beikov: IMO, you can remove this.

entity-view/api/src/main/java/com/blazebit/persistence/view/With.java
```diff
@@ -0,0 +1,74 @@
...
+     * @return true if recursive, false otherwise
+     */
+    boolean recursive() default false;
+
+    /**
+     * The CTE provider.
+     *
+     * @return the CTE provider
+     */
+    Class<? extends CteProvider> provider();
```
beikov: Rename to `value()`

entity-view/api/src/main/java/com/blazebit/persistence/view/With.java
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * This annotation invokes the {@link CteProvider} to bind the {@code cte} class
+ * to the query.
+ *
+ * @author Giovanni Lovato
+ * @since 1.4.0
+ */
+@Target({ ElementType.TYPE })
+@Retention(RetentionPolicy.RUNTIME)
+@Repeatable(RepeatableWith.class)
```
beikov: Unfortunately, this is Java 8 only and we are currently still on Java 7.

entity-view/api/src/main/java/com/blazebit/persistence/view/With.java
heruan: The idea was that the `With` annotation would already call `with` and provide a `FullSelectCTECriteriaBuilder` instead of a `CTEBuilder`. If we provide a `CTEBuilder` instead, then yeah there's no need for `cte` and `recursive` attributes!

entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/metamodel/AnnotationViewMappingReader.java
heruan: Do you mean `Set<Class<? extends CteProvider>>` or che actual instance?

beikov: Of course, I meant `Set<Class<? extends CteProvider>>`, sorry for that.

heruan: About naming conventions: should I use `CteProvider` or `CTEPRovider`? I see `CTE` is all uppercase in other classes.

heruan: I'll go with `CTEProvider`, to be consistent with `CTEBuilder`.

entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/metamodel/AnnotationViewMappingReader.java
```diff
@@ -84,6 +88,16 @@ public ViewMapping readViewMapping(Class<?> entityViewClass) {
...
+        Set<Class<? extends CTEProvider>> cteProviders = new LinkedHashSet<>();
+        for (Annotation with : AnnotationUtils.getAllAnnotations(With.class)) {
+            cteProviders.add(((With) with).value());
+        }
+        for (Annotation withs : AnnotationUtils.getAllAnnotations(Withs.class)) {
+            for (With with : ((Withs) withs).value()) {
+                cteProviders.add(with.value());
+            }
+        }
+
```
beikov: Don't forget to set the cte providers on the view mapping

beikov:
Now that we have an API, I'd like to see tests that cover

* Simple usage
* Usage in subviews
* Usage in polymorphic context
* Usage in polymorphic context of subview
* Multiple providers sprinkled around various subviews

It's also necessary to collect all CTEProviders, instantiate them and keep them in the metamodel in `ManagedViewTypeImpl`.

heruan:
> It's also necessary to collect all CTEProviders, instantiate them and keep them in the metamodel in `ManagedViewTypeImpl`.

Of course, that's where I get lost! I'm looking for a spot where we have a `CriteriaBuilder` instance, there I would instantiate the providers (reflection?) and invoke them with the builder.

heruan:
I'm not still sure we need a `Withs` annotation, if we pass the `CTEBuilder` we can bind all the CTEs from a single provider: the idea of the repeatable annotation was if it were already bound to a specific CTE.

beikov:
Applying the CTEProviders to a CriteriaBuilder would have to happen in `com.blazebit.persistence.view.impl.objectbuilder.ViewTypeObjectBuilder#applySelects`, but you will have to cast the query builder.
In the constructor of `ViewTypeObjectBuilder`, you can access `com.blazebit.persistence.view.impl.objectbuilder.ViewTypeObjectBuilderTemplate#getViewRoot` which then should expose the CTEProviders.
IMO `Withs` will allow to reuse existing CTEProvider implementations, so it's a nice addon. I actually think we might be able to pull off the use of `@Repeatable`, because non-existing annotations(in case of running on Java 7) are just ignored, but we'd have to test that.

heruan:
Should I create instances in `ManagedViewTypeImpl` or `ViewTypeObjectBuilder`?

beikov:
> Should I create instances in `ManagedViewTypeImpl` or `ViewTypeObjectBuilder`?

In `ManagedViewTypeImpl`. It's probably best to use the `com.blazebit.persistence.view.impl.metamodel.MetamodelBuildingContext#onViewTypeFinished` mechanism like used for initializing `recursiveAttributes`

entity-view/api/src/main/java/com/blazebit/persistence/view/metamodel/ManagedViewType.java
```diff
@@ -162,4 +164,11 @@
      */
     public Set<String> getPersistabilityValidationExcludedEntityAttributes();
 
+    /**
+     * Returns the set of {@link CTEProvider} defined in the view using the {@link With} annotation.
+     *
+     * @return the set of {@link CTEProvider}
+     */
```
beikov: Add `@since 1.4.0`

entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/CTEProviderFactory.java
```diff
@@ -0,0 +1,29 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.blazebit.persistence.view.impl;
+
+import com.blazebit.persistence.view.CTEProvider;
+
+/**
+ *
```
beikov: Javadoc

entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/CTEProviderFactory.java
```diff
@@ -0,0 +1,29 @@
...
+ * @author Giovanni Lovato
+ * @since 1.4.0
+ */
+public interface CTEProviderFactory {
+
+    public CTEProvider create();
```
beikov: Javadoc

entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/metamodel/ManagedViewTypeImpl.java
```diff
@@ -345,6 +348,17 @@ public void run() {
                 }
             }
         }
+
+        Set<Class<? extends CTEProvider>> cteProviderClasses = viewMapping.getCteProviders();
+        if (cteProviderClasses != null) {
+            this.cteProviders = new LinkedHashSet<>();
```
beikov: Always init the `cteProviders`, as we will have to add CTE providers from attribute subview types recursively.

heruan:
Initialization done using `onViewTypeFinished`, if I understood correctly!

heruan:
The current implementation calls the providers defined in the view root, not subviews: I guess this is because I collect them via `template.getViewRoot().getCteProviders()` in `ViewTypeObjectBuilder`. Can I also access the subview hierarchy from there?

beikov:
> The current implementation calls the providers defined in the view root, not subviews: I guess this is because I collect them via `template.getViewRoot().getCteProviders()` in `ViewTypeObjectBuilder`. Can I also access the subview hierarchy from there?

You have to iterate `recursiveAttributes` and in your `onViewTypeFinished` runnable

heruan:
I'm iterating over `attributes` since `recursiveAttributes` is missing subview attributes. Am I missing something?

I've also added a subview test and now both tests pass, but I have no idea about the polymorphic tests you asked.

entity-view/testsuite/src/test/java/com/blazebit/persistence/view/testsuite/cte/model/PersonWithPartnerDocumentFullAged.java
```diff
@@ -0,0 +1,29 @@
...
+package com.blazebit.persistence.view.testsuite.cte.model;
+
+import com.blazebit.persistence.testsuite.entity.Person;
+import com.blazebit.persistence.view.EntityView;
+import com.blazebit.persistence.view.EntityViewInheritance;
+import com.blazebit.persistence.view.EntityViewInheritanceMapping;
+import com.blazebit.persistence.view.testsuite.inheritance.subview.correlated.model.PersonBaseView;
+
+@EntityView(Person.class)
+@EntityViewInheritanceMapping("age < 18")
```
beikov: Wrong predicate

entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/metamodel/ManagedViewTypeImpl.java
```diff
@@ -719,6 +670,118 @@ public String getTypeConstraintMapping() {
...
+            }
+        }
+    }
+
+    /**
+     *
+     * @author Christian Beikov
+     * @since 1.4.0
+     */
+    private final class ElementTypeSubviewAttributeCollector implements Runnable {
```
beikov: Make static

entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/metamodel/ManagedViewTypeImpl.java
```diff
@@ -719,6 +670,118 @@ public String getTypeConstraintMapping() {
         return cteProviders;
     }
 
+    /**
+     *
+     * @author Giovanni Lovato
+     * @since 1.4.0
+     */
+    private final class CTEProviderCollector implements Runnable {
```
beikov: Make static

entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/metamodel/ManagedViewTypeImpl.java
```diff
@@ -719,6 +670,118 @@ public String getTypeConstraintMapping() {
...
+            }
+        }
+    }
+
+    /**
+     *
+     * @author Christian Beikov
+     * @since 1.4.0
+     */
+    private final class KeyTypeSubviewAttributeCollector implements Runnable {
```
beikov: Make static

entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/metamodel/ManagedViewTypeImpl.java
```diff
@@ -719,6 +670,118 @@ public String getTypeConstraintMapping() {
...
+            this.context = context;
+        }
+
+        @Override
+        public void run() {
+            Set<Class<? extends CTEProvider>> rootProviders = viewMapping.getCteProviders();
+            if (rootProviders != null) {
+                Map<Class<?>, CTEProvider> providers = context.getCteProviders();
+                for (Class<? extends CTEProvider> clazz : rootProviders) {
+                    if (!providers.containsKey(clazz)) {
```
beikov: Use
```
CTEProvider provider = providers.get(clazz);
if (provider == null) {
  CTEProvider provider = new SimpleCTEProviderFactory(clazz).create();
  providers.put(clazz, provider);
}
ManagedViewTypeImpl.this.cteProviders.add(provider);
```

entity-view/testsuite/src/test/java/com/blazebit/persistence/view/testsuite/cte/model/PersonWithPartnerDocumentFullAged.java
```diff
@@ -0,0 +1,29 @@
...
+public interface PersonWithPartnerDocumentFullAged extends PersonWithPartnerDocument {
```
beikov: You will have to add a test that also adds that entity view to the configuration. Regardless of that, you might need to add a `@With` with a `CTEProvider`. Your tests don't have to check end-to-end that it works, just check that the metamodel contains an instance of the CTEProvider.

heruan: > […] just check that the metamodel contains an instance of the CTEProvider.

Cool! Can I check that from `evm.getMetamodel()`?

beikov: Yeah, you added the method to `ManagedViewType` in this PR, so you can ;)

beikov:
Implementation-wise we are done, we only need a section in the documentation about this now.

beikov:
Also, please squash the commits to a single one.

heruan:
> Also, please squash the commits to a single one.

Won't this happen when merging?

beikov:
Unfortunately rebase and squash can't be used together from within the GitHub UI so please squash it. I added the documentation part and removed the `@Withs` annotation in favor of using an array of CTEProvider classes in `@With`. If you are ok with the changes, please squash it so I can merge it.

PR #834, opened 2019-08-23
[#259] [#690] Window Function and boolean aggregate support

### LINKED ISSUE #259: Implement window function support
We should create derivates of the standard aggregate functions e.g. `count` -> `count_window` that support the FILTER and OVER clauses. The FILTER clause will always have to be rendered by using a CASE WHEN statement in the function. The OVER clause should support PARTITION, ORDER BY and ROWS/RANGE clauses and also accept a named window.
In order to support a named window, we would first have to add support for adding windows to a query. I propose to add a method to start a window builder on a `QueryBuilder` e.g. `window(String name)`. There you can define partitions, orderings and framing. Note that windows defined in a subquery have precedence for that subquery. So windows can shadow same named windows from the parent query.

Note that since there is semantically no difference between reusing a window name or inlining the window clause, we will always inline the contents to the window functions as this is easier to implement because it doesn't require SQL replacements.

Here are some links to documentation about this
- https://msdn.microsoft.com/en-us/library/ms189461.aspx
- https://www.simple-talk.com/sql/t-sql-programming/window-functions-in-sql/
- https://blog.jooq.org/2013/11/03/probably-the-coolest-sql-feature-window-functions/
- http://modern-sql.com/feature/filter

The following functions would be affected by this
- AVG
- MIN
- MAX
- SUM
- EVERY?
- COUNT
- COUNT_STAR
- GROUP_CONCAT

Also see the following reference: https://en.wikibooks.org/wiki/SQL_Dialects_Reference/Functions_and_expressions/Math_functions/Aggregate_functions

While doing that we could also look into adding some window only functions: https://www.postgresql.org/docs/8.4/static/functions-window.html

### LINKED ISSUE #343: Get rid of sematic predicates in ANTLR grammar
Since we can do the necessary checks that are currently done in semantic predicates in the visitor instead, we should remove them from the grammar. I expect a major parsing performance improvement from that.

### LINKED ISSUE #566: Settig enum parameter values fails with parameterRenderingMode LITERAL
When selecting an enum valued parameter, the query rendering fails because the query generator tries to render the parameter as literal but `TypeUtils` does not have an enum converter.
```java
.select(":myEnum").setParameter(MyEnum.VAL1)
```
Exception:
```
17:28:28,900 SEVERE [com.ordami.geo.service.impl.service.MatrixUpdateServiceImpl] (EE-ManagedExecutorService-default-Thread-4) null: java.lang.NullPointerException
	at com.blazebit.persistence.impl.ResolvingQueryGenerator.getLiteralParameterValue(ResolvingQueryGenerator.java:508)
	at com.blazebit.persistence.parser.SimpleQueryGenerator.visit(SimpleQueryGenerator.java:451)
	at com.blazebit.persistence.impl.ResolvingQueryGenerator.visit(ResolvingQueryGenerator.java:492)
	at com.blazebit.persistence.parser.expression.ParameterExpression.accept(ParameterExpression.java:52)
	at com.blazebit.persistence.impl.ResolvingQueryGenerator.generate(ResolvingQueryGenerator.java:114)
	at com.blazebit.persistence.impl.SelectManager.applySelect(SelectManager.java:503)
	at com.blazebit.persistence.impl.SelectManager.buildSelect(SelectManager.java:297)
	at com.blazebit.persistence.impl.AbstractCommonQueryBuilder.appendSelectClause(AbstractCommonQueryBuilder.java:1878)
	at com.blazebit.persistence.impl.AbstractCommonQueryBuilder.buildBaseQueryString(AbstractCommonQueryBuilder.java:1851)
	at com.blazebit.persistence.impl.AbstractCommonQueryBuilder.buildBaseQueryString(AbstractCommonQueryBuilder.java:1846)
	at com.blazebit.persistence.impl.AbstractCommonQueryBuilder.getBaseQueryString(AbstractCommonQueryBuilder.java:1774)
	at com.blazebit.persistence.impl.AbstractCommonQueryBuilder.getBaseQueryStringWithCheck(AbstractCommonQueryBuilder.java:1512)
	at com.blazebit.persistence.impl.AbstractCommonQueryBuilder.getTypedQuery(AbstractCommonQueryBuilder.java:1526)
	at com.blazebit.persistence.impl.AbstractCommonQueryBuilder.getTypedQueryForFinalOperationBuilder(AbstractCommonQueryBuilder.java:1518)
	at com.blazebit.persistence.impl.BaseFinalSetOperationBuilderImpl.getTypedQuery(BaseFinalSetOperationBuilderImpl.java:221)
	at com.blazebit.persistence.impl.AbstractCommonQueryBuilder.getTypedQueryForFinalOperationBuilder(AbstractCommonQueryBuilder.java:1518)
	at com.blazebit.persistence.impl.BaseFinalSetOperationBuilderImpl.getTypedQuery(BaseFinalSetOperationBuilderImpl.java:221)
	at com.blazebit.persistence.impl.BaseFinalSetOperationBuilderImpl.getQuery(BaseFinalSetOperationBuilderImpl.java:322)
	at com.blazebit.persistence.impl.BaseFinalSetOperationBuilderImpl.getQuery(BaseFinalSetOperationBuilderImpl.java:45)
	at com.blazebit.persistence.impl.AbstractCommonQueryBuilder.getQuery(AbstractCommonQueryBuilder.java:1717)
	at com.blazebit.persistence.impl.AbstractCommonQueryBuilder.getCteNodes(AbstractCommonQueryBuilder.java:1613)
	at com.blazebit.persistence.impl.AbstractCommonQueryBuilder.getTypedQuery(AbstractCommonQueryBuilder.java:1553)
	at com.blazebit.persistence.impl.AbstractQueryBuilder.getQuery(AbstractQueryBuilder.java:52)
	at com.blazebit.persistence.impl.AbstractQueryBuilder.getResultList(AbstractQueryBuilder.java:57)
	at com.ordami.geo.service.impl.data.LocationDataAccessImpl.getMatrixDiff(LocationDataAccessImpl.java:166)
```

### LINKED ISSUE #690: Provide built-in boolean aggregate functions
Provide built-in boolean aggregate functions for supporting DBMS:
* `or_agg`
* `and_agg`

### LINKED ISSUE #776: NullPointerException when performing manual select("1") in exists
Happens to me with the current master (73376d90ad3f5894ee6899d0459e03dc61f7957b), don't know if this was also an issue before.
Relevant portion of the stack trace:
```
java.lang.NullPointerException: null
	at com.blazebit.persistence.impl.JoinManager.removeSelectOnlyNodes(JoinManager.java:269)
	at com.blazebit.persistence.impl.JoinManager.removeSelectOnlyNodes(JoinManager.java:256)
	at com.blazebit.persistence.impl.JoinManager.removeSelectOnlyNodes(JoinManager.java:247)
	at com.blazebit.persistence.impl.SelectManager.clearDefaultSelects(SelectManager.java:566)
	at com.blazebit.persistence.impl.SelectManager.select(SelectManager.java:441)
	at com.blazebit.persistence.impl.SelectManager.select(SelectManager.java:436)
	at com.blazebit.persistence.impl.AbstractCommonQueryBuilder.select(AbstractCommonQueryBuilder.java:1061)
	at com.blazebit.persistence.impl.AbstractCommonQueryBuilder.select(AbstractCommonQueryBuilder.java:1050)
```

### LINKED ISSUE #818: Move com.blazebit.persistence.view.impl.ConfigurationProperties to API instead Impl artifact.
Would be convenient to have statically typed property-keys without throwing whole impl library to compile classpath.

JWGmeligMeyling:
## Description
This PR introduces Window Function support. Tests pass on H2 and PostgreSQL, currently investigating which dialect specific changes are required to make the entire test suite pass.

It would be nice to extend the JPQL+ language with a special window syntax, but I have not looked into that yet. I'll probably need some pointers for that 😄 

Missing things

- [x] Named window support
- [x] Parser support for `FILTER` and `OVER` clause for window functions + nicer function invocation syntax without the need for the `FUNCTION` wrapper
- [x] Parser tests for new window function syntax
- [x] Query execution/translation tests. Maybe just rewrite `WindowFunctionTest` to use the new syntax
- [x] Documentation about window functions and other newly introduced functions

## Related Issue
* Fixes #343 Get rid of semantic predicates
* Fixes #259 Implement Window Function Support
* Fixes #690 Provide built-in boolean aggregate functions
* Fixes #818 Move configuration properties to API projects
* Fixes #776 NPE when performing `select("1")` after `whereExists`
* Fixes #566 Avoid NPE when trying to render parameter as literal


core/impl/src/main/java/com/blazebit/persistence/impl/function/window/AbstractWindowFunction.java
```diff
@@ -0,0 +1,557 @@
...
+import java.util.List;
+
+/**
+ *
+ * @author Jan-Willem Gmelig Meyling
+ * @author Sayra Ranjha
+ * @since 1.4.0
+ */
+public abstract class AbstractWindowFunction implements JpqlFunction {
+
```
beikov: IMO a constructor should capture the window support level which is then provided on a per DBMS level. Also the function name should IMO be passed as constructor argument.
Something like this

```
publict AbstractWindowFunction(String functionName, DbmsDialect dbmsDialect) {
  //...
}
```

And then use newly introduced methods e.g. `dbmsDialect.supportsFilter()`, `dbmsDialect.supportsNullPrecedence()`

core/impl/src/main/java/com/blazebit/persistence/impl/function/window/AbstractWindowFunction.java
```diff
@@ -0,0 +1,557 @@
...
+        context.addChunk(windowFunction.getFunctionName());
+        context.addChunk("(");
+
+        renderArguments(context, windowFunction);
+
+        context.addChunk(") ");
+    }
+
+    protected void renderArguments(FunctionRenderContext context, WindowFunction windowFunction) {
+        Iterator<String> argumentIterator = windowFunction.getArguments().iterator();
```
beikov: Since we know we are dealing with array lists here, I'd prefer an iteration like:

```
        List<String> arguments = windowFunction.getArguments();
        int size = arguments.size();
        if (size != 0) {
            context.addChunk(arguments.get(0));
            for (int i = 1; i < size; i++) {
                context.addChunk(", ");
                context.addChunk(arguments.get(i));
            }
        }
```

Also for all other iterations in this class

core/impl/src/main/java/com/blazebit/persistence/impl/function/window/AbstractWindowFunction.java
```diff
@@ -0,0 +1,557 @@
...
+    
+    /**
+     *
+     * @author Jan-Willem Gmelig Meyling
+     * @author Sayra Ranjha
+     * @since 1.4.0
+     */
+    protected static final class WindowFunction {
+
+        private String functionName;
```
beikov: Make final and initialize in the constructor.

core/impl/src/main/java/com/blazebit/persistence/impl/function/window/AbstractWindowFunction.java
```diff
@@ -0,0 +1,557 @@
...
+
+    /**
+     *
+     * @author Jan-Willem Gmelig Meyling
+     * @author Sayra Ranjha
+     * @since 1.4.0
+     */
+    protected static final class WindowDefinition {
+
+        private Partition partition = new Partition();
```
beikov: Inline the array list

core/impl/src/main/java/com/blazebit/persistence/impl/function/window/AbstractWindowFunction.java
```diff
@@ -0,0 +1,557 @@
...
+
+        private List<Order> orderBys  = new ArrayList<>();
+
+        private Frame frame;
```
beikov: Inline the Frame

core/impl/src/main/java/com/blazebit/persistence/impl/function/window/AbstractWindowFunction.java
```diff
@@ -0,0 +1,557 @@
...
+     *
+     * @author Jan-Willem Gmelig Meyling
+     * @author Sayra Ranjha
+     * @since 1.4.0
+     */
+    protected static final class Frame {
+        
+        private FrameMode mode;
+        
+        private FramePosition start;
```
beikov: Inline the FramePosition

beikov:
I will work on the two other issues(named window and parser support) as soon as you are done with the basics.

core/impl/src/main/java/com/blazebit/persistence/impl/function/window/oragg/WindowOrAggFunction.java
```diff
@@ -0,0 +1,40 @@
...
+package com.blazebit.persistence.impl.function.window.oragg;
+
+import com.blazebit.persistence.impl.function.window.AbstractWindowFunction;
+
+/**
+ * @author Jan-Willem Gmelig Meyling
+ * @author Sayra Ranjha
+ * @since 1.4.0
+ */
+public class WindowOrAggFunction extends AbstractWindowFunction {
```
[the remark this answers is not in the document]
JWGmeligMeyling: Because the standard one is an aggregrate function so it should have a different name so that the window function does not trigger for example implicit group bys to be rendered. I thought it would be the cleanest to keep aggregate and window functions separate, similar to how the implementations for sum, max, min, avg are already seperate (although they are defined in JPA alteady)

core/impl/src/main/java/com/blazebit/persistence/impl/function/window/AbstractWindowFunction.java
```diff
@@ -0,0 +1,557 @@
...
+            return new Order(expression, false, false);
+        }
+
+        return null;
+    }
+    
+    protected void render(FunctionRenderContext context, WindowFunction windowFunction) {
+        renderFunction(context, windowFunction);
+
+        List<String> filterExpressions = windowFunction.getFilterExpressions();
```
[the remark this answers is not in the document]
JWGmeligMeyling: Yup this is still in progress :)

core/impl/src/main/java/com/blazebit/persistence/impl/function/window/AbstractWindowFunction.java
JWGmeligMeyling: Agreed. I actually was suspicious why we didnt already do that. Group concat could benefit from this as well. However, the actual dialect used (with any potential configuration) is not available at construction time. Not sure how many dialects actually use the extra constructor with additional configuration/types.

core/impl/src/main/java/com/blazebit/persistence/impl/function/window/AbstractWindowFunction.java
JWGmeligMeyling: Are you sure? I think for custom extensions the abstraction could really ease translating the structure  to a new SQL dialect and which exceptions to keep in mind. See for example how I broke out the render method in individual protected methods that can be overridden.

JWGmeligMeyling:
With regard to the named window support: we discussed this before on slack or on the issue somewhere, but you only intend to manage some named windows in the criteria builder and inline them in the query correct? Because I can just as easily reference named windows here but then we have to think how we amend the actual window definitions at the end of the query.

core/impl/src/main/java/com/blazebit/persistence/impl/function/window/AbstractWindowFunction.java
JWGmeligMeyling: I have addressed this comment.

core/impl/src/main/java/com/blazebit/persistence/impl/function/window/every/WindowEveryFunction.java
```diff
@@ -0,0 +1,40 @@
...
+package com.blazebit.persistence.impl.function.window.every;
+
+import com.blazebit.persistence.impl.function.window.AbstractWindowFunction;
+
+/**
+ * @author Jan-Willem Gmelig Meyling
+ * @author Sayra Ranjha
+ * @since 1.4.0
+ */
+public class WindowEveryFunction extends AbstractWindowFunction {
```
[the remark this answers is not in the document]
JWGmeligMeyling: Answered in https://github.com/Blazebit/blaze-persistence/pull/834#discussion_r317241843

core/impl/src/main/java/com/blazebit/persistence/impl/function/window/oragg/WindowOrAggFunction.java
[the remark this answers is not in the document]
beikov: Yeah, you are totally right. I missed that.

core/impl/src/main/java/com/blazebit/persistence/impl/function/window/AbstractWindowFunction.java
beikov: You can still have separate methods that accept a subset of the window fields to do the rendering. The intention is to avoid unnecessary objects i.e. flatten them away.

integration/deltaspike-data/testsuite/src/test/java/com/blazebit/persistence/deltaspike/data/testsuite/FullEntityViewRepositoryTest.java
```diff
@@ -112,6 +115,7 @@ public void testFindByNameKeysetPaginated() {
     }
 
     @Test
+    @Category(NoH2.class)
```
beikov: Would you mind noting the reason for disabling the test here and above?

entity-view/testsuite/src/test/java/com/blazebit/persistence/view/testsuite/update/converter/EntityViewUpdateBlobTest.java
```diff
@@ -215,7 +216,7 @@ public void testSetBlob() throws Exception {
     // Oracle keeps LOBs open/tied to a result set so we can't write to it by means of setBytes.
     // For Oracle it would be more appropriate to treat writes like "replacements" i.e. remember the written bytes and replace the underlying object on flush
     // NOTE: No Datanucleus support yet
-    @Category({ NoDatanucleus.class, NoEclipselink.class, NoOracle.class })
+    @Category({ NoDatanucleus.class, NoEclipselink.class, NoOracle.class, NoH2.class})
```
beikov: Again, please add a note explaining why this test fails on H2

JWGmeligMeyling: Yes. There reason is that it broke in a new H2 :( still investigating that, and report issues where needed

core/impl/src/main/java/com/blazebit/persistence/impl/function/window/count/CountFunction.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    @Override
+    protected String getFunctionName() {
+        return "COUNT";
+    }
+
+    @Override
+    public Class<?> getReturnType(Class<?> firstArgumentType) {
+        return Integer.class;
```
beikov: I think JPA requires this to return a Long

core/impl/src/main/java/com/blazebit/persistence/impl/function/every/FallbackEveryFunction.java
```diff
@@ -0,0 +1,45 @@
...
+public class FallbackEveryFunction extends EveryFunction {
+
+    @Override
+    public void render(FunctionRenderContext context) {
+        renderLhs(context);
+        renderRhs(context);
+    }
+
+    public void renderLhs(FunctionRenderContext context) {
+        context.addChunk("SUM(CASE WHEN ");
```
beikov: Since `false` is typically smaller than `true` this function can be implemented as `MIN(argument)` instead which will also preserve the underlying type which might be `BIT`, `NUMBER(1)`, `CHAR(1)` etc. This should fix the Oracle and MSSQL issue: https://travis-ci.org/Blazebit/blaze-persistence/jobs/576190641 and https://travis-ci.org/Blazebit/blaze-persistence/jobs/576190639

core/impl/src/main/java/com/blazebit/persistence/impl/function/oragg/FallbackOrAggFunction.java
```diff
@@ -0,0 +1,46 @@
...
+
+    @Override
+    public void render(FunctionRenderContext context) {
+        renderLhs(context);
+        renderRhs(context);
+    }
+
+
+    public void renderLhs(FunctionRenderContext context) {
+        context.addChunk("SUM(CASE WHEN ");
```
beikov: Same as https://github.com/Blazebit/blaze-persistence/pull/834/files#r317388047

This can be implemented by using `MAX(argument)`

core/impl/src/main/java/com/blazebit/persistence/impl/ResolvingQueryGenerator.java
```diff
@@ -343,6 +388,88 @@ protected void renderFunctionFunction(String functionName, List<Expression> argu
...
+                sb.append(",'ORDER BY',");
+                visit(orderByExpressions.get(0));
+                for (int i = 1; i < size; i++) {
+                    sb.append(", ");
+                    visit(orderByExpressions.get(i));
+                }
+            }
+
+            if (windowDefinition.getFrameMode() != null) {
+                sb.append(",'");
```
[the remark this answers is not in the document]
beikov: Agree, could you fix this and the other comma issues please?

core/impl/src/main/java/com/blazebit/persistence/impl/WindowManager.java
```diff
@@ -0,0 +1,291 @@
...
+            return windowDefinition;
+        }
+
+        WindowDefinition baseWindowDefinition = windows.get(windowDefinition.getWindowName());
+        // See https://www.postgresql.org/docs/current/sql-select.html
+        if (baseWindowDefinition == null) {
+            throw new IllegalArgumentException("There is no window named '" + windowDefinition.getWindowName() + "' registered!");
+        }
+        if (baseWindowDefinition.getFrameMode() != null) {
+            throw new IllegalArgumentException("The base window '" + windowDefinition.getWindowName() + "' is not allowed to specify a frame clause when being reused!");
```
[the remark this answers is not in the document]
JWGmeligMeyling: The reason being:

> For rather inscrutable reasons, SQL:2008 disallows copying-and-modifying a window definition that has any explicit framing clause.  The error message we gave for this only made sense if the referencing window definition itself contains an explicit framing clause, which it might well not. Moreover, in the context of an OVER clause it’s not exactly obvious that “OVER (windowname)” implies copy-and-modify while “OVER windowname” does not.  This has led to multiple complaints, eg bug #5199 from Iliya Krapchatov.  Change to a hopefully more intelligible error message, and in the case where we have just “OVER (windowname)“, add a HINT suggesting that omitting the parentheses will fix it.  Also improve the related documentation.  Back-patch to all supported branches.

The documentation in the code elaborates:

>  ```
> /*
>   * Per spec, a windowdef that references a previous one copies the
>   * previous partition clause (and mustn't specify its own).  It can
>   * specify its own ordering clause, but only if the previous one had
>   * none.  It always specifies its own frame clause, and the previous
>   * one must not have a frame clause.  Yeah, it's bizarre that each of
>   * these cases works differently, but SQL:2008 says so; see 7.11
>   * <window clause> syntax rule 10 and general rule 1.  The frame
>   * clause rule is especially bizarre because it makes "OVER foo"
>   * different from "OVER (foo)", and requires the latter to throw an
>   * error if foo has a nondefault frame clause.	Well, ours not to
>   * reason why, but we do go out of our way to throw a useful error
>   * message for such cases.
>   */
> ```

So the subtility here is the difference between `OVER x` and `OVER (x)` - the first one executing the window function over window `x` and the latter one using a copied-and-modified version of `x`, which is not allowed for windows that specify a range. So in conclusion: reuse of named windows with range clauses can be allowed just fine, als long as they are rendered without copy operation if not inlined. As we currently inline the windows anyway, we should imo just allow reuse of windows that specify a range clause.

Not having seen the copy-and-modify syntax used to its full extend before, I decided to look into it more, and it seems that its even allowed between different named windows. Obviously with the aforementioned limitations in mind. 

For example:

```pgSQL
SELECT foo.*,
       bar.*,
       row_numbeR() OVER w
FROM (
      VALUES (1) , (2), (3)) AS foo(a),
     (
      VALUES (1), (2), (3)) AS bar(b) window x AS (PARTITION BY a),
                                             w AS (x
                                                   ORDER BY b ROWS BETWEEN unbounded preceding AND CURRENT ROW) ;
```

```
 a | b | row_number 
---+---+------------
 1 | 1 |          1
 1 | 2 |          2
 1 | 3 |          3
 2 | 1 |          1
 2 | 2 |          2
 2 | 3 |          3
 3 | 1 |          1
 3 | 2 |          2
 3 | 3 |          3
```

Above is allowed in postgres (copy-and-modify window definition for another named window)


The below is not:

```pgSQL
SELECT foo.*,
       bar.*,
       row_numbeR() OVER w
FROM (
      VALUES (1) , (2), (3)) AS foo(a),
     (
      VALUES (1), (2), (3)) AS bar(b) window x AS (PARTITION BY a ROWS BETWEEN unbounded preceding AND CURRENT ROW),
                                             w AS (x
                                                   ORDER BY b) ;
```

```
ERROR:  cannot copy window "x" because it has a frame clause
LINE 1: ...ween unbounded preceding and current row ) , w as ( x order ...
```

JWGmeligMeyling:
I've added a couple of query execution/translation tests (https://github.com/Blazebit/blaze-persistence/pull/834/commits/8db3010931c9175689a7b1b8e9933f7bae607443#diff-427304b646cd29ed38cf18d0c24051bfR482), most of which are currently failing. Please review them, we may not want to support all used language features in the tests. During your work on the parser we should decide which of the language features we want to support. I'll base the documentation on these language features. We can disable the other tests (if any).

beikov:
Would you mind changing uses of `WINDOW_SUM(...)` and `FUNCTION('WINDOW_SUM', ...)` to `SUM(...)`? If that's done and we have decent documentation for the window clause definition as well as the window syntax and available window functions, we can merge this.

JWGmeligMeyling:
> Would you mind changing uses of WINDOW_SUM(...) and FUNCTION('WINDOW_SUM', ...) to SUM(...)? 

Sure, but how does the parser differentiate between the SUM aggregation and the SUM window function then? And how does this apply to the other window functions, such as WINDOW_AVG ?

beikov:
> Sure, but how does the parser differentiate between the SUM aggregation and the SUM window function then? And how does this apply to the other window functions, such as WINDOW_AVG ? Or do you mean replacing FUNCTION("WINDOW_SUM",... ) with WINDOW_SUM(...) ?
No worries, I started working on that already. The parser differentiates based on whether there is an OVER clause.

JWGmeligMeyling:
I've pushed some documents but I'll add some more. Anything in particular that you still miss?

documentation/src/main/asciidoc/core/manual/en_US/18_jpql_functions.adoc
```diff
@@ -300,6 +300,38 @@ Aggregates/concatenates the values produced by `expression` to a single string s
...
+* `PERCENT_RANK`
+* `CUME_DIST`
+* `NTILE`
+* `LEAD`
+* `LAG`
+* `FIRST_VALUE`
+* `LAST_VALUE`
+* `NTH_VALUE`
+
+``WARNING: This function might not be supported by all DBMS, so make sure your target database does before using it
```
JWGmeligMeyling: Remove ``

documentation/src/main/asciidoc/core/manual/en_US/21_window_functions.adoc
```diff
@@ -0,0 +1,84 @@
...
+
+[source,java]
+----
+CriteriaBuilder<Tuple> criteria = cbf.create(em, Tuple.class)
+    .from(Person.class, "per")
+    .window("x").partitionBy("per.age").end()
+    .select("SUM(per.age) OVER (x ORDER BY per.id)")
+----
+
+WARNING: Note that the partition, order or range of a window definition can only bee specified if the base window does not specify any partition, order or range.
```
sranjha: "be specified"

documentation/src/main/asciidoc/core/manual/en_US/18_jpql_functions.adoc
```diff
@@ -300,6 +300,38 @@ Aggregates/concatenates the values produced by `expression` to a single string s
 
 WARNING: This function might not be supported by all DBMS, so make sure your target database does before using it
 
+===== WINDOW functions
```
beikov: Give a warning that this is an internal function and the syntax might change at any time. Refer to chapter 21 where the supported syntax is explained. See the PAGE_POSITION function description for a warning example

documentation/src/main/asciidoc/core/manual/en_US/18_jpql_functions.adoc
```diff
@@ -300,6 +300,38 @@ Aggregates/concatenates the values produced by `expression` to a single string s
...
+
+Syntax: `FUNCTION( 'WINDOW_XXX', arguments+, (, 'FILTER BY' (, filter_by_expression)+ )? (, 'PARTITION BY' (, partition_by_expression)+ )? (, 'ORDER BY' (, order_by_expression (, order_specification ) )+ )? ((, 'RANGE' | 'ROWS' | 'GROUPS') (, ( 'UNBOUNDED PRECEDING' | ( number_expression, 'PRECEDING') | 'CURRENT ROW'), 'AND', ( 'UNBOUNDED FOLLOWING' | ( number_expression, 'FOLLOWING') | 'CURRENT ROW' ) ) )`
+
+Where `order_specification` is one of `'ASC'`, `'DESC'`, `'ASC NULLS FIRST'`, `'ASC NULLS LAST'`, `'DESC NULLS FIRST'`, `'DESC NULLS LAST'`.
+
+For every aggregate function, there is a window function.
```
beikov: You forgot COUNT

documentation/src/main/asciidoc/core/manual/en_US/21_window_functions.adoc
```diff
@@ -0,0 +1,84 @@
...
+
+
+Window functions provide the ability to perform aggregate calculations across sets of rows that are related to the current query row.
+Unlike regular aggregate functions, use of a window function does not cause rows to become grouped into a single output row.
+
+WARNING: This function might not be supported by all DBMS, so make sure your target database does before using it
+
+=== Aggregate window functions
+
+Any built-in aggregate function can be used as a window function. These are:
```
beikov: You forgot COUNT

JWGmeligMeyling:
Looks good! 👍

entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/metamodel/AbstractAttribute.java
```diff
@@ -392,7 +392,7 @@ public static String replaceThisFromMapping(String mapping, String root) {
             return Collections.emptyMap();
         }
 
-        context.getTypeValidationExpressionFactory().createSimpleOrObjectExpression(expression, false).accept(visitor);
+        context.getTypeValidationExpressionFactory().createSimpleExpression(expression, false).accept(visitor);
```
Mobe91: @beikov Why has `createSimpleOrObjectExpression` been removed at this point?

beikov: There is a new form for this now `public Expression createSimpleExpression(String expression, boolean allowOuter, boolean allowQuantifiedPredicates, boolean allowObjectExpression)`

PR #865, opened 2019-10-06
[#864] Add transaction access SPI and implement it for the entity view spring integration. Also fix snapshot uploading

### LINKED ISSUE #864: Shared EntityManager exception when updating with an EnittyView
### Description
Creating an an `Entity` (using a JPA repository ) and trying to update it using an `EntityView` with the `EntityViewManager`

**Code/POC**

- https://github.com/EugenMayer/java-spring-pocs/tree/master/src/main/java/de/kontextwork/poc/spring/blaze
- https://github.com/EugenMayer/java-spring-pocs/blob/master/src/test/java/de/kontextwork/poc/spring/blaze/BlazeServiceTest.java

run it using

```
./gradlew bpTest --tests BlazeServiceTest.updateUserProfile
```

You will need to have the current master/HEAD build of BP, change the snapshot name accordingly to your local one at 

https://github.com/EugenMayer/java-spring-pocs/blob/master/build.gradle#L41

### Actual behavior
An exception is throwm

```
  java.lang.IllegalStateException: Could not access entity transaction!
      at de.kontextwork.poc.spring.blaze.BlazeServiceTest.updateUserProfile(BlazeServiceTest.java:83)
  Caused by: java.lang.IllegalStateException: Not allowed to create transaction on shared EntityManager - use Spring transactions or EJB CMT instead
      at de.kontextwork.poc.spring.blaze.BlazeServiceTest.updateUserProfile(BlazeServiceTest.java:83)
```

### Environment
Version: master HEAD on `f6405863e`   				
PA-Provider:       Hibernate 5.4.5
DBMS:               h2 / mariadb
Application Server: spring boot 2.1.9

beikov:
## Related Issue

Fixes #864

### DIFF
--- a/.travis-deploy.sh
+++ b/.travis-deploy.sh
@@ -8,7 +8,7 @@ if [ "$TRAVIS_REPO_SLUG" == "Blazebit/blaze-persistence" ] &&
     [ "x$JDK" == "x" ]; then
 
   echo "Starting snapshot deployment..."
-  mvn -s .travis-settings.xml -DperformRelease -DskipTests -Dgpg.skip=true -Dquiet=true deploy
+  mvn -s .travis-settings.xml -DperformRelease -DskipTests -Dgpg.skip=true -Dquiet=true clean deploy
   echo "Snapshots deployed!"
 
 else
--- a/entity-view/api/src/main/java/com/blazebit/persistence/view/spi/TransactionAccess.java
+++ b/entity-view/api/src/main/java/com/blazebit/persistence/view/spi/TransactionAccess.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2014 - 2019 Blazebit.
+ *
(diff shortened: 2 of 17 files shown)


EugenMayer:
Any reason to not include any test here?

beikov:
The current tests already cover the usage. In fact, we could remove the custom shared entity manager invocation handler that allowed TX access now since we implement this.

EugenMayer:
That is odd, if the current tests already have covered this case, why did they not detect the issue? Maybe i missunderstand something

beikov:
The repository tests rely on the TX support. Previously, the hibernate TX strategy would have been used which requires `EntityManager.getTransaction()` to be accessible which is why we had the custom shared entity manager invocation handler implementation. The new strategy that is used now integrates with Spring managed TX and doesn't require that anymore, so the ultimate test would be to remove the custom shared entity manager invocation handler.

PR #877, opened 2019-10-14
[#876] Remove custom SharedEntityManagerCreator as it became unnecessary thanks to a good integration

### LINKED ISSUE #876: Remove SharedEntityManagerCreator
We have an integration with spring managed transactions now, so we don't need that anymore. In fact, it looks like this is causing problems now that we properly integrate. Just remove SharedEntityManagerCreator

beikov:
## Related Issue

Fixes  #876


integration/spring-data/base/src/main/java/com/blazebit/persistence/spring/data/base/repository/AbstractEntityViewAwareRepository.java
```diff
@@ -71,6 +71,7 @@
  * @author Christian Beikov
  * @since 1.2.0
  */
+@Transactional(readOnly = true)
```
JWGmeligMeyling: Is this sensible for updatable EV's?

beikov: Yeah, the write operations have a normal transactional annotation

PR #881, opened 2019-10-16
Exclude MappedSuperclassType from IdMapping checks

heruan:
## Description
`MappedSuperclassType` is an `IdentifiableType` but could also be an embeddable, so it's not always necessary for its views to have an `@IdMapping`.


beikov:
Not sure what this failure is about, but it's there: https://travis-ci.org/Blazebit/blaze-persistence/jobs/598821478#L2321

heruan:
Should be fixed now.

PR #885, opened 2019-10-17
Fix metamodel generation for NotFound(Ignore) ManyToOne associations

### LINKED ISSUE #886: Update and fix integration issues with Spring data 2.2
We testsed against an RC version of Spring Data 2.2. Let's update to the release version and see if everything still works fine.

### LINKED ISSUE #887: Metamodel generation fails for "nullable" ManyToOne associations
### Description

Metamodel generation fails for "nullable" ManyToOne associations. This happens for ManyToOne associations which are mapped with `@Ignore(NotFound)` since Hibernate 5.4.x, where nested nullable properties are omitted from the metamodel as property dereference is not safe in this case.

### Steps to reproduce

Example metamodel:

```java

@Entity
public static class ItemEntity {

    @Id
    String id;

}

@Entity
@Table(name = "problem")
public static class ProblemEntity {

    @Id
    String id;

    @NotFound(action = NotFoundAction.IGNORE)
    @ManyToOne
    ItemEntity associatedItemEntity;
}
```

### Environment

Version:            1.4.0-Alpha2
JPA-Provider:       Hibernate 5.4.x
DBMS:               N/a
Application Server: N/a

### LINKED ISSUE #888: ID dereference for ManyToOne association property in joined inheritance hierarchy using group join
### Description

A non-“nullable” but optional many to one association is treated as if it can not be dereferenced, thus requiring a subselect or join.

### Expected behavior

The value to be dereferenced, and no subquery being rendered in the query. I.e.:

```sql
SELECT subB_0.id,
       subA_0.id
FROM DereferenceOptionalManyToOneAssociationTest$SubB subB_0
JOIN DereferenceOptionalManyToOneAssociationTest$SubA subA_0 ON (subA_0.id = subB_0.parent.id)
```

### Actual behavior
An unneeded subquery being rendered in the query:

```sql
SELECT subB_0.id,
       subA_0.id
FROM DereferenceOptionalManyToOneAssociationTest$SubB subB_0
JOIN DereferenceOptionalManyToOneAssociationTest$SubA subA_0 ON (EXISTS
                                                                   (SELECT 1
                                                                    FROM subB_0.parent _synth_subquery_0
                                                                    WHERE subA_0.id = _synth_subquery_0.id))
```

### Steps to reproduce

Metamodel:

```java
    @Id
    @Column(name = "scope_id")
    @GeneratedValue(strategy = GenerationType.IDENTITY)
    private Long id;

    @ManyToOne(targetEntity = Scope.class, cascade = {CascadeType.PERSIST})
    @JoinColumn(name = "parent_scope_id", referencedColumnName = "scope_id")
    private ParentScopeType parent;
```

Query fragment:

```java
cbf.create(em, Tuple.class)
                .from(SubB.class, "subB_0")
                .innerJoinOn(SubA.class, "subA_0").on("subA_0.id").eqExpression("subB_0.parent.id").end()
                .select("subB_0.id")
                .select("subA_0.id")
```

### Environment

Version:            1.4.0.Alpha2
JPA-Provider:       Hibernate 5.4.x
DBMS:               n/a
Application Server: n/a

JWGmeligMeyling:
## Related Issue

Fixes #886. Fixes #887. Fixes #888

### DIFF
--- a/core/testsuite/src/test/hibernate/com/blazebit/persistence/testsuite/DereferenceManyToOneAssociationInJoinedInheritanceHierarchyTest.java
+++ b/core/testsuite/src/test/hibernate/com/blazebit/persistence/testsuite/DereferenceManyToOneAssociationInJoinedInheritanceHierarchyTest.java
@@ -1,0 +1,303 @@
+/*
+ * Copyright 2014 - 2019 Blazebit.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.blazebit.persistence.testsuite;
+
+import com.blazebit.persistence.CriteriaBuilder;
+import com.blazebit.persistence.testsuite.base.jpa.category.NoHibernate42;
+import com.blazebit.persistence.testsuite.base.jpa.category.NoHibernate43;
+import com.blazebit.persistence.testsuite.base.jpa.category.NoHibernate50;
+import com.blazebit.persistence.testsuite.base.jpa.category.NoHibernate51;
+import com.blazebit.persistence.testsuite.base.jpa.category.NoHibernate52;
+import com.blazebit.persistence.testsuite.base.jpa.category.NoHibernate53;
+import org.hibernate.annotations.ForeignKey;
+import org.junit.Assume;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.FetchType;
+import javax.persistence.Id;
+import javax.persistence.Inheritance;
+import javax.persistence.InheritanceType;
+import javax.persistence.ManyToOne;
+import javax.persistence.Tuple;
+import java.util.List;
+import java.util.Properties;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * @author Jan-Willem Gmelig Meyling
+ * @since 1.4.0
+ */
+public class DereferenceManyToOneAssociationInJoinedInheritanceHierarchyTest extends AbstractCoreTest {
+
+    @Override
+    protected Properties applyProperties(Properties properties) {
+        // Hibernate 5.3+ broke queries on JOINED inheritance hierarchies.
+        properties.put("hibernate.query.omit_join_of_superclass_tables", "false");
+        return super.applyProperties(properties);
+    }
+
+    @Before
+    public void setUp() {
+        SubA a_1 = new SubA(1L);
+        SubA a_2 = new SubA(2L);
+        SubA a_3 = new SubA(3L);
+        SubA a_14 = em.getReference(SubA.class, 10L);
+        SubB b_4 = new SubB(4L, null);
+        SubB b_5 = new SubB(5L, a_3);
+        SubB b_6 = new SubB(6L, b_4);
+        SubB b_7 = new SubB(7L, a_14);
+
+        em.merge(a_1);
+        em.merge(a_2);
+        em.merge(a_3);
+        em.merge(b_4);
+        em.merge(b_5);
+        em.merge(b_6);
+        em.merge(b_7);
+    }
+
+    @Test
+    @Category({
+            // Although unused, this test dependent on the availability of group join support introduced in Hibernate 5.2.8
+            NoHibernate42.class, NoHibernate43.class, NoHibernate51.class
+    })
+    public void testRootTypeJoinWithGroupJoins() {
+        Assume.assumeTrue(supportsTableGroupJoins());
+
+        CriteriaBuilder<Tuple> criteriaBuilder = cbf.create(em, Tuple.class)
+                .from(SubB.class, "subB_0")
+                .leftJoinOn(Super.class, "subA_0").on("subA_0.id").eqExpression("subB_0.parent.id").end()
+                .select("subB_0")
+                .select("subA_0.id")
+                .orderByAsc("subB_0.id")
+                .orderByAsc("subA_0.id");
+        String queryString = criteriaBuilder
+                .getQueryString();
+
+        assertFalse("Query contains unexpected EXIST clause: " + queryString,
+                queryString.toUpperCase().contains("EXIST"));
+
+        List<Tuple> resultList = criteriaBuilder.getResultList();
+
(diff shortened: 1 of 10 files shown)


core/testsuite/src/test/java/com/blazebit/persistence/testsuite/SingleValuedAssociationOneToOneJoinTableTest.java
```diff
@@ -18,12 +18,16 @@
...
+import com.blazebit.persistence.testsuite.base.jpa.category.NoEclipselink;
+import com.blazebit.persistence.testsuite.base.jpa.category.NoHibernate;
+import com.blazebit.persistence.testsuite.base.jpa.category.NoOpenJPA;
 import com.blazebit.persistence.testsuite.entity.DocumentForOneToOneJoinTable;
 import com.blazebit.persistence.testsuite.entity.DocumentForOneToOne;
 import com.blazebit.persistence.testsuite.entity.DocumentInfo;
 import org.junit.Assert;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
+import org.mockito.cglib.proxy.NoOp;
```
beikov: Remove

JWGmeligMeyling:
Requires a bit more work, will provide a new PR later today.

integration/hibernate-base/src/main/java/com/blazebit/persistence/integration/hibernate/base/HibernateJpaProvider.java
```diff
@@ -484,19 +487,27 @@ public boolean isForeignJoinColumn(EntityType<?> ownerType, String attributeName
...
-            OneToOneType oneToOneType = (OneToOneType) propertyType;
-            // It is foreign if there is a mapped by attribute
-            // But as of Hibernate 5.4 we noticed that we have to treat nullable one-to-ones as "foreign" as well
-            return (oneToOneType).getRHSUniqueKeyPropertyName() != null || isNullable(oneToOneType);
+        if (propertyType instanceof org.hibernate.type.EntityType) {
+            org.hibernate.type.EntityType entityType = (org.hibernate.type.EntityType) propertyType;
+            // As of Hibernate 5.4 we noticed that we have to treat nullable associations as "foreign" as well
+            if (isNullable(entityType)) {
+                return true;
+            }
```
JWGmeligMeyling: This is the fix for #887

integration/hibernate-base/src/main/java/com/blazebit/persistence/integration/hibernate/base/HibernateJpaProvider.java
```diff
@@ -484,19 +487,27 @@ public boolean isForeignJoinColumn(EntityType<?> ownerType, String attributeName
...
+
+            // OneToOnes can't have JoinTables as per spec
+            // ManyToOnes can have JoinTables, which can be treated as non-foreign
+            // if table group joins are supported.
+            if (supportsTableGroupJoins || propertyType instanceof OneToOneType) {
```
JWGmeligMeyling: `supportsTableGroupJoins` is the implementation for #888

core/api/src/main/java/com/blazebit/persistence/spi/JpaProvider.java
```diff
@@ -523,6 +523,14 @@
      */
     public void setCacheable(Query query);
 
+    /**
+     * Returns whether the JPA provider supports table group joins.
+     *
+     * @return true if supported, otherwise false
+     * @since 1.4.0
+     */
+    boolean supportsTableGroupJoins();
```
beikov: I don't see a a reason for exposing this. Please remove it and place the relevant logic for hibernate in the `AbstractPersistenceTest` class of the hibernate testsuite base project.

core/testsuite/src/test/hibernate/com/blazebit/persistence/testsuite/DereferenceManyToOneAssociationInJoinedInheritanceHierarchyTest.java
```diff
@@ -0,0 +1,183 @@
...
+import java.util.Properties;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * @author Jan-Willem Gmelig Meyling
+ * @since 1.4.0
+ */
+public class DereferenceManyToOneAssociationInJoinedInheritanceHierarchyTest extends AbstractCoreTest {
```
beikov: IMO the tests need to make use of data to verify the desired behavior. The assertions should be the same for all the tests.

A/B | id | parent_id
A    |  1  | null
A    |  2  | null
A    |  3  | 1
B    |  4  | null
B    |  5  | 3
B    |  6  | 4  <-- wrong data on purpose
B    |  7  | 10  <-- wrong data on purpose

Results

C1 | C2
4   |  null
5   |  3
6   |  null
7   |  null

JWGmeligMeyling: Ive verified the generated SQL for all versions of Hibernate, but I can add some additional assertions too.

core/api/src/main/java/com/blazebit/persistence/spi/JpaProvider.java
JWGmeligMeyling: Im not really using it for anything but some test assumptions so I dont have a bunch of failing tests that shouldnt have been run for this category in the first place. I can remove the assumptions altogether, it just makes it a little bit harder to execute the tests outside of the Maven run profiles.

core/testsuite/src/test/hibernate/com/blazebit/persistence/testsuite/DereferenceManyToOneAssociationInJoinedInheritanceHierarchyTest.java
JWGmeligMeyling: I don't think the tests for the "wrong data" entries are sensible. The generic bound only affects setters and getters, not the actual target type. So in the schema / for hibernate its perfectly valid for a B to reference another B. Or do you just want to see that this B is ommitted, as we're actively joining against A?

core/api/src/main/java/com/blazebit/persistence/spi/JpaProvider.java
beikov: You can still use the assumptions, just move the logic for `supportsTableGroupJoins` to `AbstractPersistenceTest`

core/testsuite/src/test/hibernate/com/blazebit/persistence/testsuite/DereferenceManyToOneAssociationInJoinedInheritanceHierarchyTest.java
beikov: Correct, B(6) and B(7) are IMO edge cases that might produce wrong results when implemented in a wrong way

beikov:
There are test failures because you use `@JoinColumn(foreignKey)` which was only added in JPA 2.1 and Hibernate 4.2 is still on JPA 2.0. Since the problematic tests are hibernate only, you can use the hibernate proprietary annotation `@ForeignKey` to achieve the same.

JWGmeligMeyling:
Hmm Hibernates `@ForeignKey` seems to miss an equivalent for `ConstraintMode` :(

beikov:
The name `none` is special and treated like "NO_CONSTRAINT"

PR #900, opened 2019-11-01
Inherit range clause from base window definition and adjust rendering

jwgmeligmeyling:
for it

### DIFF
--- a/core/impl/src/main/java/com/blazebit/persistence/impl/function/window/AbstractWindowFunction.java
+++ b/core/impl/src/main/java/com/blazebit/persistence/impl/function/window/AbstractWindowFunction.java
@@ -240,7 +240,7 @@ public abstract class AbstractWindowFunction implements JpqlFunction {
             context.addChunk(" ");
         }
         renderOrderBy(context, windowFunction.getOrderBys());
-        if (windowFunction.getFrameMode() != null && !windowFunction.getOrderBys().isEmpty()) {
+        if (windowFunction.getFrameMode() != null && (!windowFunction.getOrderBys().isEmpty() || !windowFunction.getPartitionExpressions().isEmpty())) {
             context.addChunk(" ");
         }
         renderFrame(context, windowFunction);
--- a/core/testsuite/src/test/java/com/blazebit/persistence/testsuite/WindowFunctionTest.java
+++ b/core/testsuite/src/test/java/com/blazebit/persistence/testsuite/WindowFunctionTest.java
@@ -17,24 +17,20 @@
 package com.blazebit.persistence.testsuite;
 
 import com.blazebit.persistence.CriteriaBuilder;
-import com.blazebit.persistence.testsuite.base.jpa.category.NoDB2;
 import com.blazebit.persistence.testsuite.base.jpa.category.NoMSSQL;
 import com.blazebit.persistence.testsuite.base.jpa.category.NoMySQLOld;
 import com.blazebit.persistence.testsuite.base.jpa.category.NoOracle;
-import com.blazebit.persistence.testsuite.entity.Document;
 import com.blazebit.persistence.testsuite.entity.Person;
-import com.blazebit.persistence.testsuite.entity.Version;
 import com.blazebit.persistence.testsuite.tx.TxVoidWork;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
 
 import javax.persistence.EntityManager;
 import javax.persistence.Tuple;
-
 import java.util.List;
 
-import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
 
 /**
  *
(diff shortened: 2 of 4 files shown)


core/testsuite/src/test/java/com/blazebit/persistence/testsuite/WindowFunctionTest.java
```diff
@@ -579,6 +575,14 @@ public void testParsedNamedWindowDefinitionWithRange() {
                 .orderByAsc("per.age")
                 ;
 
+        String queryString = criteria.getQueryString();
```
jwgmeligmeyling: Without the port, the frame clause is omitted from the function parameters. So this test shows the problem, and the added assertion will fail without the fix.

core/testsuite/src/test/java/com/blazebit/persistence/testsuite/WindowFunctionTest.java
```diff
@@ -579,6 +575,14 @@ public void testParsedNamedWindowDefinitionWithRange() {
...
+
+        assertEquals("SELECT per.age, window_sum(per.age,'ORDER BY',per.age,'ASC NULLS LAST', " +
```
jwgmeligmeyling: Not sure if we're completely consistent w.r.t. the rendering of spaces here. Shouldn't this be:

```suggestion
        assertEquals("SELECT per.age, window_sum(per.age, 'ORDER BY', per.age, 'ASC NULLS LAST', " +
```
?

core/testsuite/src/test/java/com/blazebit/persistence/testsuite/WindowFunctionTest.java
```diff
@@ -579,6 +575,14 @@ public void testParsedNamedWindowDefinitionWithRange() {
...
+                        "'ROWS','BETWEEN','UNBOUNDED PRECEDING','AND','CURRENT ROW') " +
+                        "FROM Person per WINDOW x AS (ORDER BY per.age ASC  NULLS " +
```
jwgmeligmeyling: Same question here:
```suggestion
                        "FROM Person per WINDOW x AS (ORDER BY per.age ASC NULLS " +
```
?

core/impl/src/main/java/com/blazebit/persistence/impl/function/window/AbstractWindowFunction.java
```diff
@@ -240,7 +240,7 @@ protected void render(FunctionRenderContext context, WindowFunction windowFuncti
             context.addChunk(" ");
         }
         renderOrderBy(context, windowFunction.getOrderBys());
-        if (windowFunction.getFrameMode() != null && !windowFunction.getOrderBys().isEmpty()) {
+        if (windowFunction.getFrameMode() != null && (!windowFunction.getOrderBys().isEmpty() || !windowFunction.getPartitionExpressions().isEmpty())) {
```
jwgmeligmeyling: The issue is a bit artificial, because one probably shouldn't use an order by without a partition by clause, but if one does, the space wasn't added resuling in an error (because it renders `ORDER BY columnnameROWS` in SQL)

core/testsuite/src/test/java/com/blazebit/persistence/testsuite/WindowFunctionTest.java
beikov: That's on purpose. It's internal, so no need for assertions about that here.

core/testsuite/src/test/java/com/blazebit/persistence/testsuite/WindowFunctionTest.java
beikov: If it's easy to fix let's do it.

core/testsuite/src/test/java/com/blazebit/persistence/testsuite/WindowFunctionTest.java
beikov: Should use `function('window_sum', ...` for eclipselink: https://travis-ci.org/Blazebit/blaze-persistence/jobs/606361914#L1536

core/testsuite/src/test/java/com/blazebit/persistence/testsuite/WindowFunctionTest.java
beikov: Good catch, thanks!

jwgmeligmeyling:
I Will fix the remaining issues later today

beikov:
Not sure you saw it, but there are test failures:

- MySQL8: https://travis-ci.org/Blazebit/blaze-persistence/jobs/609789435#L1715
- MSSQL: https://travis-ci.org/Blazebit/blaze-persistence/jobs/609789437#L1719
- DB2: https://travis-ci.org/Blazebit/blaze-persistence/jobs/609789438#L1749
- EclipseLink: https://travis-ci.org/Blazebit/blaze-persistence/jobs/609789446#L1537

jwgmeligmeyling:
Ah crap, I hoped the assertion would be fixed by using `AbstractCoreTest#function` and `AbstractCoreTest#renderNullPrecedence`. I'll change the assertion once again.

PR #911, opened 2019-11-07
[#373] Add support for deserializing creatable and updatable entity views via Jackson in Spring MVC and JAX-RS

### LINKED ISSUE #373: Integrate entity views with Jackson in Spring MVC and JAX-RS
It should be possible to consume creatable and updatable entity views via Jackson in Spring MVC and JAX-RS.

beikov:
This is not yet done, just a preview

## Related Issue

Fixes #373


EugenMayer:
This would be plain awesome!

PR #922, opened 2019-11-09
spring-data-rest - fix the way sortResolver is determined - us springs exposed bean

EugenMayer:
Trying to fix the data-rest integration


beikov:
In `com.blazebit.persistence.spring.data.testsuite.DocumentControllerTest` if you use the following instead, the tests should work:
`@ContextConfiguration(classes = {DocumentControllerTest.TestConfig.class, BlazePersistenceWebConfiguration.class, SpringDataWebConfiguration.class})`

This will add the sort resolver bean from spring data. Could you please add that, then this could be merged.

beikov:
We need this kind of change also in `com.blazebit.persistence.examples.spring.data.rest.repository.SampleTest`

PR #934, opened 2019-11-13
[#774] Fix limit rendering in main query if preceded by a returning c…

### LINKED ISSUE #774: setMaxResults does not work with CriteriaBuilder involving returning CTE queries
```java
cbf.create(entityManager, Tuple.class)
.withReturning(MyCte.class)
.end()
.from(VendorGlobalCompany.class)
.select("bla")
.setMaxResults(10)
```
In this case the `limit` is rendered after the `returning` keyword inside the CTE and not at the global query level.

This is because the `com.blazebit.persistence.impl.dialect.PostgreSQLDbmsLimitHandler` searches for the `returning` keyword globally in the sql query string and appends the `limit` after the found `returning`. So this may be only relevant for Postgres.

Mobe91:
fixes #774

### DIFF
--- a/core/testsuite/src/test/java/com/blazebit/persistence/testsuite/LimitTest.java
+++ b/core/testsuite/src/test/java/com/blazebit/persistence/testsuite/LimitTest.java
@@ -17,8 +17,22 @@
 package com.blazebit.persistence.testsuite;
 
 import com.blazebit.persistence.CriteriaBuilder;
+import com.blazebit.persistence.testsuite.base.jpa.category.NoDB2;
+import com.blazebit.persistence.testsuite.base.jpa.category.NoDatanucleus;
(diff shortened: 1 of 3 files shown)


core/testsuite/src/test/java/com/blazebit/persistence/testsuite/LimitTest.java
```diff
@@ -91,4 +115,25 @@ public void testSubqueryAndOuterQueryLimit() {
         assertEquals(1, result.size());
         assertEquals("P1", result.get(0).getName());
     }
+
+    // Test for issue #774
+    @Test
+    @Category({NoH2.class, NoDB2.class, NoFirebird.class, NoMSSQL.class, NoMySQL.class, NoOracle.class, NoSQLite.class})
```
beikov: Also disable it for Eclipselink, Datanucleus and OpenJPA

PR #955, opened 2019-12-18
[#633] Allow to flush updatable entity views to entities directly

### LINKED ISSUE #471: Treat on super type too restrictive, should allow subtypes too
### Description

When having an entity model like
```java
@Entity class A{}
@Entity class B extends A{}
@Entity class C extends B{}
@Entity class D extends B{}
```

A treat expression like `TREAT(x AS B)` currently is too strict.

### Expected behavior

Objects of type C and D are also returned.

### Actual behavior

Only objects of type B are returned, because the type discriminator predicate checks for equality.

### Environment

Version:            1.2.0.Alpha3-SNAPSHOT
JPA-Provider:       Hibernate
DBMS:               N/A
Application Server: N/A

### LINKED ISSUE #633: Allow to flush updatable entity views to entities directly
Add an API that allows to flush dirty changes to an existing entity instance.

### LINKED ISSUE #809: Throw exception when setFirstResult or setMaxResults is used in a correlation provider other than select with batch size 1
### Description

We had that issue a few times now already. A `CorrelationProvider` that uses setMaxResults which leads to wrong results. IMO we should throw an exception.

### Environment

Version:            
JPA-Provider:       
DBMS:               
Application Server:

beikov:
## Related Issue

Fixes #633. Fixes  #809. Fixes #471

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ Not yet released
 * Official support for Java 13 and 14
 * Add support for enum literals in builder API
 * Allow to flush updatable entity views to entities directly
+* Throw exception when `setFirstResult` or `setMaxResults` is used in a `CorrelationProvider` other than select with batch size 1
 
 ### Bug fixes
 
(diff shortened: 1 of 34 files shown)


beikov:
Merged via 535e844b49d29df0230f77dceeed74d94978e7ae

PR #956, opened 2019-12-19
[#678] Allow entity view conversion with fine grained control of attribute subview types and fix a few smaller issues

### LINKED ISSUE #678: Allow entity view conversion with fine grained control of attribute subview types
Currently, we only convert the top level type and just copy over the objects with the declared types but this is limiting as it is often required to copy whole trees for i.e. duplicating objects. One option for now is to declare the creatable types in the creatable root view, but that is often cumbersome. I propose the following API

```
evm.convertWith(view, CreateView.class)
  .convertAttribute("subview", SubviewCreateView.class)
  .convertAttribute("collection.subview", SubSubviewCreateView.class)
.convert();
```

Any thoughts?

beikov:
## Related Issue

Fixes #678


beikov:
Merged via b956cb54669a39d8d9ff28840c3eb3c45588af9a

PR #960, opened 2019-12-28
Support for CTE's for JPA Criteria API integration

jwgmeligmeyling:
@beikov I had to do quite some heavy work on the CTE criteria interface structure. I decided to stay close to the criteria builder API. Some interfaces are yet unnecessary, but will become important if we start adding set operations, which should not be available in i.e. recursive CTEs.

I ported two tests from the existing CTETest (one partially) as a proof of concept for the CTE api. The API is very much based on the existing criteria API - which is not particularly fluent. I'll see how the fluency of the API can be improved by additional extensions/helpers after we have more of the  puzzle pieces together.

Please review the general structure before I proceed.


beikov:
Looks good so far, but I wonder what would happen if a path created via a CTE query would be used somewhere othe than for a binding. Due to this, I'd rather not expose the path directly.
Also, in general, I think we should first think about a way to do set operations for the JPA Criteria API and then see how we can make recursive CTEs work with that.

jwgmeligmeyling:
> Looks good so far, but I wonder what would happen if a path created via a CTE query would be used somewhere othe than for a binding. Due to this, I'd rather not expose the path directly.

Hmm. It's still technically a Path however, just not a Path that references any of the query roots. I'd rather just throw an exception if we detect a Path that references an unreferenced query root. This allows us to keep the API analog to how modification queries work which I think is important.

> Also, in general, I think we should first think about a way to do set operations for the JPA Criteria API and then see how we can make recursive CTEs work with that.

Recursive CTE's don't support set operations through the `SetBuilder` though (It does via the `SelectRecursiveCTECriteriaBuilder`), only the `FullSelectCTECriteriaBuilder` does implement `SetBuilder`. I think the implementation of recursive CTE's is going to be similar but separate from set operations. I am working on a POC for set operations now.

PR #967, opened 2020-01-04
[#248, #249, #716, #717] id-query extraction, reusing CriteriaBuilder for CTE and count/id-query embedding into object query

### LINKED ISSUE #248: Embedd id query into object query in PaginatedCriteriaBuilder
Implement embedding of the id query into the object query as the new default strategy for PaginatedCriteriaBuilder and create a configuration option for disabling this.

### LINKED ISSUE #249: Allow embedding of count query into id/object query
Allow to embedd the count query into the id/object query with activating a configuration property.

### LINKED ISSUE #716: Allow extraction of CriteriaBuilder for id-query from PaginatedCriteriaBuilder
Provide a way to allow the extraction of an CriteriaBuilder from a PaginatedCriteriaBuilder that represents the id-query of the PaginatedCriteriaBuilder.

### LINKED ISSUE #717: Allow definition of CTE based on CriteriaBuilder
The current CTEBuilder interface shall be extended by a method
```java
public FullSelectCTECriteriaBuilder<T> with(CriteriaBuilder<T> sourceCriteriaBuilder);
```
This method returns a CTE builder that mimics query represented by the passed `sourceCriteriaBuilder`. Any select clauses present on the  `sourceCriteriaBuilder` must be discarded.

beikov:
## Related Issue

Fixes #716. Fixes #717. Fixes #248. Fixes #249


core/api/src/main/java/com/blazebit/persistence/CTEBuilder.java
```diff
@@ -36,6 +36,16 @@
...
 
+    /**
+     * Creates a builder for a CTE with the given CTE type with the given criteria builder as basis.
+     *
+     * @param cteClass The type of the CTE
+     * @param criteriaBuilder The criteria builder to copy the query from
+     * @return The CTE builder
+     * @since 1.4.1
+     */
+    public FullSelectCTECriteriaBuilder<T> with(Class<?> cteClass, CriteriaBuilder<?> criteriaBuilder);
```
jwgmeligmeyling: At first I was confused as to what the difference is with `withCtesFrom`, but I think the idea behind this method is the following:

```suggestion
    public <X> FullSelectCTECriteriaBuilder<T> with(Class<X> cteClass, CriteriaBuilder<X> criteriaBuilder);
```

Right?

beikov: No, the CTE type does not have to match the passed `CriteriaBuilder` as the attribute bindings have to be done explicitly.

PR #968, opened 2020-01-06
[#854, #969] Using SELECT or SUBSELECT fetch strategies for element collections produces a null entry

### LINKED ISSUE #854: Using SELECT or SUBSELECT fetch strategies for element collections produces a null entry
### Description

The identity of an element collection entry is defined by the parent identifier and either the index or the whole element state. When SUBSELECT or SELECT fetching a list element collection, the index is not considered which is why a null elment is added to the entity view list AFAICT. Using the JOIN strategy is obviously a workaround but will not be applicable in all situations. IMO we should fetch the index as well as we correlate the `this` object anyway and base the nullability on that.

### Environment

Version:            
JPA-Provider:       
DBMS:               
Application Server:

### LINKED ISSUE #969: Single valued association id where the association is part of id class attributes isn't considered for uniqueness analysis
### Description

Pagination with the following model doesn't work

```java

    @Entity(name = "BasicEntity")
    public static class BasicEntity {
        @Id
        Long key1;
    }

    @Entity(name = "IdClassEntity")
    @IdClass(IdClassEntity.IdClassEntityId.class)
    public static class IdClassEntity {
        @Id
        @ManyToOne
        BasicEntity basicEntity;
        @Id
        Long key2;

        public static class IdClassEntityId implements Serializable {
            Long basicEntity;
            Long key2;
        }
    }

    @Test
    public void metaModelInstantiationWithNestedIdClassAssociationTest() {
        cbf.create(em, IdClassEntity.class)
                .orderByAsc("key2")
                .orderByAsc("basicEntity.key1")
                .page(0, 1)
                .getResultList();
    }
```

### Environment

Version:            
JPA-Provider:       
DBMS:               
Application Server:

beikov:
## Related Issue

Fixes #854. Fixes #969

### DIFF
--- a/core/impl/src/main/java/com/blazebit/persistence/impl/ConstantifiedJoinNodeAttributeCollector.java
+++ b/core/impl/src/main/java/com/blazebit/persistence/impl/ConstantifiedJoinNodeAttributeCollector.java
@@ -376,9 +376,17 @@ class ConstantifiedJoinNodeAttributeCollector extends VisitorAdapter {
             // TODO: at some point we should build an equivalence class to transitively propagate constantification
             // TODO: also, it would be nice if we could detect constantifications in disjuncts and work with that
             if (isConstant(predicate.getLeft())) {
-                predicate.getRight().accept(this);
+                if (isParameterOrLiteral(predicate.getRight())) {
+                    predicate.getLeft().accept(this);
+                } else {
+                    predicate.getRight().accept(this);
+                }
             } else if (isConstant(predicate.getRight())) {
-                predicate.getLeft().accept(this);
+                if (isParameterOrLiteral(predicate.getLeft())) {
+                    predicate.getRight().accept(this);
+                } else {
+                    predicate.getLeft().accept(this);
+                }
             }
         } finally {
             negated = originNegated;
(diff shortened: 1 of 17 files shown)


core/impl/src/main/java/com/blazebit/persistence/impl/ConstantifiedJoinNodeAttributeCollector.java
```diff
@@ -376,9 +376,17 @@ public void visit(EqPredicate predicate) {
             // TODO: at some point we should build an equivalence class to transitively propagate constantification
             // TODO: also, it would be nice if we could detect constantifications in disjuncts and work with that
             if (isConstant(predicate.getLeft())) {
-                predicate.getRight().accept(this);
+                if (isParameterOrLiteral(predicate.getRight())) {
+                    predicate.getLeft().accept(this);
+                } else {
+                    predicate.getRight().accept(this);
```
jwgmeligmeyling: Why is this dependent on the lhs not being a parameter or literal?

beikov: The idea of the visitor is to visit all PathExpression that should be considered constantified i.e. they are in an equality relationshipt with something that is constant. The problem is that a PathExpression might already be constant due to e.g. a parent being constantified. In such a case, we'd miss out on initializing the constantification information for that Path which is essential. Before, we didn't have that, but only considered literals and parameters. That's why we do a special check here.

PR #971, opened 2020-01-11
[#337, #495, #671, #781, #822, #970, #974, #975, #978, #979] Various fixes and support for inlining CTEs/Subqueries in FROM clause as well as LATERAL support

### LINKED ISSUE #337: Implement option for inlining CTEs/Subquery in from clause and lateral support
Especially for DBs like PostgreSQL where a CTE also is an optimization fence, it might be desireable to inline the subquery directly into the FROM clauses. This should be doable on a per from clause basis and obviously only works for non-recursive CTEs.
The option could be turned on by default for DBMS that don't support CTEs to emulate non-recursive CTEs. This should also be documented in _Chapter 14. DML statements_ and _Chapter 15. CTEs_

Be careful when implementing this. Reordering the CTE queries logically requires the same tricks as for https://github.com/Blazebit/blaze-persistence/issues/305 because of the parameters

Maybe we can also introduce new `from` and `join` methods, something like the following

* `fromSubquery(Class<?> cteClass)` returning a cte builder? conflict with entity version?
* `from(CriteriaBuilder<?> builder, String alias)`
* `fromSubquery(Class<?> entityClass, String alias)` returns a builder that binds all attributes of the entity automatically and doesn't allow an additional select clause or join fetching. DBMS will optimize away unused columns anyway, so this would be very convenient

Also think about allowing something like the following

* `fromTopOfPartition(Class<?> entityClass, String alias, int topCount).partition(String expression).orderBy(...)`

This will transparently implement the top values per category/partition for the user behind the scenes which is quite important for analytics queries. This will render:

```
select ...
from (
  select * from(
    select ..., row_number() over(partition by PARTITION order by ORDER) as rn
    from entity sub
  ) x
  where x.rn <= TOP_COUNT
) alias
```

or 

```
select ...
from entity alias join (
  select ..., row_number() over(partition by PARTITION order by ORDER) as rn
  from entity sub
) x on alias.id = x.id and x.rn <= TOP_COUNT
```

So this is a kind of wrapper/entity-function around the query but also injects the row_number part.

According to https://blog.jooq.org/2017/09/22/how-to-write-efficient-top-n-queries-in-sql/ it is better in PostgreSQL to use `lateral`, so it might be worth rendering the following

```
select ...
from entity alias lateral join (
  select sub.id
  from entity sub
  where alias.PARTITION = sub.PARTITION
  order by ORDER
  limit TOP_COUNT
) x on alias.id = sub.id
```

Maybe the following can be rendered as a last resort

```
select ...
from entity alias 
where alias.id in (
  select sub.id
  from entity sub
  where alias.PARTITION = sub.PARTITION
  order by ORDER
  limit TOP_COUNT
)
```

Support for `lateral` is probably not very hard after we have that. Maybe we could introduce lateral support as part of a PR for this issue. On the contrary, it easier to add lateral than subquery support, because for subqueries we will have to ensure they are not correlated. We need a validation for the subqueries in the from clause that throws when encountering a correlated reference.

Any thoughts on additional methods or naming?

### LINKED ISSUE #495: Enhancement: make line separator in CTE query string configurable
When using CTE queries, the produced CTE query string contains line separators. When I enable query logging by enabling `hibernate.show_sql` the query spans over multiple log lines. As I am using Logback / Slf4j, each line is prefixed with a timestamp / thread name etc. I therefore can't easily copy the SQL string (for example to inspect it or run an explain query against it). But it potentially has some side effects for monitoring as well (I haven't observed such however).

- Is the new added line separator required for some DB dialects?
- If not, can we make this pretty print behaviour configurable?
- If we do, should / could we reuse the `hibernate.format_sql` property (and possibly any Datanucleus equivalent) or should we ship our own property for this?

### LINKED ISSUE #671: Using parameter in SELECT clause not possible with Collection DML INSERT-SELECT
With the normal INSERT-SELECT API it's possible to put a parameter into the select clause because it is matched against the insert attribute. In the Collection DML API we introduce a synthetic select query for the bound expressions which is a _naked select_ and thus doesn't support parameters.
I'm not yet sure how to fix this, but a workaround is to introduce a value clauses for the parameters instead.

### LINKED ISSUE #781: Use JpqlFunction for NULL literal expressions
Currently we render `NULLIF(1,1)` but this has issues with set operations because the type of the resulting sql expression is integer and may not match with corresponding columns of other queries that are part of the set operation.

### LINKED ISSUE #822: Problems with setFirstResult/setMaxResults in subquery when in EXISTS predicate
When using a subquery with an exists predicate, using e.g. setFirstResult/setMaxResults will result in a function wrapping the subquery in the resulting JPQL/HQL. This is problematic for Hibernate as it will run into a parse error when within a CASE WHEN expression, maybe also without that.
Maybe we have to introduce a custom exists predicate support function to handle this? Something like `_exists(subquery) = 1`

### LINKED ISSUE #970: Introduce internal JPQL function for rendering of parameters for SELECT statements
The idea is to introduce a function e.g. `param(:param_0)` that can be used in the SELECT clause to render parameters. AFAIK Hibernate and some other JPA Providers don't support parameters in the SELECT clause which causes a few problems. The biggest problem might be the rendering of a enum literal. There is no "enum literal" in Hibernate, so we can't really use that in a non-comparison context. We could workaround that issue by using a special function to which we pass the value or maybe a predicate that we then deconstruct in the JPQLFunction to make this work.

I think this will fix https://github.com/Blazebit/blaze-persistence/issues/671 and als, I think we don't need a function. We could use the cast function for this which we would need anyway to be able to use the null value for a parameter.

### LINKED ISSUE #974: Prefix update statement set paths with alias
### Description

Seems we are not prefixing the set paths with the alias but that could prevent HQL/JPQL parsing issues when attribute names match keywords, so consider prefixing.

### Environment

Version:            
JPA-Provider:       
DBMS:               
Application Server:

### LINKED ISSUE #975: Indexed element collection action optimization fails with IOOBE
I have a private example project. Will update the issue when I analyzed it.

### LINKED ISSUE #978: ValuesStrategy.SELECT_UNION doesn't work on MySQL 8.0.19
As of MySQL 8.0.19 the optimizer does something new that coverd up a wrong query. We are using limit 1,N to skip over the first all-null query result that is necessary to name columns but in order for that to work, we need to wrap the whole thing and order by null first which we don't.

### LINKED ISSUE #979: Support field access for entities as well for VALUES clause usage
It was reported that using a `VALUES` clause doesn't work properly when using FIELD access.

```
The stacktrace is :
java.lang.NullPointerException
	at com.blazebit.reflection.PropertyPathExpression.getLeafObject(PropertyPathExpression.java:192)
	at com.blazebit.reflection.PropertyPathExpression.getValue(PropertyPathExpression.java:142)
	at com.blazebit.reflection.PropertyPathExpression.getNullSafeValue(PropertyPathExpression.java:134)
	at com.blazebit.persistence.impl.ValuesParameterBinder.bind(ValuesParameterBinder.java:49)
	at com.blazebit.persistence.impl.query.AbstractCustomQuery$ValuesParameter.bind(AbstractCustomQuery.java:433)
	at com.blazebit.persistence.impl.query.AbstractCustomQuery.bindParameters(AbstractCustomQuery.java:142)
	at com.blazebit.persistence.impl.query.CustomSQLTypedQuery.getResultList(CustomSQLTypedQuery.java:50)
```

beikov:
## Related Issue

Fixes #337. Fixes #495. Fixes #671. Fixes #781. Fixes #822. Fixes #970. Fixes #974. Fixes #975. Fixes #978. Fixes #979


core/api/src/main/java/com/blazebit/persistence/FromBaseBuilder.java
```diff
@@ -223,4 +223,58 @@
...
+
+    /**
+     * Like {@link FromBaseBuilder#fromEntitySubquery(Class, String)} with the
+     * alias equivalent to the camel cased result of what {@link Class#getSimpleName()} of the entity class returns.
+     *
+     * @param cteClass The entity type for the subquery in the FROM clause
+     * @return The CTE builder for the subquery in the FROM clause
+     * @since 1.4.1
+     */
+    public FullSelectCTECriteriaBuilder<X> fromEntitySubquery(Class<?> cteClass);
```
jwgmeligmeyling: I believe for the `fromEntityX` variants, the parameter should be renamed to `entityClass`.

```suggestion
     * @param entityClass The entity type for the subquery in the FROM clause
      * @return The CTE builder for the subquery in the FROM clause
      * @since 1.4.1
      */
     public FullSelectCTECriteriaBuilder<X> fromEntitySubquery(Class<?> entityClass);
```

Issue #409, opened 2017-04-13
Emulate TREAT operator with entity joins


beikov:
This is a follow up of #120 and #123 as this requires quite some additional efforts.
In case the treat operator is broken as it is the case with many JPA providers, we can only try to emulate it by doing entity joins.

beikov:
The emulation for this was mainly aimed to fix issues with EclipseLink, but it's impossible to fix this for EclipseLink as it has a broken entity join rendering as well which makes this impossible.

PR #1006, opened 2020-02-18
[#1005] Contradictory where condition generated for multiple TREATs and discriminator column

Mobe91:


jwgmeligmeyling:
I have a feeling this is caused by limitations of Hibernate with respect to the implementation of treat. Perhaps this is a good test case where the optimization from #409 should be applied?

jwgmeligmeyling:
@beikov if rewriting TREAT paths to joins is too difficult for now, perhaps we can simply omit them during rendering and rely on the parameter being implicitly accessible through the polymorphism anyway?

beikov:
@jwgmeligmeyling this is an emulation already, but using the implicit downcast feature of HQL. I didn't look too deep yet, but this sounds like a simple bug in the emulation code.

beikov:
Superseded by https://github.com/Blazebit/blaze-persistence/pull/1015

Issue #1029, opened 2020-03-23
updateCollection API not working with CTE inlining


Mobe91:
The `_collection` alias referenced by inlined CTEs is undefined.

PR #1032, opened 2020-03-23
[#1029] Test case for updateCollection with inline CTE

Mobe91:
Test case for #1029


beikov:
Moved over to https://github.com/Blazebit/blaze-persistence/pull/1031

PR #1052, opened 2020-04-20
Issues/entity view listener

Mobe91:
Fixes issues with global entity view listeners.


Mobe91:
As stated in the commit message, the POM changse should fix the IDEA import errors that were introduced in a previous PR.

PR #1071, opened 2020-05-08
[#1070] QueryDSL integration

### LINKED ISSUE #1070: QueryDSL integration
### Description

A lot of projects use QueryDSL for generating JPQL/HQL queries, rather than dealing with the Criteria API or JPQL directly. For this userbase, dealing with the CriteriaBuilder API may not be so straightforward. A lot of QueryDSL users also face the limitations of plain JPQL/JPA, so it would be great to introduce this user base to the rich feature set of Blaze-Persistence. This integration allows a smooth transition for existing QueryDSL users to Blaze-Persistence. 

### Environment

Version:            
JPA-Provider:       
DBMS:               
Application Server:

jwgmeligmeyling:
## Description

This PR provides my initial contribution to the QueryDSL integration. This extends QueryDSL's API so that queries can be executed through Blaze-Persist, making its rich feature set available to QueryDSL domains.

The implementation is tested through integration tests which are derived from similar existing test scenarios. The implementation is also battle-tested in one of my projects for which I now run all QueryDSL queries through this integration.

*I have not abandoned my work on the fluent API version that is supposed to come in a separate module. However, due to time constraints (this PR is already taking much longer than I wanted to - if been working on this on and off since late December...), I am forced to prioritize. I really rather work on the metamodel changes required for
 Hibernate Composite User Type and JSON path dereference support first. As the point when I get back to it, we may already have the static metamodel in BP itself which also decreases its value relatively.*

## Related Issue
Fixes #1070

## Motivation and Context

From the original issue:

> A lot of projects use QueryDSL for generating JPQL/HQL queries, rather than dealing with the Criteria API or JPQL directly. For this userbase, dealing with the CriteriaBuilder API may not be so straightforward. A lot of QueryDSL users also face the limitations of plain JPQL/JPA, so it would be great to introduce this user base to the rich feature set of Blaze-Persistence. This integration allows a smooth transition for existing QueryDSL users to Blaze-Persistence.


documentation/src/main/asciidoc/core/manual/en_US/querydsl.adoc
```diff
@@ -0,0 +1,298 @@
...
+
+This module aims an API that is as close to the original QueryDSL API as possible.
+Where features did not exist in `querydsl-jpa`, but did exist in `querydsl-sql`, we stayed as close to the existing SQL implementation as close as possible.
+This includes the implementation for window functions, common table expressions (CTE's) and union queries which was the basis for all types of set expressions.
+
+Staying close to QueryDSL's API however, also means that the API is not as fluent as Blaze-Persistence users are accustomed to.
+This means that creating common table expressions or complex set operations may lead to superfluous code.
+
+
+[[spring-data-setup]]
```
jwgmeligmeyling, line 22:
```suggestion
[[querydsl-setup]]
```

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/WindowFirstLast.java
```diff
@@ -0,0 +1,146 @@
...
+        value = null;
+        Collections.addAll(this.orderBy, orderBy);
+        return this;
+    }
+
+    SimpleExpression<T> getValue() {
+        if (value == null) {
+            if (orderBy.isEmpty()) {
+                // TODO this check should be static
+                throw new IllegalStateException("No order by arguments given");
```
beikov, line 126: Is that really necessary? I think in SQL if you use an empty over clause like in `first_value(x) over()` it will use the ordering of the query.

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/WindowFirstLast.java
```diff
@@ -0,0 +1,146 @@
...
+            ImmutableList.Builder<Expression<?>> args = ImmutableList.builder();
+            StringBuilder builder = new StringBuilder();
+            builder.append("{0} keep (dense_rank ");
+            args.add(target);
+            builder.append(first ? "first " : "last ");
+            builder.append(ORDER_BY);
+            builder.append("{1}");
+            args.add(ExpressionUtils.orderBy(orderBy));
+            builder.append(")");
+            value = Expressions.template(target.getType(), builder.toString(), (List<Expression<?>>) args.build());
```
beikov, line 137: Please create two static final string templates for the first and last variant and conditionally use that here instead of building that up on demand.

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/WindowFirstLast.java
```diff
@@ -0,0 +1,146 @@
...
+     * @since 1.5.0
+     */
+    public WindowFirstLast<T> orderBy(OrderSpecifier<?>... orderBy) {
+        value = null;
+        Collections.addAll(this.orderBy, orderBy);
+        return this;
+    }
+
+    SimpleExpression<T> getValue() {
+        if (value == null) {
```
beikov, line 123:
```suggestion
        SimpleExpression<T> value = this.value;
        if (value == null) {
```

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/WindowFirstLast.java
beikov:
```suggestion
            this.value = value = Expressions.template(target.getType(), builder.toString(), (List<Expression<?>>) args.build());
```

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/WindowFirstLast.java
```diff
@@ -0,0 +1,146 @@
...
+            if (orderBy.isEmpty()) {
+                // TODO this check should be static
+                throw new IllegalStateException("No order by arguments given");
+            }
+            ImmutableList.Builder<Expression<?>> args = ImmutableList.builder();
+            StringBuilder builder = new StringBuilder();
+            builder.append("{0} keep (dense_rank ");
```
beikov, line 130: What kind of syntax is this? Not sure where this is used, but I don't think we support this.

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/WindowFunction.java
```diff
@@ -0,0 +1,159 @@
...
+     * @since 1.5.0
+     */
+    public WindowFunction(Expression<A> expr, String baseWindowName) {
+        super(expr.getType(), baseWindowName);
+        this.target = expr;
+    }
+
+    @Override
+    public SimpleExpression<A> getValue() {
+        if (value == null) {
```
beikov, line 73:
```suggestion
        SimpleExpression<A> value = this.value;
        if (value == null) {
```

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/WindowFunction.java
```diff
@@ -0,0 +1,159 @@
...
+            value = Expressions.template(target.getType(), "{0} over ({1})", (List<?>) ImmutableList.of(target, super.getValue()));
```
beikov, line 74:
```suggestion
            this.value = value = Expressions.template(target.getType(), "{0} over ({1})", (List<?>) ImmutableList.of(target, super.getValue()));
```

integration/querydsl/pom.xml
```diff
@@ -0,0 +1,25 @@
...
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+    <artifactId>blaze-persistence-integration-querydsl-parent</artifactId>
+    <name>Blazebit Persistence Integration QueryDSL</name>
+    <packaging>pom</packaging>
+    <modules>
+        <module>blaze-persistence-querydsl-expressions</module>
+        <module>blaze-persistence-querydsl-testsuite</module>
+    </modules>
+    <properties>
```
beikov, line 19: These properties here seem unnecessary. Please remove them

integration/querydsl/blaze-persistence-querydsl-expressions/pom.xml
```diff
@@ -0,0 +1,82 @@
...
+        </dependency>
+        <dependency>
+            <groupId>com.blazebit</groupId>
+            <artifactId>blaze-persistence-core-api</artifactId>
+            <version>${project.version}</version>
+            <scope>compile</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.blazebit</groupId>
+            <artifactId>blaze-persistence-core-parser</artifactId>
```
beikov, line 33: You can add core-impl with a runtime scope

integration/querydsl/blaze-persistence-querydsl-expressions/pom.xml
```diff
@@ -0,0 +1,82 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <groupId>com.blazebit</groupId>
+        <artifactId>blaze-persistence-integration-querydsl-parent</artifactId>
+        <version>1.5.0-SNAPSHOT</version>
+        <relativePath>../</relativePath>
```
beikov, line 9:
```suggestion
        <relativePath>../pom.xml</relativePath>
```

integration/querydsl/pom.xml
```diff
@@ -0,0 +1,25 @@
...
+        <querydsl.version>4.2.2</querydsl.version>
```
beikov, line 20: Move that up to the parent pom please

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/package-info.java
```diff
@@ -0,0 +1,206 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * The {@code blaze-persistence-querydsl-expressions} module implements an extended expression model for
```
beikov, line 18:
```suggestion
 * The <pre>blaze-persistence-querydsl-expressions</pre> module implements an extended expression model for
```

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/package-info.java
```diff
@@ -0,0 +1,206 @@
...
+ * Blaze-Persistence JPQL Next.
```
beikov, line 19:
```suggestion
 * Blaze-Persistence JPQL.Next.
```

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/package-info.java
```diff
@@ -0,0 +1,206 @@
...
+ * {@link com.blazebit.persistence.querydsl.BlazeJPAQuery} is analog to {@link com.querydsl.jpa.impl.JPAQuery}. Users can
+ * implement extensions on top of {@link com.blazebit.persistence.querydsl.ExtendedJPAQuery} by extending {@link
+ * com.blazebit.persistence.querydsl.AbstractBlazeJPAQuery}.
+ * </p>
+ *
+ * <p>
+ * {@link com.blazebit.persistence.querydsl.BlazeJPAQuery} can be serialized using the {@link
+ * com.blazebit.persistence.querydsl.JPQLNextSerializer}, and may be rendered to a {@link
+ * com.blazebit.persistence.CriteriaBuilder} using the {@link com.blazebit.persistence.querydsl.BlazeCriteriaBuilderRenderer}.
+ * This allows for the queries to be executed through Blaze-Persistence JPQL Next query engine. Be sure to use the
```
beikov, line 33:
```suggestion
 * This allows for the queries to be executed through Blaze-Persistence JPQL.Next query engine. Be sure to use the
```

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/package-info.java
```diff
@@ -0,0 +1,206 @@
...
+ * {@link com.blazebit.persistence.querydsl.JPQLNextTemplates} or any {@code Templates} implementation that includes the
+ * extensions from {@code JPQLNextTemplates} when using JPQL Next specific features (e.g. window functions, values
```
beikov, line 35:
```suggestion
 * extensions from <pre>JPQLNextTemplates</pre> when using JPQL.Next specific features (e.g. window functions, values
```

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/package-info.java
```diff
@@ -0,0 +1,206 @@
...
+ * clauses, set operations, common table expressions).
+ * </p>
+ *
+ * <p>
+ * This module aims an API that is as close to the original QueryDSL API as possible. Where features did not exist in
+ * {@code querydsl-jpa}, but did exist in {@code querydsl-sql}, we stayed as close to the existing SQL implementation as
```
beikov, line 41:
```suggestion
 * <pre>querydsl-jpa</pre>, but did exist in <pre>querydsl-sql</pre>, we stayed as close to the existing SQL implementation as
```

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/package-info.java
```diff
@@ -0,0 +1,206 @@
...
+ * implement extensions on top of {@link com.blazebit.persistence.querydsl.ExtendedJPAQuery} by extending {@link
+ * com.blazebit.persistence.querydsl.AbstractBlazeJPAQuery}.
+ * </p>
+ *
+ * <p>
+ * {@link com.blazebit.persistence.querydsl.BlazeJPAQuery} can be serialized using the {@link
+ * com.blazebit.persistence.querydsl.JPQLNextSerializer}, and may be rendered to a {@link
+ * com.blazebit.persistence.CriteriaBuilder} using the {@link com.blazebit.persistence.querydsl.BlazeCriteriaBuilderRenderer}.
+ * This allows for the queries to be executed through Blaze-Persistence JPQL Next query engine. Be sure to use the
+ * {@link com.blazebit.persistence.querydsl.JPQLNextTemplates} or any {@code Templates} implementation that includes the
```
beikov, line 34:
```suggestion
 * {@link com.blazebit.persistence.querydsl.JPQLNextTemplates} or any <pre>Templates</pre> implementation that includes the
```

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/package-info.java
```diff
@@ -0,0 +1,206 @@
...
+ * </p>
+ *
+ * <p>
+ * Staying close to QueryDSL's API however, also means that the API is not as fluent as Blaze-Persistence users are
+ * accustomed to. This means that creating common table expressions or complex set operations may lead to superfluous
+ * code.
+ * </p>
+ *
+ * <h2>Examples</h2>
+ * The following chapters demonstrate some of the possibilities of the {@code blaze-persistence-querydsl-expressions}
```
beikov, line 53:
```suggestion
 * The following chapters demonstrate some of the possibilities of the <pre>blaze-persistence-querydsl-expressions</pre>
```

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/package-info.java
```diff
@@ -0,0 +1,206 @@
...
+ *
+ * BlazeJPAQuery&lt;Tuple&gt; query = new BlazeJPAQuery&lt;Tuple&gt;(entityManager, cbf).from(testEntity)
+ *         .select(testEntity.field.as("blep"), testEntity.field.substring(2))
+ *         .where(testEntity.field.length().gt(1));
+ *
+ * List&lt;Tuple&gt; fetch = query.fetch();
+ * </pre>
+ *
+ * <h3>Implicit joins</h3>
+ * <p>Contrary to JPQL, JPQL Next allows for implicit joins.
```
beikov, line 68:
```suggestion
 * <p>Contrary to JPQL, JPQL.Next allows for implicit joins.
```

integration/querydsl/blaze-persistence-querydsl-testsuite/pom.xml
```diff
@@ -0,0 +1,2092 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <groupId>com.blazebit</groupId>
+        <artifactId>blaze-persistence-integration-querydsl-parent</artifactId>
+        <version>1.5.0-SNAPSHOT</version>
+        <relativePath>../</relativePath>
```
beikov, line 9:
```suggestion
        <relativePath>../pom.xml</relativePath>
```

integration/querydsl/blaze-persistence-querydsl-testsuite/src/test/java/com/blazebit/persistence/querydsl/BasicQueryTest.java
```diff
@@ -0,0 +1,744 @@
...
+                    .select(document.name.as("documentName"), rowNumber().over(namedWindow), lastValue(document.name).over(namedWindow))
+                    .where(document.id.in(select(sub.id).from(sub)));
+
+            List<Tuple> fetch = query.fetch();
+            assertFalse(fetch.isEmpty());
+        });
+    }
+
+    @Test
+    @Category({NoMSSQL.class})
```
beikov, line 230: Please add a comment why it doesn't work. I assume your limit without ordering is the issue. You should always define a order by clause when using limit.

integration/querydsl/blaze-persistence-querydsl-testsuite/src/test/java/com/blazebit/persistence/querydsl/BasicQueryTest.java
```diff
@@ -0,0 +1,744 @@
...
+                    .select(document.name.as("documentName"), rowNumber().over().orderBy(document.id), lastValue(document.name).over().partitionBy(document.id).orderBy(document.id))
+                    .where(document.id.in(select(sub.id).from(sub)));
+
+            List<Tuple> fetch = query.fetch();
+            assertFalse(fetch.isEmpty());
+        });
+    }
+
+    @Test
+    @Category({ NoMySQLOld.class })
```
beikov, line 210: Again, add a comment like e.g. `// NOTE: Window functions were only introduced in MySQL8`

integration/querydsl/blaze-persistence-querydsl-testsuite/src/test/java/com/blazebit/persistence/querydsl/BasicQueryTest.java
```diff
@@ -0,0 +1,744 @@
...
+                    .select(document.name.as("blep"), document.name.substring(0, 2))
+                    .where(document.id.in(select(sub.id).from(sub)));
+
+            List<Tuple> fetch = query.fetch();
+            assertFalse(fetch.isEmpty());
+        });
+    }
+
+    @Test
+    @Category({ NoMySQLOld.class })
```
beikov, line 193: Again, add a comment like e.g. `// NOTE: Window functions were only introduced in MySQL8`

integration/querydsl/blaze-persistence-querydsl-testsuite/src/test/java/com/blazebit/persistence/querydsl/BasicQueryTest.java
```diff
@@ -0,0 +1,744 @@
...
+                    .innerJoin(otherBook).on(otherBook.owner.eq(otherAuthor))
+                    .transform(GroupBy.groupBy(otherAuthor).as(GroupBy.list(otherBook)));
+
+            assertNotNull(booksByAuthor);
+        });
+    }
+
+
+    @Test
+    @Category({ NoDatanucleus.class, NoEclipselink.class, NoOpenJPA.class })
```
beikov, line 289: Add a comment like e.g. `// NOTE: No advanced sql support for datanucleus, eclipselink and openjpa yet`

integration/querydsl/blaze-persistence-querydsl-testsuite/src/test/java/com/blazebit/persistence/querydsl/BasicQueryTest.java
```diff
@@ -0,0 +1,744 @@
...
+                    .fromValues(document, Collections.singleton(theBook))
+                    .select(document)
+                    .fetch();
+
+            assertNotNull(fetch);
+        });
+    }
+
+    @Test
+    @Category({ NoDatanucleus.class, NoEclipselink.class, NoOpenJPA.class })
```
beikov, line 306: Add a comment like e.g. `// NOTE: No advanced sql support for datanucleus, eclipselink and openjpa yet`

integration/querydsl/blaze-persistence-querydsl-testsuite/src/test/java/com/blazebit/persistence/querydsl/BasicQueryTest.java
```diff
@@ -0,0 +1,744 @@
...
+                    .fromValues(document.name, bookName, Collections.singleton("book"))
+                    .select(bookName)
+                    .fetch();
+
+            assertNotNull(fetch);
+        });
+    }
+
+    @Test
+    @Category({ NoMySQL.class, NoFirebird.class, NoDatanucleus.class, NoEclipselink.class, NoOpenJPA.class })
```
beikov, line 321: Add a comment like e.g. `// NOTE: No advanced sql support for datanucleus, eclipselink and openjpa yet` and also `// NOTE: MySQL only supports the UNION set operation`

integration/querydsl/blaze-persistence-querydsl-testsuite/src/test/java/com/blazebit/persistence/querydsl/BasicQueryTest.java
```diff
@@ -0,0 +1,744 @@
...
+                                        select(document).from(document).where(document.id.eq(43L)))),
+                        select(document).from(document).where(document.id.eq(46L)))
+                    .fetch();
+
+            assertNotNull(fetch);
+        });
+    }
+
+    @Test
+    @Category({ NoMySQL.class, NoFirebird.class, NoDatanucleus.class, NoEclipselink.class, NoOpenJPA.class })
```
beikov, line 357: Add a comment like e.g. `// NOTE: No advanced sql support for datanucleus, eclipselink and openjpa yet` and also `// NOTE: MySQL only supports the UNION set operation`

integration/querydsl/blaze-persistence-querydsl-testsuite/src/test/java/com/blazebit/persistence/querydsl/BasicQueryTest.java
```diff
@@ -0,0 +1,744 @@
...
+                    .select(book2)
+                    .from(book2).where(book2.id.in(union))
+                    .fetch();
+
+            assertNotNull(fetch);
+        });
+    }
+
+    @Test
+    @Category({ NoDatanucleus.class, NoEclipselink.class, NoOpenJPA.class, NoMySQLOld.class })
```
beikov, line 400: Add a comment like e.g. `// NOTE: No advanced sql support for datanucleus, eclipselink and openjpa yet`
Old MySQL version should work fine because we normally inline the CTE. Any reason to disable this here?

integration/querydsl/blaze-persistence-querydsl-testsuite/src/test/java/com/blazebit/persistence/querydsl/BasicQueryTest.java
```diff
@@ -0,0 +1,744 @@
...
+                    .with(idHolderCTE, idHolderCTE.id).as(select(document.id).from(document))
+                    .select(idHolderCTE.id).from(idHolderCTE)
+                    .fetch();
+
+            assertNotNull(fetch);
+        });
+    }
+
+    @Test
+    @Category({ NoDatanucleus.class, NoEclipselink.class, NoOpenJPA.class, NoMySQLOld.class })
```
beikov, line 413: Add a comment like e.g. `// NOTE: No advanced sql support for datanucleus, eclipselink and openjpa yet`
Old MySQL version should work fine because we normally inline the CTE. Any reason to disable this here?

integration/querydsl/blaze-persistence-querydsl-testsuite/src/test/java/com/blazebit/persistence/querydsl/BasicQueryTest.java
```diff
@@ -0,0 +1,744 @@
...
+                    .with(idHolderCTE, select(new Binds<IdHolderCTE>().bind(idHolderCTE.id, document.id)).from(document))
+                    .select(idHolderCTE.id).from(idHolderCTE)
+                    .fetch();
+
+            assertNotNull(fetch);
+        });
+    }
+
+    @Test
+    @Category({ NoDatanucleus.class, NoEclipselink.class, NoOpenJPA.class, NoMySQLOld.class })
```
beikov, line 426: Add a comment like e.g. `// NOTE: No advanced sql support for datanucleus, eclipselink and openjpa yet`
Old MySQL version should work fine because we normally inline the CTE. Any reason to disable this here?

integration/querydsl/blaze-persistence-querydsl-testsuite/src/test/java/com/blazebit/persistence/querydsl/BasicQueryTest.java
```diff
@@ -0,0 +1,744 @@
...
+                        bind(idHolderCTE.id, document.id)).from(document))
+                .select(idHolderCTE.id).from(idHolderCTE)
+                .fetch();
+
+                assertNotNull(fetch);
+        });
+    }
+
+    @Test
+    @Category({ NoDatanucleus.class, NoEclipselink.class, NoOpenJPA.class, NoMySQL.class })
```
beikov, line 440: Add a comment like e.g. `// NOTE: No advanced sql support for datanucleus, eclipselink and openjpa yet`
Old MySQL version should work fine because we normally inline the CTE. Any reason to disable this here?

integration/querydsl/blaze-persistence-querydsl-testsuite/src/test/java/com/blazebit/persistence/querydsl/BasicQueryTest.java
```diff
@@ -0,0 +1,744 @@
...
+                    .fetch();
+
+            assertNotNull(fetch);
+        });
+    }
+
+
+
+    @Test
+    @Category({ NoDatanucleus.class, NoEclipselink.class, NoOpenJPA.class, NoMySQLOld.class })
```
beikov, line 455: Add a comment like e.g. `// NOTE: No advanced sql support for datanucleus, eclipselink and openjpa yet`
Old MySQL version should work fine because we normally inline the CTE. Any reason to disable this here?

integration/querydsl/blaze-persistence-querydsl-testsuite/src/test/java/com/blazebit/persistence/querydsl/BasicQueryTest.java
```diff
@@ -0,0 +1,744 @@
...
+                            .select(document.id))
+                    .select(idHolderCTE.id).from(idHolderCTE)
+                    .fetch();
+
+            assertNotNull(fetch);
+        });
+    }
+
+    @Test
+    @Category({ NoDatanucleus.class, NoEclipselink.class, NoOpenJPA.class, NoMySQLOld.class, NoDB2.class })
```
beikov, line 474: Add a comment like e.g. `// NOTE: No advanced sql support for datanucleus, eclipselink and openjpa yet` and also `// NOTE: Recursive CTE support was only added in MySQL8`.
Apart from that, why do you exclude DB2 here?

integration/querydsl/blaze-persistence-querydsl-testsuite/src/test/java/com/blazebit/persistence/querydsl/BasicQueryTest.java
```diff
@@ -0,0 +1,744 @@
...
+                            .join(idHolderCTE).on(idHolderCTE.id.add(1L).eq(document.id))))
+                    .select(idHolderCTE.id).from(idHolderCTE)
+                    .fetch();
+
+            assertNotNull(fetch);
+        });
+    }
+
+    @Test
+    @Category({ NoDatanucleus.class, NoEclipselink.class, NoOpenJPA.class, NoMySQLOld.class })
```
beikov, line 490: Add a comment like e.g. `// NOTE: No advanced sql support for datanucleus, eclipselink and openjpa yet` and also `// NOTE: Recursive CTE support was only added in MySQL8`.

integration/querydsl/blaze-persistence-querydsl-testsuite/src/test/java/com/blazebit/persistence/querydsl/BasicQueryTest.java
```diff
@@ -0,0 +1,744 @@
...
+                                select(document.id).from(document).join(idHolderCTE).on(idHolderCTE.id.add(1L).eq(document.id))))
+                    .select(idHolderCTE.id).from(idHolderCTE)
+                    .fetch();
+
+            assertNotNull(fetch);
+        });
+    }
+
+    @Test
+    @Category({ NoDatanucleus.class, NoEclipselink.class, NoOpenJPA.class })
```
beikov, line 508: Add a comment like e.g. `// NOTE: No advanced sql support for datanucleus, eclipselink and openjpa yet`

integration/querydsl/blaze-persistence-querydsl-testsuite/src/test/java/com/blazebit/persistence/querydsl/BasicQueryTest.java
```diff
@@ -0,0 +1,744 @@
...
+                            .orderBy(recursiveEntity.name.asc())
+                            .limit(1L), recursiveEntity)
+                    .fetch();
+
+            assertNotNull(fetch);
+        });
+    }
+
+    @Test
+    @Category({ NoDatanucleus.class, NoEclipselink.class, NoOpenJPA.class, NoHibernate42.class, NoHibernate43.class, NoHibernate50.class, NoHibernate51.class})
```
beikov, line 527: Add a comment like e.g. `// NOTE: No advanced sql support for datanucleus, eclipselink and openjpa yet`

Apart from that, why do you exclude old Hibernate versions? Inner entity joins are emulated in older versions.

integration/querydsl/blaze-persistence-querydsl-testsuite/src/test/java/com/blazebit/persistence/querydsl/BasicQueryTest.java
```diff
@@ -0,0 +1,744 @@
...
+                    .on(t.eq(subT2))
+                    .fetch();
+
+            assertNotNull(fetch);
+
+        });
+    }
+
+    @Test
+    @Category({ NoDatanucleus.class, NoEclipselink.class, NoOracle.class, NoOpenJPA.class, NoHibernate42.class, NoHibernate43.class, NoHibernate50.class, NoHibernate51.class, NoH2.class, NoMySQLOld.class })
```
beikov, line 575: Please add comments for the exclusions. The Oracle and old MySQL version we test with doesn't support lateral, H2 also doesn't.

integration/querydsl/blaze-persistence-querydsl-testsuite/src/test/java/com/blazebit/persistence/querydsl/BasicQueryTest.java
```diff
@@ -0,0 +1,744 @@
...
+                            .on(t.eq(subT))
+                            .where(t.parent.name.eq("root1"))
+                            .orderBy(t.name.asc())
+                            .limit(1L), t)
+                    .leftJoin(selectFrom(subT2)
+                        .where(subT2.parent.name.eq("root1"))
+                        .orderBy(subT2.name.asc())
+                        .limit(1), subT3)
+                        .on(t.eq(subT3))
+                        .lateral()
```
beikov, line 598: Just thinking out loud, wouldn't it be better to have methods like `leftJoinLateral` etc. since lateral can only be left or inner?

integration/querydsl/blaze-persistence-querydsl-testsuite/src/test/java/com/blazebit/persistence/querydsl/BasicQueryTest.java
```diff
@@ -0,0 +1,744 @@
...
+                    .fetch();
+
+        });
+    }
+
+    // NOTE: Entity joins are only supported on Hibernate 5.1+
+    @Test
+    @Category({ NoDatanucleus.class, NoEclipselink.class, NoOpenJPA.class, NoHibernate42.class, NoHibernate43.class, NoHibernate50.class, NoH2.class, NoMySQLOld.class, NoOracle.class })
```
beikov, line 606: Please add comments for the exclusions. The Oracle and old MySQL version we test with doesn't support lateral, H2 also doesn't.

integration/querydsl/blaze-persistence-querydsl-testsuite/src/test/java/com/blazebit/persistence/querydsl/BasicQueryTest.java
```diff
@@ -0,0 +1,744 @@
...
+                    .lateral()
+                    .fetch();
+
+            assertNotNull(fetch);
+        });
+    }
+
+
+    @Test
+    @Category({ NoDatanucleus.class, NoEclipselink.class, NoOpenJPA.class, NoHibernate42.class, NoHibernate43.class, NoHibernate50.class, NoHibernate51.class, NoH2.class, NoMySQLOld.class, NoOracle.class })
```
beikov, line 628: Please add comments for the exclusions. The Oracle and old MySQL version we test with doesn't support lateral, H2 also doesn't.

integration/querydsl/blaze-persistence-querydsl-testsuite/src/test/java/com/blazebit/persistence/querydsl/BasicQueryTest.java
```diff
@@ -0,0 +1,744 @@
...
+                    .lateral()
+                    .fetch();
+
+            assertNotNull(fetch);
+        });
+    }
+
+    @Test
+
+    @Category({ NoDatanucleus.class, NoEclipselink.class, NoOpenJPA.class, NoHibernate42.class, NoHibernate43.class, NoHibernate50.class, NoHibernate51.class, NoMySQL.class })
```
beikov, line 653: Add a comment like e.g. `// NOTE: No advanced sql support for datanucleus, eclipselink and openjpa yet` and also `// NOTE: Entity joins are only supported on Hibernate 5.1+`

integration/querydsl/blaze-persistence-querydsl-testsuite/src/test/java/com/blazebit/persistence/querydsl/BasicQueryTest.java
```diff
@@ -0,0 +1,744 @@
...
+                    .select(t, subT2)
+                    .from(t)
+                    .leftJoin(union, subT2).on(subT2.eq(subT2))
+                    .fetch();
+        });
+    }
+
+    @Test
+
+    @Category({ NoDatanucleus.class, NoEclipselink.class, NoOpenJPA.class, NoHibernate42.class, NoHibernate43.class, NoHibernate50.class, NoHibernate51.class, NoMySQL.class })
```
beikov, line 689: Add a comment like e.g. `// NOTE: No advanced sql support for datanucleus, eclipselink and openjpa yet` and also `// NOTE: Entity joins are only supported on Hibernate 5.1+`

Also, why is MySQL excluded here?

integration/querydsl/blaze-persistence-querydsl-testsuite/src/test/java/com/blazebit/persistence/querydsl/BasicQueryTest.java
```diff
@@ -0,0 +1,744 @@
...
+            new BlazeJPAQuery<RecursiveEntity>(entityManager, cbf)
+                    .select(t, subT2)
+                    .from(t)
+                    .leftJoin(union, subT2).on(subT2.eq(subT2))
+                    .fetch();
+        });
+    }
+
+    @Test
+    @Category({ NoDatanucleus.class, NoEclipselink.class, NoOpenJPA.class, NoHibernate42.class, NoHibernate43.class, NoHibernate50.class, NoHibernate51.class, NoH2.class, NoMySQLOld.class })
```
beikov, line 714: Add a comment like e.g. `// NOTE: No advanced sql support for datanucleus, eclipselink and openjpa yet` and also `// NOTE: Entity joins are only supported on Hibernate 5.1+`
Also add comments for the other exclusions. The Oracle and old MySQL version we test with doesn't support lateral, H2 also doesn't.

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/BlazeCriteriaBuilderRenderer.java
```diff
@@ -0,0 +1,1308 @@
...
+            }
+        }
+    }
+
+    private void renderConstants(ParameterHolder<?> criteriaBuilder) {
+        for (Map.Entry<Object, String> entry : constantToLabel.entrySet()) {
+            try {
+                criteriaBuilder.setParameter(entry.getValue(), entry.getKey());
+            } catch (Exception e) {
+                throw e;
```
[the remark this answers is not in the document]
jwgmeligmeyling: Debugging left over. It actually did compile.

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/BlazeCriteriaBuilderRenderer.java
```diff
@@ -0,0 +1,1308 @@
...
+                }
+                boolean entityJoin = entityPath.getMetadata().isRoot();
+
+                switch (joinExpression.getType()) {
+                    case DEFAULT:
+                        if (fromBuilder instanceof FromBuilder) {
+                            criteriaBuilder = (X) fromBuilder;
+                            From from = criteriaBuilder.getFrom(alias);
+
+                            // TODO find a clean way to detect pre-set FROM clauses...
```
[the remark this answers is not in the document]
jwgmeligmeyling: The issue we talked about, that I currently need to instantiate a CBF with return type `Object.class` because I need to reliably remove/detect the default join target.

integration/querydsl/blaze-persistence-querydsl-testsuite/src/test/java/com/blazebit/persistence/querydsl/BasicQueryTest.java
jwgmeligmeyling: I copied the test profiles from the spring test suite, which still uses the old profiles for MySQL, so my tests with NoMySQLOld were not properly excluded. I hope this is a left over from my attempts figuring that one out :)

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/BlazeCriteriaBuilderRenderer.java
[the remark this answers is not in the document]
beikov: Hmm, if you always invoke a from method, there will be no default from element, so that might be good enough to clear this one.

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/AbstractBlazeJPAQuery.java
```diff
@@ -0,0 +1,527 @@
...
+            return (Long) query.getSingleResult();
+        } finally {
+            reset();
+        }
+    }
+
+    @Override
+    public QueryResults<T> fetchResults() {
+        try {
+            Query countQuery = createQuery(null, true);
```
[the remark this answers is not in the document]
jwgmeligmeyling: `createQuery`(..., true : forCount)`  returns the `TypedQuery` from `getFullQueryBuilder(...).getCountQuery()` under the hood.

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/BlazeCriteriaBuilderRenderer.java
```diff
@@ -0,0 +1,1308 @@
...
+                                Path<?> alias = null;
+                                Path<?> cteEntityPath = null;
+
+                                if (projExpression instanceof Operation) {
+                                    Operation<?> projOperation = (Operation<?>) projExpression;
+                                    if (projOperation.getOperator() == BIND) {
+                                        alias = (Path<?>) projOperation.getArg(1);
+                                        cteEntityPath = alias.getRoot();
+                                        projExpression = projOperation.getArg(0);
+                                    }
```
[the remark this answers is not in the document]
jwgmeligmeyling: The `BIND` operation must be the outermost operation, similar to the `ALIAS` operation. So if we have unwrapped a possible ALIAS or BIND operation, and are still left with an operation, then this is just an operation that should be added to the projection. We don't reassign the expression to be projected and just follow the normal flow.

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/BlazeCriteriaBuilderRenderer.java
```diff
@@ -0,0 +1,1308 @@
...
+                                        public Object apply(String expression) {
+                                            return bindBuilder.select(expression);
+                                        }
+                                    }, new BlazeCriteriaBuilderRenderer.Function<String, MultipleSubqueryInitiator<?>>() {
+                                        @Override
+                                        public MultipleSubqueryInitiator<?> apply(String expression) {
+                                            return bindBuilder.selectSubqueries(expression);
+                                        }
+                                    });
+                                }
```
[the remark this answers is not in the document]
jwgmeligmeyling: I'll add an exception. This would be a select statement which is not bound to any CTE attribute.

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/BlazeCriteriaBuilderRenderer.java
```diff
@@ -0,0 +1,1308 @@
...
+
+                                if (alias != null) {
+                                    String aliasString = relativePathString(cteEntityPath, alias);
+                                    final SelectBuilder<?> bindBuilder = selectBaseCriteriaBuilder.bind(aliasString);
+                                    setExpressionSubqueries(projExpression, new BlazeCriteriaBuilderRenderer.Function<String, Object>() {
+                                        @Override
+                                        public Object apply(String expression) {
+                                            return bindBuilder.select(expression);
+                                        }
+                                    }, new BlazeCriteriaBuilderRenderer.Function<String, MultipleSubqueryInitiator<?>>() {
```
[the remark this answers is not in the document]
jwgmeligmeyling: Originally I used method references here, so the code used to be `setExpressionSubqueries(beanPath, bindBuilder::select, bindBuilder::selectSubqueries)`.

I didn't bother to create inner classes because most of them are only used once (i.e. I have these anonymous classes as well for JoinOnBuilder, WhereBuilder, Selectbuilder, HavingBuilder). This one is indeed used twice, so I can extract them. However, after the Java 8 migration these cases that can be optimized using method references will be harder to detect again. These simple anonymous classes will be detected with ease by the IDE, and I think are better than having the inner classes.

WDYT?

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/BlazeCriteriaBuilderRenderer.java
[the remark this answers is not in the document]
jwgmeligmeyling: Actually, this is to detect the from clause set by `fromEntitySubquery`

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/BlazeCriteriaBuilderRenderer.java
```diff
@@ -0,0 +1,1308 @@
...
+
+                        SubQueryExpression<?> subQueryExpression = target.accept(new FirstSubqueryResolver(), null);
+                        Path<?> fromPath = subQueryExpression.getMetadata().getJoins().get(0).getTarget().accept(new FirstSubqueryTargetPathResolver(), null);
+                        boolean entityJoin = fromPath.getMetadata().isRoot();
+
+                        if (hasCondition) {
+                            FullSelectCTECriteriaBuilder<JoinOnBuilder<X>> joinOnBuilderFullSelectCTECriteriaBuilder;
+
+                            if (isLateral) {
+                                // TODO: Visit for SET operations
```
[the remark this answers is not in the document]
jwgmeligmeyling: I mean UNION, INTERSECT, etc. here which used to need special handling here. But that is since refactored, I removed the TODO.

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/BlazeCriteriaBuilderRenderer.java
```diff
@@ -0,0 +1,1308 @@
...
+        }
+
+    }
+
+    /**
+     * Visitor for rendering window expressions.
+     *
+     * @param <X> Concrete window builder
+     */
+    private class WindowContainerBuilderDefaultVisitorImpl<X extends WindowContainerBuilder<X>> extends DefaultVisitorImpl<X, Object> {
```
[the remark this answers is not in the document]
jwgmeligmeyling: No, it needs the context of the query rendering for registering constants and parameters.

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/BlazeCriteriaBuilderRenderer.java
```diff
@@ -0,0 +1,1308 @@
...
+                },
+                new BlazeCriteriaBuilderRenderer.Function<String, MultipleSubqueryInitiator<?>>() {
+                    @Override
+                    public MultipleSubqueryInitiator<?> apply(String expression) {
+                        return finalAlias != null ? selectBuilder.selectSubqueries(expression, finalAlias) : selectBuilder.selectSubqueries(expression);
+                    }
+                });
+    }
+
+    private String renderExpression(Expression<?> select) {
```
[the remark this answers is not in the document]
jwgmeligmeyling: The `JPQLNextSerializer` is able to serialize to Strings, i.e.

```java
JPQLNextSerializer jpqlNextSerializer = new JPQLNextSerializer();
expression.accept(jpqlNextSerializer, null);
String output = jpqlNextSerializer.toString();
```

Unfortunately, the `JPQLSerializer` that it's based on treats outer queries slightly different. It probably has to. Outer queries, orthogonally to our FullQueryBuilder, just support features plain subqueries don't.

I didn't bother too much with it, because there is no full query parser for JPQL.Next, so I have to render against the CriteriaBuilder API anyway. The desired way to "serialize" a query, and the one I use in `AbstractBlazeJPAQuery`, is IMO to just render it and invoke `getQueryString()` on the `Queryable`.

Another complication here is that the `BlazeCriteriaBuilderRenderer` has to hook directly on the `JPQLSerializer`. For example, it renders subqueries as an alias, which is not really practical for to-string serialization.

integration/querydsl/blaze-persistence-querydsl-expressions/pom.xml
jwgmeligmeyling: I need the parser in compile scope for `TypeUtils` (literals) and `EntityMetamodel` (from entity-binds)

integration/querydsl/blaze-persistence-querydsl-testsuite/src/test/java/com/blazebit/persistence/querydsl/BasicQueryTest.java
jwgmeligmeyling: I'll run it again

integration/querydsl/blaze-persistence-querydsl-testsuite/src/test/java/com/blazebit/persistence/querydsl/BasicQueryTest.java
jwgmeligmeyling: Perhaps its not supported because I used join/on, instead of from/where. Including the test for DB2 again to see if it still fails with the change.

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/WindowFirstLast.java
jwgmeligmeyling: First/last is something I ported over from the Window function implementation from QueryDSL by accident. I have now removed it, we do not support it.

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/BlazeCriteriaBuilderRenderer.java
```diff
@@ -0,0 +1,1308 @@
...
+        this.subQueryToLabel = new IdentityHashMap<>();
+        return subQueryToLabel;
+    }
+
+    /**
+     * Drop in for java.util.Function
+     * @param <T> argument type
+     * @param <R> return type
+     */
+    private interface Function<T, R> {
```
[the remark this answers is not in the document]
jwgmeligmeyling: See comment about https://github.com/Blazebit/blaze-persistence/pull/1071#discussion_r422485918

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/JPQLNextExpressions.java
```diff
@@ -0,0 +1,1145 @@
...
+     * Create a dateadd(unit, date, amount) expression
+     *
+     * @param unit date part
+     * @param date date
+     * @param amount amount
+     * @param <D> date type
+     * @return converted date
+     * @see com.querydsl.sql.SQLExpressions#dateadd(com.querydsl.sql.DatePart, DateExpression, int)
+     */
+    public static <D extends Comparable<?>> DateExpression<D> dateadd(DatePart unit, DateExpression<D> date, int amount) {
```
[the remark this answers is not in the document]
jwgmeligmeyling: Variants added.

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/JPQLNextSerializer.java
```diff
@@ -0,0 +1,129 @@
...
+    public String takeBuffer() {
+        String res = builder.toString();
+        clearBuffer();
+        return res;
+    }
+
+    private static StringBuilder getStringBuilder(JPQLNextSerializer serializer) {
+        try {
+            // Unfortunately, builder is private...
+            Field builderField = SerializerBase.class.getDeclaredField("builder");
```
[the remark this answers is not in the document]
jwgmeligmeyling: Fixed.

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/WindowFirstLast.java
jwgmeligmeyling: Window First/Last is removed. It was never supported.

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/WindowFirstLast.java
jwgmeligmeyling: Window First/Last is removed. It was never supported.

jwgmeligmeyling:
Removed all `@since` annotations on the methods.

integration/querydsl/blaze-persistence-querydsl-testsuite/src/test/java/com/blazebit/persistence/querydsl/BasicQueryTest.java
jwgmeligmeyling: I've used `fetchJoin()` and `fetchAll()` as comparison material here. There methods suffer from the same issue (as these are only applicable to association joins and not entity joins). Its not the most ideal API but I also do value some consistency here. Not sure though how this way of dealing with `JoinFlags` was ever well thought through. There aren't so many other types of JoinFlags around in QueryDsl to compare with.

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/AbstractBlazeJPAQuery.java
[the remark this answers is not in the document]
beikov: `PaginatedCriteriaBuilder.getTypedQuery()` returns a `PaginatedTypedQuery` which then returns the `PagedList` from `getResultList()` which is much more powerful than invoking a separate count query like you do here.

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/BlazeCriteriaBuilderRenderer.java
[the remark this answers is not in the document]
beikov: Not sure I understand, I didn't dig that much into it yet. Just wanted to make sure you give this another thought but I am sure you know what you are doing ;)

integration/querydsl/blaze-persistence-querydsl-expressions/pom.xml
beikov: That's fine, but you can still add core-impl with runtime scope

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/WindowFirstLast.java
beikov: Are we talking about `first_value` and `last_value` window functions? We support those.

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/BlazeCriteriaBuilderRenderer.java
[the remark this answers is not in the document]
beikov: The issue with method references is the same since you have a capture variable. Using BiFunction will eliminate all allocations.

documentation/src/main/asciidoc/core/manual/en_US/querydsl.adoc
```diff
@@ -75,18 +75,15 @@ Assume we have the following entity:
 
 [source,java]
 ----
-@EntityView(Cat.class)
-public interface SimpleCatView {
-
-    @IdMapping
-    public getId();
+@Entity
+public interface Cat {
```
beikov, line 79:
```suggestion
public class Cat {
```

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/AbstractBlazeJPAQuery.java
```diff
@@ -253,9 +239,11 @@ public String getQueryString() {
     }
 
     @Override
+    @SuppressWarnings({ "unchecked", "unsafe" })
     protected void clone(Q query) {
         super.clone(query);
         this.cachable = query.cachable;
+        this.binds.addBinds((List) query.binds.getArgs());
```
beikov, line 246: Are `Operation` objects immutable? Otherwise you'd have to copy them as well

integration/querydsl/pom.xml
```diff
@@ -17,9 +17,6 @@
         <module>blaze-persistence-querydsl-testsuite</module>
     </modules>
     <properties>
-        <querydsl.version>4.2.2</querydsl.version>
-        <hibernate.version>5.4.10.Final</hibernate.version>
-        <blaze-persistence.version>${project.version}</blaze-persistence.version>
         <module.name>com.blazebit.persistence.querydsl</module.name>
```
beikov, line 20: This is a parent pom, that doesn't need a module name.

integration/querydsl/blaze-persistence-querydsl-testsuite/src/test/java/com/blazebit/persistence/querydsl/BasicQueryTest.java
beikov: I understand that you want to stay as close to the QueryDSL model as possible.

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/BlazeCriteriaBuilderRenderer.java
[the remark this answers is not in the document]
beikov: I think you can make these implementations stateless singletons if you use the `BiFunction` approach which is less memory intensive.

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/BlazeCriteriaBuilderRenderer.java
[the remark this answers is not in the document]
beikov: AFAICS you only need the `renderExpression` method which I suggested could be _moved_ to the serializer class. Then you could pass just the serializer unless I didn't miss something.

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/BlazeCriteriaBuilderRenderer.java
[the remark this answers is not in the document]
beikov: So is the JPQLNextSerializer just for the _human readable_ string representation?
Anyway, I was thinking that moving this method to the serializer class would allow making `WindowContainerBuilderDefaultVisitorImpl` static and let it accept just the serializer, but I probably missed some context usage.

integration/querydsl/blaze-persistence-querydsl-testsuite/src/test/java/com/blazebit/persistence/querydsl/BasicQueryTest.java
beikov: I know DB2 has an issue with joins in recursive queries, but please add a note if that is the issue here.

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/BlazeCriteriaBuilderRenderer.java
```diff
@@ -0,0 +1,1308 @@
...
+         * Drop in for java.util.Function, TODO: fix after JDK 8
+         *
+         * @param t argument
+         * @return return value
+         */
+        R apply(T t);
+
+    }
+
+    private <X> X setExpressionSubqueries(Expression<?> expression, Function<String, ? extends X> setExpression, Function<String, MultipleSubqueryInitiator<? extends X>> setExpressionSubqueries) {
```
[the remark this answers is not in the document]
jwgmeligmeyling: I'm not entirely seeing this yet. The original intent behind it was the one I described in https://github.com/Blazebit/blaze-persistence/pull/1071#discussion_r422485918

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/AbstractBlazeJPAQuery.java
jwgmeligmeyling: Usually they should be, but there are some mutable expression types that theoretically could be referenced transitively. To be sure, I am pulling it through a clone visitor now.

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/AbstractBlazeJPAQuery.java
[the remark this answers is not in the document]
jwgmeligmeyling: Ah I only now see what this method is doing..

Usually one uses `fetch()` wich is equivalent to our `getResultList()`. There is also `fetchOne()` (`getSingleResult`) and `fetchFirst()`.

Because implementing `Queryable` would only throw more confusing methods in the game, I did not do that. But I did port some of the pagination API's over to `BlazeJPAQuery`.

The `fetchResults` method seems to be some sort of mechanism to perform the count query first (without any limit or offset), and only perform the query with projection if there is any result (and use limit and offset in that case).

I could add something like:

```
QueryModifiers modifiers = getMetadata().getModifiers();
Long offset = modifiers.getOffset();
Long limit = modifiers.getLimit();

if (offset != null && limit != null) {
    Queryable<T, ?> queryable = getQueryable(null);
    if (queryable instanceof FullQueryBuilder) {
        FullQueryBuilder<T, ?> fullQueryBuilder = (FullQueryBuilder<T, ?>) queryable;
        PaginatedCriteriaBuilder<T> page = fullQueryBuilder.page(offset.intValue(), limit.intValue());
        PagedList<T> resultList = page.getResultList();
        return new QueryResults<>(resultList, modifiers, resultList.getTotalSize());
    }
} else {
    // Perform the old code.
}
```

But this wouldn't prevent users from using this in cases where pagination is not allowed, for example when no viable sorting is used).

In reality, I think `fetchResults`  should be discouraged in favour for the actual pagination API methods.

I however had to override it, because it calls `createQuery` internally, which is a private method. So rather than overriding `createQuery`, I had to "copy" all methods that use it over so that they call my extended version of `createQuery` that is actually able to render JPQL.Next queries... My PR to open `createQuery` up for extension actually got merged ( https://github.com/querydsl/querydsl/pull/2517 ) but we can't rely on this change being available or we would have to force our users to be on the absolute newest version of Querydsl.

Perhaps I should just leave the code as is, but throw a warning on the logger indicating that the pagination methods are way more performant for this type of use case?

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/BlazeCriteriaBuilderRenderer.java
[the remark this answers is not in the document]
jwgmeligmeyling: We probably could use a plain `JPQLNextSerializer` to serialize the expressions used in the window clause, because I _think_ subqueries are not allowed there. I am not even sure of that. If they are allowed, it gets a lot harder.

I don't think that we can easily move the `BlazeCriteriaBuilderRenderer#renderExpression` to `JPQLNextSerializer`. We need special handling for some expression types. Nested CTE's, SubqueryInitiators, QueryFlags, window functions, all get in the way here. For example, the `JPQLNextSerializer` just writes the entire window to an expression string, but the renderer has to use a visitor to it can call the right builder methods.

I can add a whole lot of context to the JPQLNextSerializer that can be pushed and popped, cte's, subqueries, windows, CTE binds, so that the `BlazeCriteriaBuilderRenderer` can retrieve it from there, but I think that would just eventually mean that we integrate `BlazeCriteriaBuilderRenderer` with `JPQLNextSerializer` into a single class. This is actually something I had actually done before but I really wanted to pull out again because the `BlazeCriteriaBuilderRenderer`  can't satisfy the regular `JPQLSerializer` interface and would publically expose a lot of unwanted methods.

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/AbstractBlazeJPAQuery.java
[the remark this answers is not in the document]
beikov: I'm not sure I understand what you mean here. The `PaginatedCriteriaBuilder` API does the same i.e. first execute a count query and then the main query with limit/offset if the count is > 0.
The `PaginatedCriteriaBuilder` API additionally allows to inline the count query into the id or main query to avoid multiple queries, which is the default, but can be turned off.

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/BlazeCriteriaBuilderRenderer.java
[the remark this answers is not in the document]
beikov: Ok fine, it was just an idea.

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/BlazeCriteriaBuilderRenderer.java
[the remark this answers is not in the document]
beikov: I understand your desire for using method references, but the fact is that this is memory pressure. You capture context so the JVM has to create objects for every usage. If we use the BiFunction approach you can use singletons here and have no allocations here.

integration/querydsl/blaze-persistence-querydsl-expressions/src/main/java/com/blazebit/persistence/querydsl/BlazeCriteriaBuilderRenderer.java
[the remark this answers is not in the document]
jwgmeligmeyling: The JPQLNextSerializer just makes the required changes to the JPQLSerializer to deal with set operations and work around some issues in QueryDSL (for example how they use HQL casts).

It generally produces executable queries, but there are some quirks when it gets to CTEs, because the bindings are not "expanded" until the query is actually rendered in the renderer.

Furthermore it may produce some JPQLNext specific dialect such as window clauses for which no full query parser yet exists.

All that magic is handled in the renderer instead.

jwgmeligmeyling:
Just two more failures among the profiles. One seems related to a flaky Quarkus test. The other one is still caused by https://bugs.openjdk.java.net/browse/JDK-8212233 under one of the JDK 11 profiles. Oddly enough the same issue for all other JDK9+ profiles is gone. Hopefully the patch below will fix that one:

```diff
diff --git a/integration/querydsl/blaze-persistence-querydsl-expressions/pom.xml b/integration/querydsl/blaze-persistence-querydsl-expressions/pom.xml
index e859a612a..e25ec5711 100644
--- a/integration/querydsl/blaze-persistence-querydsl-expressions/pom.xml
+++ b/integration/querydsl/blaze-persistence-querydsl-expressions/pom.xml
@@ -58,7 +58,8 @@
                 <artifactId>maven-javadoc-plugin</artifactId>
                 <version>3.2.0</version>
                 <configuration>
-                    <source>7</source> <!-- https://bugs.openjdk.java.net/browse/JDK-8212233 -->
+                    <source>8</source> <!-- https://bugs.openjdk.java.net/browse/JDK-8212233 -->
+                    <release>8</release> <!-- https://bugs.openjdk.java.net/browse/JDK-8212233 -->
                     <dependencySourceIncludes>com.querydsl:*</dependencySourceIncludes>
                     <links>
                         <link>https://javadoc.io/doc/com.querydsl/querydsl-jpa/${querydsl.version}</link>
```

So I'll commit that and test the failing profile again. If not, I just remove the custom Javadoc generation at all. Then we lose the links to QueryDSL's Javadoc but I'm done with it 😉 I also cant reproduce it locally. Perhaps my patch version for JDK11 is too high and already has the issue resolved...

PR #1098, opened 2020-05-18
publish quarkus test jars on release

Mobe91:


examples/quarkus/testsuite/native/pom.xml
```diff
@@ -38,6 +38,15 @@
                 <module>mssql</module>
             </modules>
         </profile>
+        <profile>
+            <id>release</id>
```
beikov, line 42: The profile is called `blazebit-release`
```suggestion
            <id>blazebit-release</id>
```

Mobe91:
I tried a clean install after adding the test-jar goals.

examples/quarkus/testsuite/pom.xml
```diff
@@ -21,5 +21,11 @@
                 <module>native</module>
             </modules>
         </profile>
+        <profile>
+            <id>blazebit-release</id>
+            <modules>
+                <module>native</module>
```
beikov, line 27: You forgot to also add the module `base`. Specifying modules here overrides the modules AFAIK.

PR #1113, opened 2020-06-09
[#1112] Fix query plan caching for Hibernate and add config option to…

### LINKED ISSUE #1112: Invalid use of hibernate query plan cache when participating queries use limit/offset

Mobe91:
… enable/disable query plan caching

fixes #1112

### DIFF
--- a/core/api/src/main/java/com/blazebit/persistence/ConfigurationProperties.java
+++ b/core/api/src/main/java/com/blazebit/persistence/ConfigurationProperties.java
@@ -179,7 +179,19 @@ public final class ConfigurationProperties {
      * @see CTEBuilder#with(Class, CriteriaBuilder, boolean)
      */
     public static final String INLINE_CTES = "com.blazebit.persistence.inline_ctes";
-    
+
+    /**
+     * If set to true, the query plans are cached and reused.
+     * Valid values for this property are <code>true</code> and <code>false</code>.
+     * Default is <code>true</code>.
+     * This configuration option currently only takes effect when Hibernate is used as JPA provider.
+     *
+     * The property can be changed for a criteria builder before constructing a query.
+     *
+     * @since 1.5.0
+     */
+    public static final String QUERY_PLAN_CACHE_ENABLED = "com.blazebit.persistence.query_plan_cache_enabled";
+
     private ConfigurationProperties() {
     }
 }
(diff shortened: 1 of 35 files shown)


core/api/src/main/java/com/blazebit/persistence/ConfigurationProperties.java
```diff
@@ -179,7 +179,19 @@
...
+     * If set to true, the query plans are cached and reused.
+     * Valid values for this property are <code>true</code> and <code>false</code>.
+     * Default is <code>true</code>.
+     * This configuration option currently only takes effect when Hibernate is used as JPA provider.
+     *
+     * The property can be changed for a criteria builder before constructing a query.
+     *
+     * @since 1.5.0
+     */
+    public static final String QUERY_PLAN_CACHE_ENABLED = "com.blazebit.persistence.query_plan_cache_enabled";
```
jwgmeligmeyling, line 193: Isn't query caching opt-in anyway through `setCacheable(true)`?

Mobe91: `setCacheable(true)` affects query result caching.

core/impl/src/main/java/com/blazebit/persistence/impl/AbstractCTECriteriaBuilder.java
```diff
@@ -118,9 +118,11 @@ protected Query getQuery(String baseQueryString) {
             // The main query will handle that separately
             if (!isMainQuery) {
                 if (firstResult != 0) {
+                    query.setFirstResult(firstResult);
                     offset = Integer.toString(firstResult);
                 }
                 if (maxResults != Integer.MAX_VALUE) {
+                    query.setMaxResults(maxResults);
```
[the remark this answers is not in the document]
Mobe91: This is part of the actual fix. The other part is the extension of the query plan cache key to also consider limit/offset of participating queries.

core/api/src/main/java/com/blazebit/persistence/ConfigurationProperties.java
jwgmeligmeyling: Ah! My bad!

Mobe91:
Closed in favor of #1115

PR #1135, opened 2020-07-20
[#841] Add json functions

### LINKED ISSUE #841: Add JSON support functions and path navigation support
I'd like add cast functions for json and jsonb as well as add support for dereferencing a json attribute via normal path navigation I.e. `jsonAttribute.person.age`

In order to make this happen, we need a json metamodel in java that we can query to know what to cast a result to and to decide if a specific navigation operation is allowed. Maybe we could also support that without a model but then the user would have to do casts of the final values.

Using json access functions multiple times in a more complex expression could become inefficient, so we might want to consider wrapping jsonish expressions in a special function that takes care of optimizing this.

Selecting sub-objects of the JSON type should be possible by making use of a JSON binding provider in an internal ObjecBuilder.

In addition we can take a look at adding support for other json functions as well like e.g. functions for adding or removing objects from a JSON object.

This is blocked by #765 which will enable us to model foreign model structures and allow to easily de-reference it.

Mobe91:
Fixes #841


core/impl/src/main/java/com/blazebit/persistence/impl/function/jsonget/OracleJsonGetFunction.java
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * @author Moritz Becker
+ * @since 1.5.0
+ */
+public class OracleJsonGetFunction extends AbstractJsonGetFunction {
+
+    @Override
+    protected void render0(FunctionRenderContext context) {
+        Object[] jsonPathElements = new Object[context.getArgumentsSize() - 1];
```
beikov, line 29: Same as for JSON_SET. A utility method that accepts `FunctionRenderContext context, int offset` would be more appropriate.

```
        Object[] jsonPathElements = new Object[context.getArgumentsSize() - offset];
        for (int i = 0; i < jsonPathElements.length; i++) {
            try {
                jsonPathElements[i] = Integer.parseInt(JpqlFunctionUtil.unquoteSingleQuotes(context.getArgument(i + offset)));
            } catch (NumberFormatException e) {
                jsonPathElements[i] = JpqlFunctionUtil.unquoteDoubleQuotes(JpqlFunctionUtil.unquoteSingleQuotes(context.getArgument(i + offset)));
            }
        }
        return jsonPathElements;
```

core/impl/src/main/java/com/blazebit/persistence/impl/function/jsonget/MySQL8JsonGetFunction.java
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * @author Moritz Becker
+ * @since 1.5.0
+ */
+public class MySQL8JsonGetFunction extends AbstractJsonGetFunction {
+
+    @Override
+    protected void render0(FunctionRenderContext context) {
+        Object[] jsonPathElements = new Object[context.getArgumentsSize() - 1];
```
beikov, line 29: Same as for JSON_SET. A utility method that accepts `FunctionRenderContext context, int offset` would be more appropriate.

core/impl/src/main/java/com/blazebit/persistence/impl/function/jsonget/MSSQLJsonGetFunction.java
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * @author Moritz Becker
+ * @since 1.5.0
+ */
+public class MSSQLJsonGetFunction extends AbstractJsonGetFunction {
+
+    @Override
+    protected void render0(FunctionRenderContext context) {
+        Object[] jsonPathElements = new Object[context.getArgumentsSize() - 1];
```
beikov, line 29: It seems this can use the utility function I mentioned as well.

core/impl/src/main/java/com/blazebit/persistence/impl/function/jsonset/MSSQLJsonSetFunction.java
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * @author Moritz Becker
+ * @since 1.5.0
+ */
+public class MSSQLJsonSetFunction extends AbstractJsonGetFunction {
+
+    @Override
+    protected void render0(FunctionRenderContext context) {
+        Object[] jsonPathElements = new Object[context.getArgumentsSize() - 2];
```
beikov, line 30: It seems this can use the utility function I mentioned as well.

core/testsuite/src/test/java/com/blazebit/persistence/testsuite/JsonGetAndSetTest.java
```diff
@@ -0,0 +1,202 @@
...
+public class JsonGetAndSetTest extends AbstractCoreTest {
+
+    private final ObjectMapper objectMapper = new ObjectMapper();
+
+    @Override
+    protected Class<?>[] getEntityClasses() {
+        return new Class[] { JsonDocument.class };
+    }
+
+    @Parameterized.Parameters
```
beikov, line 55: Remove

core/impl/src/main/java/com/blazebit/persistence/impl/function/cast/DB2CastFunction.java
```diff
@@ -0,0 +1,100 @@
...
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private boolean isClobCompatibleCastTarget(String castTargetType) {
+        for (int i = 0; i < clobCompatibleCastTargetTypes.length; i++) {
+            if (castTargetType.equalsIgnoreCase(clobCompatibleCastTargetTypes[i]) ||
+                    castTargetType.toLowerCase().startsWith(clobCompatibleCastTargetTypes[i] + "(")) {
```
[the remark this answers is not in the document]
beikov: This is still missing.

core/impl/src/main/java/com/blazebit/persistence/impl/function/cast/DB2CastFunction.java
```diff
@@ -0,0 +1,100 @@
...
+ * @author Christian Beikov
+ * @since 1.2.0
+ */
+public class DB2CastFunction extends CastFunction {
+
+    private static final String[] clobReturningFunctions = new String[] {
+        "json_value(",
+        "json_query("
+    };
+    private static final String[] clobCompatibleCastTargetTypes = new String[] {
```
beikov, line 33:
```suggestion
    private static final String[] CLOB_COMPATIBLE_CAST_TARGET_TYPES = new String[] {
```

docker_db.sh
```diff
@@ -51,20 +51,15 @@ mssql() {
 
 oracle() {
     docker rm -f oracle || true
-    docker run --shm-size=1536m --name oracle -d -p 1521:1521 wnameless/oracle-xe-11g || echo "Clone and build the docker image from Github first.
-    git clone https://github.com/wnameless/docker-oracle-xe-11g.git
-    cd docker-oracle-xe-11g
-    docker build -t wnameless/oracle-xe-11g .
-    cd -
-    docker run --shm-size=1536m --name oracle -d -p 1521:1521 wnameless/oracle-xe-11g"
+    docker run --shm-size=1536m --name oracle -d -p 1521:1521 quillbuilduser/oracle-18-xe
```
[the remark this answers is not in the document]
Mobe91: What bat file?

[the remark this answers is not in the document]
beikov: Ah, we deleted that a while ago. But still, there are mentions about how to get the JDBC driver which are unnecessary now.

core/testsuite/src/main/basic/com/blazebit/persistence/testsuite/entity/JsonDocument.java
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * @author Moritz Becker
+ * @since 1.5.0
+ */
+@Entity
+@Table(name = "json_document")
+public class JsonDocument {
+    private Long id;
+    @Column(nullable = false, columnDefinition = "text")
```
beikov, line 31: I believe this annotation is ignored(it should lead to an error as not all DBMS support the text datatype) because you already annotated the getter with `@Id` which causes the use of PROPERTY access, leading to ignoring this.

Also, this class is the same as the one in src/main/hibernate. Why have both? Apart from that, you are missing a default constructor.

beikov:
I believe this error here https://travis-ci.com/github/Blazebit/blaze-persistence/jobs/363212273#L1696 is because Oracle doesn't support grouping by a blob? Maybe we need to adapt the tests that do this.

README.md
```diff
@@ -494,15 +494,6 @@ In IntelliJ when defining the Oracle database, go to the Advanced tab an specify
 
 When using the Oracle docker container via `docker_db.sh oracle` you might want to specify the following properties when executing tests `-Djdbc.url=jdbc:oracle:thin:@192.168.99.100:1521/xe -Djdbc.user=SYSTEM -Djdbc.password=oracle`
```
beikov, line 495:
```suggestion
When using the Oracle docker container via `docker_db.sh oracle` you might want to specify the following properties when executing tests `-Djdbc.url=jdbc:oracle:thin:@192.168.99.100:1521:XE -Djdbc.user=SYSTEM -Djdbc.password=Oracle18 -Doracle.jdbc.DateZeroTime=true`
```

entity-view/testsuite/pom.xml
```diff
@@ -430,6 +430,9 @@
                                 <!-- Careful, we need this otherwise the ordering will be wrong... -->
                                 <user.country>us</user.country>
                                 <user.language>en</user.language>
+                                <!-- Dumb, dumber, Oracle -->
```
beikov, line 433: You also need `<jdbc.url>jdbc:oracle:thin:@localhost:1521:XE</jdbc.url>` and `<jdbc.password>Oracle18</jdbc.password>`

integration/deltaspike-data/testsuite/pom.xml
```diff
@@ -379,6 +379,9 @@
                                 <jdbc.password>oracle</jdbc.password>
                                 <jdbc.driver>oracle.jdbc.driver.OracleDriver</jdbc.driver>
                                 <activeProfiles>${spring.activeProfiles}</activeProfiles>
+                                <!-- Dumb, dumber, Oracle -->
```
beikov, line 382: You also need `<jdbc.url>jdbc:oracle:thin:@localhost:1521:XE</jdbc.url>` and `<jdbc.password>Oracle18</jdbc.password>`

integration/querydsl/testsuite/pom.xml
```diff
@@ -349,6 +349,9 @@
                                 <!-- Careful, we need this otherwise the ordering will be wrong... -->
                                 <user.country>US</user.country>
                                 <user.language>en</user.language>
+                                <!-- Dumb, dumber, Oracle -->
```
beikov, line 352: You also need `<jdbc.url>jdbc:oracle:thin:@localhost:1521:XE</jdbc.url>` and `<jdbc.password>Oracle18</jdbc.password>`

integration/spring-data/testsuite/webflux/pom.xml
```diff
@@ -398,6 +398,9 @@
                                 <jdbc.password>oracle</jdbc.password>
                                 <jdbc.driver>oracle.jdbc.driver.OracleDriver</jdbc.driver>
                                 <activeProfiles>${spring.activeProfiles}</activeProfiles>
+                                <!-- Dumb, dumber, Oracle -->
```
beikov, line 401: You also need `<jdbc.url>jdbc:oracle:thin:@localhost:1521:XE</jdbc.url>` and `<jdbc.password>Oracle18</jdbc.password>`

integration/spring-data/testsuite/webmvc/pom.xml
```diff
@@ -408,6 +408,9 @@
                                 <jdbc.password>oracle</jdbc.password>
                                 <jdbc.driver>oracle.jdbc.driver.OracleDriver</jdbc.driver>
                                 <activeProfiles>${spring.activeProfiles}</activeProfiles>
+                                <!-- Dumb, dumber, Oracle -->
```
beikov, line 411: You also need `<jdbc.url>jdbc:oracle:thin:@localhost:1521:XE</jdbc.url>` and `<jdbc.password>Oracle18</jdbc.password>`

PR #1141, opened 2020-08-01
[#1129] Add default fromString parameter conversion for JAX-RS integr…

### LINKED ISSUE #1129: EntityViewId annotation has no effect in Quarkus resource
### Description
The [Quarkus guide on using blaze-persistence](https://quarkus.io/guides/blaze-persistence) and [blaze-persistence-examples-quarkus sample project](https://github.com/Blazebit/blaze-persistence/blob/master/examples/quarkus/base/src/main/java/com/blazebit/persistence/examples/quarkus/base/resource/PersonResource.java) indicate that the annotation `com.blazebit.persistence.integration.jaxrs.EntityViewId` can be used to bind a path parameter to an entity view id. However, the annotation does not seem to have any effect.

Additionally, it is not clear whether a project needs to declare `org.glassfish.jersey.core:jersey-common` as a dependency to use this annotation. If the dependency is not declared, Quarkus gives the following warning:

```
2020-07-11 00:39:36,389 WARN  [io.qua.arc.pro.BeanArchives] (build-20) Failed to index org.glassfish.jersey.model.Parameter: Class does not exist in ClassLoader QuarkusClassLoader:Deployment Class Loader
2020-07-11 00:39:36,389 INFO  [io.qua.arc.pro.IndexClassLookupUtils] (build-20) Class for name: org.glassfish.jersey.model.Parameter was not found in Jandex index. Please ensure the class is part of the index.
```

I noticed that `blaze-persistence-integration-jaxrs`, which contains `EntityViewId.class`, declares `jersey-common` as an optional dependency.

The annotation behaves the same regardless.

### Expected behavior
Entity view id is set to the value supplied in a path parameter.

### Actual behavior
Entity view id is null.

### Steps to reproduce
1. `git clone https://github.com/kekbur/blaze-persistence-quarkus-test.git && cd blaze-persistence-quarkus-test`
2. `./mvnw quarkus:dev`
3. `curl -H "Content-Type: application/json" -d '{"name":"Test"}' -X PUT "http://localhost:8080/persons/2c426b89-c102-45ad-8527-07cbbeb0a243"`

The PUT request above results in a `HTTP 400 - Bad Request` response, because [personUpdateView.getId() returns null](https://github.com/kekbur/blaze-persistence-quarkus-test/blob/06bf073e6c4bcd0d1b1606a3e566784bedd2b5a0/src/main/java/com/github/kekbur/blazepersistencequarkustest/resource/PersonResource.java#L52).

### Environment
Version:            1.5.0-Alpha5
JPA-Provider:       Hibernate 5.4.16.Final
DBMS:               H2
Application Framework: Quarkus 1.5.0.Final

Mobe91:
…ation

Fixes #1129

### DIFF
--- a/integration/jaxrs/src/main/java/com/blazebit/persistence/integration/jaxrs/EntityViewMessageBodyReader.java
+++ b/integration/jaxrs/src/main/java/com/blazebit/persistence/integration/jaxrs/EntityViewMessageBodyReader.java
@@ -29,6 +29,7 @@ import javax.annotation.Priority;
 import javax.enterprise.inject.Any;
 import javax.enterprise.inject.Instance;
 import javax.inject.Inject;
+import javax.ws.rs.BadRequestException;
 import javax.ws.rs.Consumes;
 import javax.ws.rs.Priorities;
 import javax.ws.rs.WebApplicationException;
@@ -44,7 +45,12 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.Reader;
 import java.lang.annotation.Annotation;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.lang.reflect.Type;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * @author Christian Beikov
@@ -55,6 +61,8 @@ import java.lang.reflect.Type;
 // "*/*" needs to be included since Jersey does not support the "application/*+json" notation
 @Consumes({"application/json", "application/*+json", "text/json", "*/*"})
 public class EntityViewMessageBodyReader implements MessageBodyReader<Object> {
+
+    private static final ParamConverterProvider FROM_STRING_PARAM_CONVERTER_PROVIDER = new FromStringParamConverterProvider();
 
     @Inject
     private Instance<EntityViewManager> entityViewManager;
@@ -85,7 +93,14 @@ public class EntityViewMessageBodyReader implements MessageBodyReader<Object> {
                                 break;
                             }
                         }
-                        return paramConverter == null ? null : paramConverter.fromString(value);
+                        if (paramConverter == null) {
+                            paramConverter = FROM_STRING_PARAM_CONVERTER_PROVIDER.getConverter(idType, idType, null);
+                        }
+                        if (paramConverter == null) {
+                            throw new RuntimeException("No " + ParamConverter.class.getName() + " could be found to convert to type " + idType.getName());
+                        } else {
+                            return paramConverter.fromString(value);
+                        }
                     }
                 }
             });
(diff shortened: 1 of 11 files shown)


integration/jackson/src/main/java/com/blazebit/persistence/integration/jackson/EntityViewAwareObjectMapper.java
```diff
@@ -136,12 +136,7 @@ public ObjectReader readerFor(JavaType javaType) {
         if (Collection.class.isAssignableFrom(javaType.getRawClass())) {
             return objectMapper.readerFor(javaType);
         } else {
-            return readerFor(javaType.getRawClass());
+            return objectMapper.readerFor(javaType.getRawClass());
         }
     }
-
-    public ObjectReader readerFor(Class<?> type) {
```
beikov, line 143: Please revert this. The idea of the method is that a user should be able to provide a custom object reader.

integration/quarkus/deployment/src/main/java/com/blazebit/persistence/integration/quarkus/deployment/BlazePersistenceProcessor.java
```diff
@@ -148,6 +166,15 @@ void reflection(EntityViewsBuildItem entityViewsBuildItem,
                 );
             }
         }
+
+        // add fromString methods from entity view id types
+        for (EntityViewMapping entityViewMapping : entityViewConfigurationBuildItem.getEntityViewConfiguration().getEntityViewMappings()) {
+            reflectionProducer.produce(ReflectiveClassBuildItem.builder(
+                    entityViewMapping.getIdAttribute().getDeclaredType())
+                    .methods(true)
```
beikov, line 174: Can't you provide a specific method signature instead?

integration/jaxrs/src/main/java/com/blazebit/persistence/integration/jaxrs/EntityViewMessageBodyReader.java
```diff
@@ -169,4 +181,46 @@ private boolean hasMatchingMediaType(MediaType mediaType) {
...
     }
+
+    /**
+     * ParamConverterProvider for default parameter conversion using fromString method if present
+     */
+    private static class FromStringParamConverterProvider implements ParamConverterProvider {
+        @Override
+        public <T> ParamConverter<T> getConverter(Class<T> clazz, Type type, Annotation[] annotations) {
+            try {
+                return new ParamConverter<>(clazz.getMethod("fromString", String.class));
```
beikov, line 192: According to the JAX-RS specification, here is a listing in the Resteasy documentation https://docs.jboss.org/resteasy/docs/3.5.0.Final/userguide/html/StringConverter.html#d4e1480, you should also support primitive types as well as types that have a public constructor that accepts just a string. In addition to that, you should also support a static method `valueOf`.
Apart from that, it would be good to cache the `ParamConverter` in a concurrent hash map within `FromStringParamConverterProvider`

integration/quarkus/deployment/src/main/java/com/blazebit/persistence/integration/quarkus/deployment/BlazePersistenceProcessor.java
Mobe91: No, the API does not allow for this.

PR #1163, opened 2020-09-11
Move to the new Quarkus datasource configuration

gsmet:
The new configuration was introduced in 1.3 and the old one deprecated.
We recently removed the support for the old configuration in master.


examples/quarkus/testsuite/base/src/test/resources/application.properties
```diff
@@ -1,21 +1,21 @@
-%postgresql.quarkus.datasource.url=jdbc:postgresql://localhost:5432/test
-%postgresql.quarkus.datasource.driver=org.postgresql.Driver
```
gsmet, line 2: So with the new config, we don't need to define the driver anymore... but I saw in `SchemaClearer.java` that you used it.

I was able to have the h2 tests pass so I'm not sure if we need to keep it.

If you absolutely need it, we just need to change `quarkus.datasource.driver` to `quarkus.datasource.jdbc.driver`. I can do it, just let me know.

PR #1172, opened 2020-10-02
[#448] Add support for secondary entity view roots

### LINKED ISSUE #448: Support secondary entity view roots
I am thinking of adding something like a named correlation on the entity view level.
This would allow to make use of a correlation in multiple mappings.

The name would only be _visible_ to mappings on the level at which it is defined. Making the name _globally_ available i.e. so that it can be used in a plain `CriteriaBuilder` is also interesting, although I am not sure how that works regarding possible alias duplicates. Maybe we could add a macro that resolves the actual join alias for a view root alias? Maybe forcing the use of such a macro is in general a good idea to make the use of correlated joins more apparent to readers?

@Mobe91 what do you think?

```java
@EntityView(Cat.class)
interface SimpleCatView {
  String getName();
}

@EntityView(Cat.class)
@EntityViewRoot(name = "sameAgedCat", entity = Cat.class, condition = "age = sameAgedCat.age", joinType = RootJoinType.INNER)
interface CatView {
  @IdMapping
  Integer getId();
  String getName();

  @Mapping("VIEW_ROOT(sameAgedCat)")
  List<SimpleCatView> getSameAgedCats();
}
```

In this example I overload the meaning of `VIEW_ROOT` to return the actual alias/path for a secondary root if the path equals a view root name. The resulting query might look like this

```sql
SELECT cat.id, cat.name, sameAgedCat_0.name
FROM Cat cat
INNER JOIN Cat sameAgedCat_0
        ON cat.age = sameAgedCat_0.age
```

### LINKED ISSUE #1173: Using correlation path for subquery from clause doesn't set fromClassExplicitlySet flag
The `from(String correlationPath)` implementation didn's set the `fromClassExplicitlySet` flag which leads to removing the correlation path join node and alias when adding further roots via the `from` methods on a `SubqueryBuilder`.

The workaround is pretty simple, just switch the order of the `from` calls to add the correlation path from clause call later.

beikov:
## Related Issue

Fixes #448. Fixes #1173

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ Not yet released
 
 * Fix Hibernate 5.4.19+ integration issue
 * Fix boot exception when using Hibernate `@Where` on entity level
+* Properly set `fromClassExplicitlySet` flag when using the subquery `from(String correlationPath)` methods
 
 ### Backwards-incompatible changes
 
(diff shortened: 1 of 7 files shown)


riteshka:
The getCountQuery works as expected when the underlying dialect doesn't support Select Distinct on multiple columns.

PR #1196, opened 2020-11-06
[#1195] Allow overriding the ObjectMapper used for deserializing entity views in the JAX-RS integration

### LINKED ISSUE #1195: Unable to deserialize Instant object from ISO 8601 (with jax-rs integration)
Starting with blaze 1.5, I cannot find a way to deserialize Instant object from ISO 8601 date in Entity View.

As stated in 1.4 documentation, blaze used existing ObjectMapper to deserialize entityView and could therefore be configured (https://persistence.blazebit.com/documentation/1.4/entity-view/manual/en_US/index.html#usage-5)

Starting with 1.5, (changed in [this commit](https://github.com/Blazebit/blaze-persistence/commit/3785a964c4831aa1c432299d4e74e3ac62e37a00) ) the object mapper used is created from scratch ([see here](https://github.com/Blazebit/blaze-persistence/blob/1.5/integration/jaxrs/src/main/java/com/blazebit/persistence/integration/jaxrs/EntityViewMessageBodyReader.java#L83), and cannot seams to be accessed in any way to be configured, for example for supporting ISO 8601 dates format, which is a common use case :

```
        objectMapper.registerModule(new JavaTimeModule());
        objectMapper.disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
```

### Description
Have a way to configure blaze internal objectMapper to support ISO 8601 dates format for example

### Expected behavior
Post following json object and process it correctly through jax-rs in EntityView :
{"datetime": "2018-12-04T18:47:38.927Z"}

### Actual behavior
`com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Cannot construct instance of `java.time.Instant` (no Creators, like default constructor, exist): no String-argument constructor/factory method to deserialize from String value ('2020-11-09T08:00:00.000Z')`

### Steps to reproduce
Submit the given json as body to a post endpoint (with jax-rs blaze integration) and building a simple entity view :
`@UpdatableEntityView
@CreatableEntityView
@EntityView(MyEntity.class)
public interface MyUpdateView {
    @IdMapping
    Long getId();

    Instant getDatetime();
    void setDatetime(Instant datetime);
}`

### Environment
Blaze 1.5 and upper

beikov:
## Related Issue

Fixes #1195

### DIFF
--- a/integration/jaxrs/src/main/java/com/blazebit/persistence/integration/jaxrs/EntityViewMessageBodyReader.java
+++ b/integration/jaxrs/src/main/java/com/blazebit/persistence/integration/jaxrs/EntityViewMessageBodyReader.java
@@ -37,10 +37,12 @@ import javax.ws.rs.core.Context;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.MultivaluedMap;
 import javax.ws.rs.core.UriInfo;
+import javax.ws.rs.ext.ContextResolver;
 import javax.ws.rs.ext.MessageBodyReader;
 import javax.ws.rs.ext.ParamConverter;
 import javax.ws.rs.ext.ParamConverterProvider;
 import javax.ws.rs.ext.Provider;
+import javax.ws.rs.ext.Providers;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.Reader;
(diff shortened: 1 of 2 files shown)


integration/jaxrs/src/main/java/com/blazebit/persistence/integration/jaxrs/EntityViewMessageBodyReader.java
```diff
@@ -67,6 +67,8 @@
     @Inject
     private Instance<EntityViewManager> entityViewManager;
     @Inject
+    private Instance<ObjectMapper> objectMapper;
```
jwgmeligmeyling, line 70: Normally in JAX-RS I use a `ContextResolver` for providing the `ObjectMapper`, as an `ObjectMapper` provided through a `ContextResolver` will also be recognized by the `ResteasyJackson2Provider`.


One would obtain this through:

```java
@Context
protected Providers _providers;
```

And then:

```java
ContextResolver<ObjectMapper> resolver = _providers.getContextResolver(ObjectMapper.class, mediaType);
```

Just my 2 cents that you could consider looking up an `ObjectMapper` though this as well (possibly extra), for applications that rely on this mechanism already.

beikov: Thanks, I adapted that

PR #1199, opened 2020-11-10
 [#1198] Support aliases in CTE binds

### LINKED ISSUE #1198: When bind cte can't as alias
image

jwgmeligmeyling:
Support aliases in CTE binds

Fixes #1198

### DIFF
--- a/integration/querydsl/expressions/src/main/java/com/blazebit/persistence/querydsl/BlazeCriteriaBuilderRenderer.java
+++ b/integration/querydsl/expressions/src/main/java/com/blazebit/persistence/querydsl/BlazeCriteriaBuilderRenderer.java
@@ -295,26 +295,21 @@ public class BlazeCriteriaBuilderRenderer<T> {
                         } else {
                             for (int i = 0; i < projection.size(); i++) {
                                 Expression<?> projExpression = projection.get(i);
-                                Path<?> alias = null;
-                                Path<?> cteEntityPath = null;
 
-                                if (projExpression instanceof Operation) {
-                                    Operation<?> projOperation = (Operation<?>) projExpression;
-                                    if (projOperation.getOperator() == BIND) {
-                                        alias = (Path<?>) projOperation.getArg(1);
-                                        cteEntityPath = alias.getRoot();
-                                        projExpression = projOperation.getArg(0);
-                                    }
+                                BindResolverContext bindResolverContext = new BindResolverContext();
+                                projExpression = projExpression.accept(BindResolver.INSTANCE, bindResolverContext);
+                                Path<?> cteAttribute = bindResolverContext.getCteAttribute();
+                                String alias = bindResolverContext.getAliasString();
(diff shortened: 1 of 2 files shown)


integration/querydsl/expressions/src/main/java/com/blazebit/persistence/querydsl/BlazeCriteriaBuilderRenderer.java
```diff
@@ -1342,4 +1337,92 @@ public Object visit(TemplateExpression<?> templateExpression, Object o) {
...
+        }
+
+    }
+
+    /**
+     * Resolve the bind and alias information for an expression.
+     *
+     * @since 1.6.0
+     */
+    private static class BindResolver implements Visitor<Expression<?>, BindResolverContext> {
```
jwgmeligmeyling, line 1379: I need to return three values here: the alias, the bind attribute path and the select expression. Should I integrate the Context and the Visitor? I've decided for now to keep the state seperate from the visitor.

beikov: It's fine either way. Whatever you prefer.

beikov (approved):

integration/querydsl/expressions/src/main/java/com/blazebit/persistence/querydsl/JPQLNextExpressions.java
```diff
@@ -32,6 +32,7 @@
 import com.querydsl.core.types.dsl.DateTimeExpression;
 import com.querydsl.core.types.dsl.Expressions;
 import com.querydsl.core.types.dsl.NumberExpression;
+import com.querydsl.core.types.dsl.SimpleOperation;
```
beikov, line 35: [ERROR] src/main/java/com/blazebit/persistence/querydsl/JPQLNextExpressions.java:[24,8] (imports) UnusedImports: Unused import - com.querydsl.core.types.ExpressionUtils.
[ERROR] src/main/java/com/blazebit/persistence/querydsl/JPQLNextExpressions.java:[25,8] (imports) UnusedImports: Unused import - com.querydsl.core.types.Operation.

jwgmeligmeyling: Oops.

jwgmeligmeyling: Fixed. I actually reverted this change. In hindsight I think its more appropiate to put the alias within the bind operation rather than outside it. Besides `.as(alias)` , `SimpleOperation` also introduces methods like `isNull()`, `isNotNull()` which could lead to improper use (as the bind and the alias always need to be the outermost expression and are only allowed in some scenarios).

PR #1204, opened 2020-11-14
Fix QueryDSL integration documentation

jwgmeligmeyling:
@AFbenHassine took the time to review the QueryDSL module and its documentation. This PR brings a few improvements to the documentation based on his questions and remarks.

- Fixed indentation for some headers
- Remove confusing use of recursive CTE's where any Entity would be fine
- Remove generic qualification where not required
- A little bit more explanation how the values clauses work and what the alias is for
- Some queries referred to a "cat" alias in a "Book" variable, I fixed those ;-)


beikov:
Thanks!

PR #1205, opened 2020-11-15
[#1050] Implement emulation for temporal literals through functions also for Java 8 types

### LINKED ISSUE #1050: Determine supported types for temporal literals and register converters
I'm not sure which temporal literal types the JPA providers support, but it would be nice if we could determine that and register converters in `com.blazebit.persistence.parser.util.TypeUtils` for these type. The following types need to be checked

* LocalDate
* LocalTime
* LocalDateTime
* Instant
* ZoneDateTime
* OffsetDateTime
* OffsetTime

The check is to use a model like `@Entity class MyEntity { Instant temporalValue; }` and see if a query like this works `SELECT e FROM MyEntity e WHERE e.temporalValue = {ts '2020-01-01 00:00:00'}` i.e. check if the JPA provider can handle the temporal literal for a field of a Java 8 time type.

We could introduce a `TypeConverterContributor` in core-impl that renders something like `LITERAL_LOCAL_DATE('...')`. The function would render SQL like `{ts '...'}` and return the respective type as returned type. We could also do that to emulate normal temporal literals for Hibernate.

beikov:
## Related Issue

Fixes #1050

### DIFF
--- a/documentation/src/main/asciidoc/core/manual/en_US/jpql_functions.adoc
+++ b/documentation/src/main/asciidoc/core/manual/en_US/jpql_functions.adoc
@@ -208,11 +208,44 @@ There are multiple different treat functions for different data types.
 * `java.sql.Date` - `TREAT_DATE`
 * `java.sql.Timestamp` - `TREAT_TIMESTAMP`
 * `java.util.Calendar` - `TREAT_CALENDAR`
+* `java.time.Instant` - `TREAT_INSTANT`
+* `java.time.LocalDate` - `TREAT_LOCALDATE`
+* `java.time.LocalDateTime` - `TREAT_LOCALDATETIME`
+* `java.time.LocalTime` - `TREAT_LOCALTIME`
+* `java.time.OffsetDateTime` - `TREAT_OFFSETDATETIME`
+* `java.time.OffsetTime` - `TREAT_OFFSETTIME`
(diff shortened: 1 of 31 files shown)


documentation/src/main/asciidoc/core/manual/en_US/jpql_functions.adoc
```diff
@@ -214,6 +214,32 @@ This can be used for cases when the type of an expression is actually known but
...
+
+There are multiple different literal functions for different data types.
+
+* `java.util.Calendar` - `LITERAL_CALENDAR`
+* `java.sql.Time` - `LITERAL_TIME`
+* `java.sql.Date` - `LITERAL_DATE`
+* `java.sql.Timestamp` - `LITERAL_TIMESTAMP`
+* `java.util.Date` - `LITERAL_UTIL_DATE`
+* `java.time.Instant` - `LITERAL_INSTANT`
+* `java.time.LocalDate` - `LITERAL_LOCAL_DATE`
```
jwgmeligmeyling, line 231: Don't we want a `TREAT_LOCAL_DATE` following the same mechanism as well?

beikov: Thanks for the hint. We already have that and I now added the documentation for this.

PR #1211, opened 2020-11-19
Add missing alias to code example in documentation

sullrich84:
## Description
Added missing alias to the given example.


beikov:
Hi @sullrich84, the alias is not required in this case as the name is resolved implicitly against the root alias. In fact, for the `set` call, it is even required to use the attribute name instead of a path like `c.name`. Did you have any difficulties with understanding the example or with actual code?

sullrich84:
Sorry, misunderstood the example...

beikov:
No problem and thanks for checking this :)

PR #1229, opened 2020-12-09
[#632] Allow to convert entities to entity views

### LINKED ISSUE #632: Allow to convert entities to entity views
We already have our conversion API in the `EntityViewManager` which supports conversion between entity view types. When we have a proper JPQL interpreter, it should be possible to implement conversion of entities to entity views.
This is a follow up of https://github.com/Blazebit/blaze-persistence/issues/127

danelowe:
## Description
Attempting to understand how entities can be converted to views

## Related Issue
Fixes #632

## Motivation and Context
Would like to be able to operate on a domain entity, persist it with hibernate, then return a DTO by converting the entity, without having to manually build the DTO.

### DIFF
--- a/entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/mapper/ConvertOperationBuilderImpl.java
+++ b/entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/mapper/ConvertOperationBuilderImpl.java
@@ -54,7 +54,11 @@ public class ConvertOperationBuilderImpl<T> implements ConvertOperationBuilder<T
 
     @Override
     public ConvertOperationBuilder<T> excludeAttribute(String attributePath) {
-        if (!key.getSourceType().getRecursiveAttributes().containsKey(attributePath) && !key.getSourceType().getRecursiveSubviewAttributes().containsKey(attributePath)) {
+        if (
+                (key.getSourceType() != null)
+                && !key.getSourceType().getRecursiveAttributes().containsKey(attributePath)
+                && !key.getSourceType().getRecursiveSubviewAttributes().containsKey(attributePath)
+        ) {
             throw new IllegalArgumentException("Attribute '" + attributePath + "' could not be found on type: " + key.getSourceType().getJavaType().getName());
         }
         subMappers.put(attributePath, ViewMapper.Key.EXCLUDE_MARKER);
@@ -83,6 +87,11 @@ public class ConvertOperationBuilderImpl<T> implements ConvertOperationBuilder<T
 
     @Override
     public ConvertOperationBuilder<T> convertAttribute(String attributePath, Class<?> attributeViewClass, String constructorName, ConvertOption... convertOptions) {
+        if (key.getSourceType() == null) {
+            subMappers.put(attributePath, (ViewMapper.Key<Object, Object>) ViewMapper.Key.create(entityViewManager.getMetamodel(), null, attributeViewClass, constructorName, convertOptions));
+            return this;
+        }
+
         AbstractMethodAttribute<? super Object, ?> sourceAttribute = key.getSourceType().getRecursiveSubviewAttributes().get(attributePath);
         if (sourceAttribute == null) {
             throw new IllegalArgumentException("Attribute '" + attributePath + "' could not be found on type: " + key.getSourceType().getJavaType().getName());
--- a/entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/mapper/ConvertOperationBuilderImpl.java
+++ b/entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/mapper/ConvertOperationBuilderImpl.java
@@ -54,12 +54,8 @@ public class ConvertOperationBuilderImpl<T> implements ConvertOperationBuilder<T
 
     @Override
     public ConvertOperationBuilder<T> excludeAttribute(String attributePath) {
-        if (
-                (key.getSourceType() != null)
-                && !key.getSourceType().getRecursiveAttributes().containsKey(attributePath)
-                && !key.getSourceType().getRecursiveSubviewAttributes().containsKey(attributePath)
-        ) {
-            throw new IllegalArgumentException("Attribute '" + attributePath + "' could not be found on type: " + key.getSourceType().getJavaType().getName());
+        if (!key.getTargetType().getRecursiveAttributes().containsKey(attributePath) && !key.getTargetType().getRecursiveSubviewAttributes().containsKey(attributePath)) {
+            throw new IllegalArgumentException("Attribute '" + attributePath + "' could not be found on type: " + key.getTargetType().getJavaType().getName());
         }
         subMappers.put(attributePath, ViewMapper.Key.EXCLUDE_MARKER);
         return this;
@@ -87,8 +83,20 @@ public class ConvertOperationBuilderImpl<T> implements ConvertOperationBuilder<T
 
     @Override
     public ConvertOperationBuilder<T> convertAttribute(String attributePath, Class<?> attributeViewClass, String constructorName, ConvertOption... convertOptions) {
+        AbstractMethodAttribute<?, ?> targetAttribute = key.getTargetType().getRecursiveSubviewAttributes().get(attributePath);
+        if (targetAttribute == null) {
+            throw new IllegalArgumentException("Attribute '" + attributePath + "' could not be found on type: " + key.getTargetType().getJavaType().getName());
+        }
+        Class<?> targetType = targetAttribute.getConvertedJavaType();
+        if (targetAttribute instanceof PluralAttribute<?, ?, ?>) {
+            targetType = ((PluralAttribute<?, ?, ?>) targetAttribute).getElementType().getJavaType();
+        }
+        if (!targetType.isAssignableFrom(attributeViewClass)) {
+            throw new IllegalArgumentException("The given type '" + attributeViewClass.getName() + "' is not assignable to the declared type of '" + key.getTargetType().getJavaType().getName() + "." + attributePath + "': " + targetType.getName());
+        }
+
         if (key.getSourceType() == null) {
-            subMappers.put(attributePath, (ViewMapper.Key<Object, Object>) ViewMapper.Key.create(entityViewManager.getMetamodel(), null, attributeViewClass, constructorName, convertOptions));
+            subMappers.put(attributePath, (ViewMapper.Key<Object, Object>) ViewMapper.Key.create(entityViewManager.getMetamodel(), attributeViewClass, constructorName, convertOptions));
             return this;
         }
 
--- a/entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/mapper/ViewMapper.java
+++ b/entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/mapper/ViewMapper.java
@@ -18,6 +18,7 @@ package com.blazebit.persistence.view.impl.mapper;
 
 import com.blazebit.persistence.view.ConvertOption;
 import com.blazebit.persistence.view.EntityViewManager;
+import com.blazebit.persistence.view.impl.EntityViewManagerImpl;
 import com.blazebit.persistence.view.impl.accessor.Accessors;
 import com.blazebit.persistence.view.impl.accessor.AttributeAccessor;
 import com.blazebit.persistence.view.impl.collection.CollectionInstantiatorImplementor;
@@ -74,7 +75,7 @@ public class ViewMapper<S, T> {
     private final boolean postConvertUsesSource;
 
     public ViewMapper(ManagedViewType<S> sourceType, ManagedViewType<T> targetType, MappingConstructorImpl<T> targetConstructor, boolean ignoreMissing, Boolean maybeMarkNew, EntityViewManager entityViewManager, ProxyFactory proxyFactory, String prefix, Map<String, Key<Object, Object>> subMappers) {
-        if (!targetType.getEntityClass().isAssignableFrom(sourceType.getEntityClass())) {
+        if ((sourceType != null) && !targetType.getEntityClass().isAssignableFrom(sourceType.getEntityClass())) {
             throw inconvertible("Incompatible entity types!", sourceType, targetType);
         }
 
@@ -113,7 +114,8 @@ public class ViewMapper<S, T> {
                 // Extract a mapping from target dirty state index to the source
                 int dirtyStateIndex = ((AbstractMethodAttribute<?, ?>) targetAttribute).getDirtyStateIndex();
                 if (dirtyStateIndex != -1) {
-                    MethodAttribute<? super S, ?> sourceAttribute = sourceType.getAttribute(targetAttribute.getName());
+                    // Does dirty state need to be copied from entities? Can this be done without modifying the entity class?
+                    MethodAttribute<? super S, ?> sourceAttribute = (sourceType == null) ? null : sourceType.getAttribute(targetAttribute.getName());
                     if (sourceAttribute != null) {
                         int sourceIndex = ((AbstractMethodAttribute<?, ?>) sourceAttribute).getDirtyStateIndex();
                         if (sourceIndex != -1) {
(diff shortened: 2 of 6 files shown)


entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/metamodel/ViewMetamodelImpl.java
```diff
@@ -88,6 +89,12 @@ public ViewMetamodelImpl(EntityMetamodel entityMetamodel, MetamodelBuildingConte
             ManagedViewTypeImplementor<?> managedView = context.getManagedViewType(viewMapping, null);
 
             managedViews.put(viewMapping.getEntityViewClass(), managedView);
+
+            Class<?> entityClass = viewMapping.getEntityClass();
+            if (!entitySources.containsKey(entityClass)) {
```
danelowe, line 94: @beikov I assume that the `*TypeImpl` (e.g. `ManagedViewTypeImplementor` above) class is meant to hold metadata about the entity/view that is gathered when entityViewManager is created. However I'm struggling to work out if the required metadata is already available at this stage via `managedView` or whether I'd need to implement the gathering of metadata about the entity, and create instances of `AbstractAttribute<?, ?>`

entity-view/api/src/main/java/com/blazebit/persistence/view/metamodel/SourceType.java
```diff
@@ -0,0 +1,18 @@
+package com.blazebit.persistence.view.metamodel;
+
+public interface SourceType<X> extends Type<X> {
```
danelowe, line 3: @beikov I'm not sure if it is a good idea for me to split these interfaces, but in my mind it made sense to have a separate interface for sources and targets when converting?

beikov:
@danelowe I think it would be better to make `ViewMapper#sourceType` nullable and adapt the constructor to use `com.blazebit.persistence.view.impl.accessor.Accessors#forEntityMapping(com.blazebit.persistence.view.impl.EntityViewManagerImpl, com.blazebit.persistence.view.metamodel.MethodAttribute<?,?>)` based on the target attribute if the source type is null. This way you don't have to introduce new interfaces and can reuse the existing accessor logic.

entity-view/testsuite/src/test/java/com/blazebit/persistence/view/testsuite/convert/view/ConvertEntityTest.java
```diff
@@ -0,0 +1,156 @@
...
+        assertEquals(doc.getPartners().iterator().next().getName(), clone.getPartners().iterator().next().getName());
+        assertNull(clone.getPartners().iterator().next().getFriend().getId());
+        assertEquals(doc.getPartners().iterator().next().getFriend().getName(), clone.getPartners().iterator().next().getFriend().getName());
+
+        assertNull(clone.getPeople());
+        assertTrue(((EntityViewProxy) clone.getOwner().getFriend()).$$_isNew());
+        assertTrue(clone.getOwner().getFriend() instanceof CreatablePersonView);
+        assertTrue(doc == clone.getSource());
+    }
+
```
danelowe, line 146: com.blazebit.persistence.view.testsuite.convert.view.ConvertViewTest#testCloneConvert2 is a test for `.withOptionalParameter()` and `@MappingParameter`. I'm not sure if this functionality would apply to entities.

entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/mapper/ViewMapper.java
```diff
@@ -113,7 +114,8 @@ public ViewMapper(ManagedViewType<S> sourceType, ManagedViewType<T> targetType,
                 // Extract a mapping from target dirty state index to the source
                 int dirtyStateIndex = ((AbstractMethodAttribute<?, ?>) targetAttribute).getDirtyStateIndex();
                 if (dirtyStateIndex != -1) {
-                    MethodAttribute<? super S, ?> sourceAttribute = sourceType.getAttribute(targetAttribute.getName());
+                    // Does dirty state need to be copied from entities? Can this be done without modifying the entity class?
+                    MethodAttribute<? super S, ?> sourceAttribute = (sourceType == null) ? null :sourceType.getAttribute(targetAttribute.getName());
```
danelowe, line 118: I'm assuming copying dirty state from an entity would be out of scope

danelowe:
@beikov I usually use Kotlin in preference to Java, so I'm not really sure if the way I'm dealing with nulls is correct, and 90% of this PR seems to be dealing with nulls. 

As an aside, interop with Kotlin could be improved with [Nullability Annotations](https://kotlinlang.org/docs/reference/java-interop.html#nullability-annotations), which would let Kotlin know which parameters are meant to be nullable. In this case, the change doesn't make anything nullable in the public api though.

Is there any convention in Java to indicate that the `sourceType` parameter is intended to be nullable?

entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/mapper/ConvertOperationBuilderImpl.java
```diff
@@ -54,7 +54,11 @@ public T convert() {
 
     @Override
     public ConvertOperationBuilder<T> excludeAttribute(String attributePath) {
-        if (!key.getSourceType().getRecursiveAttributes().containsKey(attributePath) && !key.getSourceType().getRecursiveSubviewAttributes().containsKey(attributePath)) {
+        if (
+                (key.getSourceType() != null)
+                && !key.getSourceType().getRecursiveAttributes().containsKey(attributePath)
+                && !key.getSourceType().getRecursiveSubviewAttributes().containsKey(attributePath)
+        ) {
             throw new IllegalArgumentException("Attribute '" + attributePath + "' could not be found on type: " + key.getSourceType().getJavaType().getName());
```
beikov, line 62: I think it's more appropriate to use the target type here.
```suggestion
        if (!key.getTargetType().getRecursiveAttributes().containsKey(attributePath) && !key.getTargetType().getRecursiveSubviewAttributes().containsKey(attributePath)) {
                    throw new IllegalArgumentException("Attribute '" + attributePath + "' could not be found on type: " + key.getTargetType().getJavaType().getName());
```

and also in the `excludeAttributes(String...)` below. Previously it didn't make much difference because views had to be compatible i.e. both contain the same/similar attributes, but it's more correct to base this on the target attributes.

entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/mapper/ConvertOperationBuilderImpl.java
```diff
@@ -83,6 +87,11 @@ public T convert() {
 
     @Override
     public ConvertOperationBuilder<T> convertAttribute(String attributePath, Class<?> attributeViewClass, String constructorName, ConvertOption... convertOptions) {
+        if (key.getSourceType() == null) {
+            subMappers.put(attributePath, (ViewMapper.Key<Object, Object>) ViewMapper.Key.create(entityViewManager.getMetamodel(), null, attributeViewClass, constructorName, convertOptions));
+            return this;
+        }
```
beikov, line 93: We should actually also check if the target type is compatible with the given `attributeViewClass`.

```suggestion
        AbstractMethodAttribute<? super Object, ?> targetAttribute = key.getTargetType().getRecursiveSubviewAttributes().get(attributePath);
        if (targetAttribute == null) {
            throw new IllegalArgumentException("Attribute '" + attributePath + "' could not be found on type: " + key.getTargetType().getJavaType().getName());
        }
        Class<?> targetType = targetAttribute.getConvertedJavaType();
        if (targetAttribute instanceof PluralAttribute<?, ?, ?>) {
            targetType = ((PluralAttribute<?, ?, ?>) targetAttribute).getElementType().getJavaType();
        }
        if (!targetType.isAssignableFrom(attributeViewClass)) {
            throw new IllegalArgumentException("The given type '" + attributeViewClass.getName() + "' is not assignable to the declared type of '" + key.getTargetType().getJavaType().getName() + "." + attributePath + "': " + targetType.getName());
        }
        if (key.getSourceType() == null) {
            subMappers.put(attributePath, (ViewMapper.Key<Object, Object>) ViewMapper.Key.create(entityViewManager.getMetamodel(), null, attributeViewClass, constructorName, convertOptions));
            return this;
        }
```

entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/mapper/ViewMapper.java
```diff
@@ -185,9 +187,15 @@ private ObjectMapper createAccessor(ManagedViewType<S> sourceType, ManagedViewTy
...
+        MethodAttribute<?, ?> sourceAttribute = (sourceType == null) ? null : sourceType.getAttribute(targetAttribute.getName());
+        AttributeAccessor accessor = (sourceAttribute == null)
+                ? Accessors.forEntityMapping((EntityViewManagerImpl) entityViewManager, targetAttribute)
+                : Accessors.forViewAttribute(null, sourceAttribute, true);
+        if (accessor == null) {
+            return null;
+        }
 
-        if (sourceAttribute == null) {
+        if ((sourceType != null) && (sourceAttribute == null)) {
```
beikov, line 198: Returning null when the accessor is null is not a good idea I think. Maybe something along these lines is more appropriate.

```suggestion
        MethodAttribute<?, ?> sourceAttribute;
        AttributeAccessor accessor;
        if (sourceType == null) {
            sourceAttribute = null;
            accessor = Accessors.forEntityMapping((EntityViewManagerImpl) entityViewManager, targetAttribute);
        } else {
            sourceAttribute = sourceType.getAttribute(targetAttribute.getName());
            if (sourceAttribute == null) {
                if (targetAttribute.getMappingType() == Attribute.MappingType.PARAMETER) {
                    return new ParameterObjectMapper(((MappingAttribute<?, ?>) targetAttribute).getMapping());
                }
                // Optionally ignore missing attributes
                if (ignoreMissing) {
                    return null;
                }
                throw inconvertible("Attribute '" + targetAttribute.getName() + "' from target type is missing in source type!", sourceType, targetType);
            }
            accessor = Accessors.forViewAttribute(null, sourceAttribute, true);
        }
```

entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/mapper/ViewMapper.java
```diff
@@ -409,9 +420,6 @@ public Key(ManagedViewTypeImplementor<S> sourceType, ManagedViewTypeImplementor<
 
         public static <X, Y> Key<X, Y> create(ViewMetamodel metamodel, Class<X> sourceEntityViewClass, Class<Y> targetEntityViewClass, String targetConstructorName, boolean ignoreMissingAttributes, boolean markNew) {
             ManagedViewTypeImplementor<X> sourceViewType = (ManagedViewTypeImplementor<X>) metamodel.managedView(sourceEntityViewClass);
-            if (sourceViewType == null) {
-                throw new IllegalArgumentException("Unknown source view type: " + sourceEntityViewClass.getName());
-            }
```
beikov, line 414: Please revert this change. I'd prefer to handle this in the create method that receives the object.

entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/mapper/ViewMapper.java
```diff
@@ -381,12 +391,13 @@ public Key(ManagedViewTypeImplementor<S> sourceType, ManagedViewTypeImplementor<
         }
 
         public static <Y> Key<Object, Y> create(ViewMetamodel metamodel, Object source, Class<Y> targetEntityViewClass, String constructorName, ConvertOption... convertOptions) {
-            if (!(source instanceof EntityViewProxy)) {
-                throw new IllegalArgumentException("Can only convert one entity view to another. Invalid source type: " + source.getClass());
+            if (source instanceof EntityViewProxy) {
+                EntityViewProxy sourceProxy = (EntityViewProxy) source;
+                return (Key<Object, Y>) create(metamodel, sourceProxy.$$_getEntityViewClass(), targetEntityViewClass, constructorName, convertOptions);
+            } else {
+                return (Key<Object, Y>) create(metamodel, null, targetEntityViewClass, constructorName, convertOptions);
```
beikov, line 398: IMO we should validate that the entity class of the target entity view is a super type of the source object.

```suggestion
            ManagedViewType<?> targetViewType = metamodel.managedView(targetEntityViewClass);
            if (targetViewType == null) {
                throw new IllegalArgumentException("Unknown target view type: " + targetEntityViewClass.getName());
            }
            if (!targetViewType.getEntityClass().isInstance(source)) {
                throw new IllegalArgumentException("The source object is not an instance of the target views entity type " + targetViewType.getEntityClass().getName() + ": " + source.getClass().getName());
            }
            MappingConstructorImpl<Y> constructor = null;
            if (targetConstructorName != null) {
                constructor = (MappingConstructorImpl<Y>) targetViewType.getConstructor(targetConstructorName);
                if (constructor == null) {
                    throw new IllegalArgumentException("Unknown constructor '" + targetConstructorName + "' on target view type: " + targetEntityViewClass.getName());
                }
            }
            boolean ignoreMissingAttributes = false;
            boolean markNew = false;
            for (ConvertOption copyOption : convertOptions) {
                switch (copyOption) {
                    case CREATE_NEW:
                        markNew = true;
                        break;
                    case IGNORE_MISSING_ATTRIBUTES:
                        ignoreMissingAttributes = true;
                        break;
                    default:
                        break;
                }
            }
            return new ViewMapper.Key<>(sourceViewType, targetViewType, constructor, ignoreMissingAttributes, markNew);
```

beikov (changes requested):
Looks great so far! I added a few comments regarding type validation issues. Could you please also add a test that checks if converting a non-compatible entity to a view fails as expected? Converting a `Person` to a `DocumentCloneView` should fail.

entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/mapper/ViewMapper.java
beikov: Right, that's simply not possible.

beikov:
And before I forget it, could you please also add a line in the CHANGELOG.md about this new feature i.e. add something like `* Support for converting entities to entity views`

beikov:
> @beikov I usually use Kotlin in preference to Java, so I'm not really sure if the way I'm dealing with nulls is correct, and 90% of this PR seems to be dealing with nulls.

Looks ok to me.

> As an aside, interop with Kotlin could be improved with [Nullability Annotations](https://kotlinlang.org/docs/reference/java-interop.html#nullability-annotations), which would let Kotlin know which parameters are meant to be nullable. In this case, the change doesn't make anything nullable in the public api though.

I'm open for contributions in this area if this helps you work with the API. So far, the value of this didn't justify the amount of work that is necessary for me yet. I think I prefer the JetBrains annotations, so if you want to work on this, use this dependency:

```
<dependency>
  <groupId>org.jetbrains</groupId>
  <artifactId>annotations</artifactId>
  <version>20.1.0</version>
  <scope>provided</scope>
</dependency>
```

Personally, I'd like to use non-null by default and annotate `@Nullable`.

> Is there any convention in Java to indicate that the `sourceType` parameter is intended to be nullable?

Only through the annotations, but that's "just" a hint/contract for frameworks/IDEs. There are not guarantees by the language.

entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/mapper/ViewMapper.java
```diff
@@ -185,26 +187,38 @@ private ObjectMapper createAccessor(ManagedViewType<S> sourceType, ManagedViewTy
...
+        MethodAttribute<?, ?> sourceAttribute;
+        AttributeAccessor accessor;
+        if (sourceType == null) {
+            sourceAttribute = null;
+            accessor = Accessors.forEntityMapping((EntityViewManagerImpl) entityViewManager, targetAttribute);
+            if (ignoreMissing || (targetAttribute.getMappingType() == Attribute.MappingType.PARAMETER)) {
                 return null;
+            } else if (accessor == null) {
+                throw inconvertible("Attribute '" + targetAttribute.getName() + "' from target type is missing in source type!", targetType);
+            }
```
beikov, line 199: IMO parameter mappings should be supported and null should only be returned if the accessor is null.

```suggestion
            if (targetAttribute.getMappingType() == Attribute.MappingType.PARAMETER) {
                return new ParameterObjectMapper(((MappingAttribute<?, ?>) targetAttribute).getMapping());
            }
            accessor = Accessors.forEntityMapping((EntityViewManagerImpl) entityViewManager, targetAttribute);
            if (accessor == null) {
                if (ignoreMissing) {
                    return null;
                }
                throw inconvertible("Attribute '" + targetAttribute.getName() + "' from target type is missing in source type!", targetType);
            }
```

beikov (changes requested):
Almost, just one more little change and this should be good to go in :)

beikov:
Merged via 7384ad172189ca4ce2de6cdf51d473b2e1ea0f86

beikov:
Thanks a lot @danelowe! Awesome work :)

danelowe:
Thanks for the help @beikov

PR #1231, opened 2020-12-09
Test Case for Issue # 1228

### LINKED ISSUE #1228: Tuple query with single scalar select returns same value for each row
Trying to use BlazeCriteriaQuery<Tuple> to return a single attribute, returns expected number of records, however, each record in the list has the same value.
Ex. Person entity with three attributes firstname, lastname, email. Table contains 10 records where 8 have distinct firstnames. Selecting only distinct firstname returns 8 records, but each has same value.

Works fine when Tuple has more than one attributes. 

### Expected behavior
Distinct values for firstname.

### Actual behavior
Same values for firstname in all 8 records.

### Steps to reproduce

```
public class Host {
	@Id
	private Long hostId;
	private String hostName;
	private Long appId;
}

@Override
@Transactional
public List<Map<String, String>> getHosts(String hostName) {
	BlazeCriteriaBuilder cb = BlazeCriteria.get(cbf);
	BlazeCriteriaQuery<Tuple> query = cb.createTupleQuery();
	BlazeRoot<?> root = query.from(Host.class);
	query.select(root.get("hostName")).distinct(true);
	CriteriaBuilder<Tuple> criteriaBuilder = query.createCriteriaBuilder(em);
	List<Tuple> result = criteriaBuilder.getResultList();
	long count = criteriaBuilder.getCountQuery().getSingleResult();
	List<Map<String, String>> hostResults = new ArrayList<Map<String,String>>();
	result.stream().forEach(tuple -> {
		Map<String, String> test = new HashMap<>();
		test.put((String)tuple.get(0), (String)tuple.get(0));
		hostResults.add(test);
	});
	//List has maps with same keys even if the underlying has different values
	return hostResults;
}
```

### Environment
Version:            1.6.0-SNAPSHOT
JPA-Provider:       Hibernate 5.4.l5.Final
DBMS:               Oracle 12c XE
Application Server: Tomcat 9.0.35

riteshka:
## Related Issue

Fixes #1228

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ Not yet released
 * Avoid transforming `SIZE` to `COUNT` when implicit group by would fail
 * Fix for implicit join in lateral join correlation path array expression index
 * Fix issues with flushing of embeddables/flat views for entity views retrieved by `EntityViewManager.getReference()`
+* Fix issues with tuple query with single scalar select returning same value for each row.
 
 ### Backwards-incompatible changes
 
--- a/core/impl/src/main/java/com/blazebit/persistence/impl/query/ObjectBuilderTypedQuery.java
+++ b/core/impl/src/main/java/com/blazebit/persistence/impl/query/ObjectBuilderTypedQuery.java
@@ -60,15 +60,13 @@ public class ObjectBuilderTypedQuery<X> extends TypedQueryWrapper<X> {
         int size = list.size();
         List<X> newList = new ArrayList<X>(size);
 
-        Object[] singleObjectTuple = new Object[1];
         for (int i = 0; i < size; i++) {
             Object tuple = list.get(i);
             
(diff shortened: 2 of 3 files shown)


beikov:
Could you please change the commit messages to align with the commit message pattern we have i.e. use the following:

* [#1228] Test case showing tuple query with single scalar select returns same value for each row
* [#1228] Fix issues with tuple query with single scalar select returning same value for each row

Also, please add the a line in the CHANGELOG.md file `* Fix issues with tuple query with single scalar select returning same value for each row`.

riteshka:
Recommended changes to commit messages and changelog.md submitted.

beikov:
Merged via f12818ea458c70d7e82f641791c9087e9cb75e88

beikov:
Thanks for the PR!

riteshka:
Thanks for merging. Appreciate your help.

PR #1287, opened 2021-04-13
Explain the additional usage scenarios of PrePersist

EugenMayer:
Also show the signature of prePersist hooks when all parameters all present


beikov:
I'll close this for now. If you want to provide some deeper explanation for the feature, that is fine for me, but this particular PR IMO will cause more confusion than help the understanding.

PR #1293, opened 2021-04-19
[#1285] Fix aliasing issues with joinOnSubquery variants on databases that don't support column aliasing at the table alias site

### LINKED ISSUE #1285: Investigate possible issue with innerJoinOnSubquery
### Description

According to https://stackoverflow.com/questions/66744651/blaze-persistence-join-on-subquery-issue there seems to be an issue with the SQL, but I'm not sure if this is due to a Hibernate issue or Blaze-Persistence. AFAICT, the aliasing doesn't happen which leads to column reference not found issues.

```
SELECT s0_.id 
FROM b s0_ 
INNER JOIN (
	SELECT 
		t0_.account_id AS col_0_0_, 
		t0_.id AS col_1_0_ 
	FROM   a t0_ 
	WHERE  t0_.account_id = 100
) t1_ -- <-- Missing aliasing. Aliasing should happen in select clause because this is MySQL
           ON ( ( NULL IS NULL ) 
                AND s0_.account_id = t1_.account_id)
WHERE  s0_.account_id = 100 
```

### Environment

Version:            1.5.1
JPA-Provider:       Hibernate 5.4
DBMS:               -
Application Server: -

beikov:
## Related Issue

Fixes #1285

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ None yet
 * Retain entity view kind when converting entity view if possible
 * Improve dialect detection for MariaDB dialects
 * Workaround Hibernate proxy field access bug for non-pk relations
+* Fix aliasing issues with `joinOnSubquery` variants on databases that don't support column aliasing at the table alias site
 
 ### Backwards-incompatible changes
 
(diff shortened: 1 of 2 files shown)


beikov:
Superseded by https://github.com/Blazebit/blaze-persistence/pull/1309

PR #1315, opened 2021-06-09
QueryDSL 5 Support

jwgmeligmeyling:

### DIFF
--- a/integration/querydsl/expressions/src/main/java/com/blazebit/persistence/querydsl/AbstractBlazeJPAQuery.java
+++ b/integration/querydsl/expressions/src/main/java/com/blazebit/persistence/querydsl/AbstractBlazeJPAQuery.java
@@ -54,6 +54,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.logging.Logger;
 
+import static com.blazebit.persistence.querydsl.HintsAccessor.getHints;
+
 /**
  * Abstract base class for JPA API based implementations of the JPQLQuery interface
  *
@@ -78,7 +80,7 @@ public abstract class AbstractBlazeJPAQuery<T, Q extends AbstractBlazeJPAQuery<T
 
     protected final CriteriaBuilderFactory criteriaBuilderFactory;
 
-    protected boolean cachable = false;
+    protected boolean cacheable = false;
 
     protected final Binds<T> binds = new Binds<>();
 
@@ -179,7 +181,7 @@ public abstract class AbstractBlazeJPAQuery<T, Q extends AbstractBlazeJPAQuery<T
             query.setFlushMode(flushMode);
         }
 
-        for (Map.Entry<String, Object> entry : hints.entries()) {
+        for (Map.Entry<String, Object> entry : getHints(this)) {
             try {
                 query.setHint(entry.getKey(), entry.getValue());
             } catch (UnsupportedOperationException e) {
@@ -187,7 +189,6 @@ public abstract class AbstractBlazeJPAQuery<T, Q extends AbstractBlazeJPAQuery<T
             }
         }
 
-        logQuery(queryable.getQueryString(), null);
         return query;
     }
 
@@ -231,11 +232,11 @@ public abstract class AbstractBlazeJPAQuery<T, Q extends AbstractBlazeJPAQuery<T
             }
         }
 
-        for (Map.Entry<String, Object> entry : hints.entries()) {
+        for (Map.Entry<String, Object> entry : getHints(this)) {
             criteriaBuilder.setProperty(entry.getKey(), entry.getValue().toString());
         }
 
-        if (cachable) {
+        if (cacheable) {
             criteriaBuilder.setCacheable(true);
         }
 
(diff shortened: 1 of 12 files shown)


integration/querydsl/expressions/src/main/java/com/blazebit/persistence/querydsl/AbstractBlazeJPAQuery.java
```diff
@@ -78,7 +79,7 @@
 
     protected final CriteriaBuilderFactory criteriaBuilderFactory;
 
-    protected boolean cachable = false;
+    protected boolean cacheable = false;
```
jwgmeligmeyling, line 82: Fix typo, unlikely to affect users because accessors where typed correctly and users ought not to subclass `BlazeJPAQuery`.

integration/querydsl/expressions/src/main/java/com/blazebit/persistence/querydsl/JPQLNextTemplates.java
```diff
@@ -147,7 +147,9 @@ public JPQLNextTemplates(char escape, QueryHandler queryHandler) {
 
 
         add(JPQLNextOps.LEAST, "LEAST({0}, {1})");
+        add(Ops.MathOps.MIN, "LEAST({0}, {1})");
```
jwgmeligmeyling, line 150: `MathOps.MIN` is `MIN(a,b)` which in Blaze-Persistence is `LEAST`. This makes sure that QueryDSL's `MathExpressions.max(a, b)` uses the right dialect specific translation.

However, we still need `JPQLNextOps.LEAST`, because the aforementioned method only accepts numbers where `LEAST` can actually compare any comparable basic value.

integration/querydsl/expressions/src/main/java/com/blazebit/persistence/querydsl/AbstractBlazeJPAQuery.java
```diff
@@ -231,22 +231,35 @@ public String getQueryString() {
...
+        if (cacheable) {
             criteriaBuilder.setCacheable(true);
         }
 
         return queryable;
     }
 
+    private Collection<Map.Entry<String, Object>> getHintsInternal() {
+        try {
+            Object hints = AbstractJPAQuery.class.getDeclaredField("hints").get(this);
```
beikov, line 247: Can we cache this in a static initializer somehow? Also, would it be possible to make the guava dependency optional?

integration/querydsl/expressions/src/main/java/com/blazebit/persistence/querydsl/WindowFunction.java
```diff
@@ -72,7 +70,7 @@ public WindowFunction(Expression<A> expr, String baseWindowName) {
     public SimpleExpression<A> getValue() {
         SimpleExpression<A> value = this.value;
         if (value == null) {
-            this.value = value = Expressions.template(target.getType(), "{0} over ({1})", (List<?>) ImmutableList.of(target, super.getValue()));
+            this.value = value = Expressions.template(target.getType(), "{0} over ({1})", target, super.getValue());
```
jwgmeligmeyling, line 73: Better to use the overload that accepts varargs now that Guava is removed.

integration/querydsl/expressions/src/main/java/com/blazebit/persistence/querydsl/AbstractBlazeJPAQuery.java
jwgmeligmeyling: Yes, I was thinking about that as well. Can we just store `AbstractJPAQuery.class.getDeclaredField("hints")` in a static variable? Or should we introduce a `HintsAccessor` class?

beikov: Moving the logic to two dedicated classes allows to use a maven optional dependency on guava which is probably nicer, but I'll leave it up to you.

jwgmeligmeyling: As far as Guava: `blaze-persistence-integration-querydsl-expressions` never had an explicit dependency on Guava. So if QueryDSL is replaced with a version that does not ship Guava, you won't have it. I guess that will cause issues with importing it here as well. That can be easily avoided if we use reflection to call `entries()` as well.

beikov: Right, but if we add an explicit optional dependency to the pom, this should work fine with both versions.

jwgmeligmeyling: Please let me now if I moved into the right direction 😄

beikov (approved):
Looks good, but maybe it's worth adding some information about the support of the new QueryDSL version to the documentation or readme?

jwgmeligmeyling:
Good point, I've updated the documentation.

PR #1320, opened 2021-07-08
Fix typo when binding the ResultCte class

vladmihalcea:
Instead of:

```
CriteriaBuilder<Long> cb = cbf.create(em, Long.class)
    .fromSubquery(ResultCte.class, "r")
        .from(Cat.class, "subCat")
        .bind("id").select("id")
        .bind("id").select("name")
        .orderByDesc("age")
        .orderByDesc("id")
        .setMaxResults(5)
    .end()
    .select("r.id");
```

We should use:

```
CriteriaBuilder<Long> cb = cbf.create(em, Long.class)
    .fromSubquery(ResultCte.class, "r")
        .from(Cat.class, "subCat")
        .bind("id").select("id")
        .bind("name").select("name")
        .orderByDesc("age")
        .orderByDesc("id")
        .setMaxResults(5)
    .end()
    .select("r.id");
```

Because the `name` property of the `ResultCte` projection should be bound to the `name` property of the `Cat` entity.


beikov:
Thanks!

vladmihalcea:
You're welcome.

Issue #1325, opened 2021-07-26
Upgrade graphql integration to graphql-java 16.0


EugenMayer:
The current integration is based on the outdated graphql-java 5.x version which is no longer binary compatible with graphql-java 16.0 and thus blaze-persistence could not be used with netflix-dgs in others.

Port the current implementation to 16.0

PR #1326, opened 2021-07-26
[#1325] Port graphql integration to graphql-java 16.0

EugenMayer:
## Description
Port integration/graphql to graphql-java 16.0

- remove dependency to graphql-java:graphql-java-tools and base on 
  graphql-java:graphql-java directly
- Fix bindings against graphql-java:graphql-java - no features or fixes intended
- Implement support for Embedded IDs within EntityViews

## Related Issue
- #1325

## Motivation and Context
Path grounding for integration with `netflix-dgs` and other `graphql-java` libraries based on the newer version which is not binary compatible

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -65,10 +65,15 @@ jobs:
             jdk: 8
             spring-data: spring-data-2.3.x
             deltaspike: deltaspike-1.9
+          - rdbms: h2
+            provider: hibernate-5.2
+            jdk: 8
+            spring-data: spring-data-2.4.x
+            deltaspike: deltaspike-1.9
           - rdbms: postgresql
             provider: hibernate-apt
             jdk: 8
-            spring-data: spring-data-2.4.x
+            spring-data: spring-data-2.5.x
             deltaspike: deltaspike-1.9
             
 ################################################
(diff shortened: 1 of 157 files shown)


EugenMayer:
@beikov I would say this one is good enough for a first review. I should cover a huge range of cases with my application right now. We have a really big spec now, with everything:
 - Deep EntityView into EntityView mapping
 - Collections nesting collections
 - Tons of Enums and Scalars of different types including `DateTime`, `LocalDataTime`, `Long`
 - All sorts of ViewFilters
 - MultiFetches, InputTypes
 - Mixing custom `Types` with the `Types` coming from this integration (EV types and other types)
 - Pagination 
 
The things I would love to discuss improving (which are beyond the port):

a)extract the actual type registry generation from `com.blazebit.persistence.integration.graphql.GraphQLEntityViewSupportFactory#create` to a custom method and use `create` to just initialize `GraphQLEntityViewSupport`. The current is a mix of concerns and this makes it unnecessarily harder to expose `GraphQLEntityViewSupport` as a bean while letting the graphql runtime wiring know about the types. Splitting those 2 would make this a lot easier and cleaner

b) Current there is no support for nullable and not-nullable values. This means that all types have the mandatory ID! and all other fields are optional. We could fetch the 'nullable' status from the entity field definition and include this. This would help a lot with the `from schema to client code` generators coming with `Appollo` or `urql` and others. Especially when using with `TypeScript` the missing information of nullable hurts a lot.
Besides that, this is an important information anyway. We could discuss this in an additional topic of course

EugenMayer:
Just have seen that i accidently commited all the pom.xml files .. trying to get rid of id

beikov:
Superseded by https://github.com/Blazebit/blaze-persistence/pull/1343

Thanks a lot for the initial implementation. After some time working with this stuff, I realized only a few reflection accesses are necessary to support 9.2 - 16.2 so I added that for now. We can rethink this again in case graphql-java matures or breaks the API again.

PR #1330, opened 2021-08-02
[#1072] Introduce CriteriaBuilderConfigurationContributor

### LINKED ISSUE #1072: Function contributor Service Provider interface
### Description
Currently we register functions in `CriteriaBuilderConfigurationImpl`. Users are allowed to extend these in their own `CriteriaConfiguration`. However, it is not possible for extensions or submodules to register functions dynamically. Maintaining various `CriteriaConfigurations` in large projects can become challenging.

Therefore it would be nice to introduce a ServiceProvider interface that is allowed to register functions during the bootstrap of the `CriteriaBuilderConfigurationImpl`.

The feature was already discussed on slack. This feature needs a `ServiceProvider` interface, ServiceProvider loading logic and some documentation on how to use it.

### Environment

Version:            
JPA-Provider:       
DBMS:               
Application Server:

jwgmeligmeyling:
## Description
Allow third-party integrations to extend the CriteriaBuilderConfiguration through a Service Provider interface.

## Related Issue
Fixes #1072

## Motivation and Context
The https://github.com/jwgmeligmeyling/hibernate-types-querydsl-apt project provides an extension for working with custom types in special dialects. For this purpose, some additional functions need to be registered with Hibernate, however, Blaze-Persistence needs to be aware which functions are aggregate functions (for implicit group by) and which ones are not. Therefore the functions need to be registered as Blaze-Persistence functions as well. A service provider interface allows third party libraries like mine to provide the required configuration within the library itself.

### DIFF
--- a/core/api/src/main/java/com/blazebit/persistence/spi/CriteriaBuilderConfigurationContributor.java
+++ b/core/api/src/main/java/com/blazebit/persistence/spi/CriteriaBuilderConfigurationContributor.java
@@ -18,6 +18,10 @@ package com.blazebit.persistence.spi;
 
 /**
  * A bootstrap process hook for contributing settings to the {@link CriteriaBuilderConfiguration}.
+ * {@code CriteriaBuilderConfigurationContributor} instances may be annotated with {@link Priority}
+ * (or {@code javax.annotation.Priority}) to influence the order in which they are registered.
+ * The range 0-500 is reserved for internal uses. 500 - 1000 is reserved for libraries and 1000+
+ * is for user provided contributors.
  *
  * Implementations are instantiated via {@link java.util.ServiceLoader}.
  *
--- a/core/api/src/main/java/com/blazebit/persistence/spi/CriteriaBuilderConfigurationContributor.java
+++ b/core/api/src/main/java/com/blazebit/persistence/spi/CriteriaBuilderConfigurationContributor.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2014 - 2021 Blazebit.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.blazebit.persistence.spi;
+
+/**
+ * A bootstrap process hook for contributing settings to the {@link CriteriaBuilderConfiguration}.
+ *
+ * Implementations are instantiated via {@link java.util.ServiceLoader}.
+ *
+ * @author Jan-Willem Gmelig Meyling
+ * @since 1.6.2
+ */
+public interface CriteriaBuilderConfigurationContributor {
+
+    /**
+     * Perform the process of contributing to the {@link CriteriaBuilderConfiguration}.
+     *
+     * @param criteriaBuilderConfiguration the {@link CriteriaBuilderConfiguration} to which to contribute
+     */
+    void contribute(CriteriaBuilderConfiguration criteriaBuilderConfiguration);
+
+}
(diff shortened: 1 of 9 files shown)


documentation/src/main/asciidoc/core/manual/en_US/configuration.adoc
```diff
@@ -250,4 +250,9 @@ For more information refer to the <<customize-dbms-dialect,Customize DBMS dialec
...
 === Entity manager factory integrators
 
-A little integration layer to extract the dbms of an entity manager and register `JpqlFunctionGroup` instances with the entity manager. This is normally provided by an integration module for persistence providers.
\ No newline at end of file
+A little integration layer to extract the dbms of an entity manager and register `JpqlFunctionGroup` instances with the entity manager. This is normally provided by an integration module for persistence providers.
+
+[[service-loader-configuration-integrators]]
+=== Extending configuration from integrations
+
+Integrations may provide additional configuration to the bootstrapping `CriteriaBuilderConfiguration` through a `CriteriaBuilderConfigurationContributor` Service Provider.
```
beikov, line 258: It would be great if you could list some concrete use cases like e.g. contributing functions etc.

core/api/src/main/java/com/blazebit/persistence/spi/CriteriaBuilderConfigurationContributor.java
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * A bootstrap process hook for contributing settings to the {@link CriteriaBuilderConfiguration}.
+ *
+ * Implementations are instantiated via {@link java.util.ServiceLoader}.
+ *
+ * @author Jan-Willem Gmelig Meyling
+ * @since 1.6.2
+ */
+public interface CriteriaBuilderConfigurationContributor {
```
beikov, line 27: IMO this should also include a `int ordinal()` method with a description like `The ordinal is used to determine ordering of the contributors. The range 0-500 is reserved for internal uses. 500 - 1000 is reserved for libraries and 1000+ is for user provided contributors.`

core/impl/src/main/java/com/blazebit/persistence/impl/CriteriaBuilderConfigurationImpl.java
```diff
@@ -1873,6 +1875,12 @@ private void loadEntityManagerIntegrator() {
         }
     }
 
+    private void loadExtensions() {
+        for (CriteriaBuilderConfigurationContributor contributor : ServiceLoader.load(CriteriaBuilderConfigurationContributor.class)) {
```
beikov, line 1879: Sort by `ordinal` and then by fully qualified class name to achieve a deterministic ordering.

beikov (changes requested):

core/api/src/main/java/com/blazebit/persistence/spi/CriteriaBuilderConfigurationContributor.java
jwgmeligmeyling: What about a `Priority` annotation (like JAX-RS filters). I do understand the wish for deterministic ordering on one hand, and control over the ordering on the other hand, but I am not sure whether I like it to bleed into the SPI itself.

beikov: Hmm, I guess that would be fine as long as we only have an optional dependency on the javax.annotation-api artifact.

jwgmeligmeyling: I was thinking to introduce our own annotation, however, I can also use the one from` javax.annotation-api` and its not unreasonable for users to expect that to just work.

I would need to extract the value through reflection though.

jwgmeligmeyling: A pro for an internal annotation would be that if we actually were to use the SPI ourselves, we would endup with a compile time dependency on `javax.annotation-api` anyway.

beikov: Hmm, I guess a custom annotation in core-api is probably a better choice indeed. It would be good to still also support `@Priority` from javax.annotation-api though.

documentation/src/main/asciidoc/core/manual/en_US/configuration.adoc
jwgmeligmeyling: Added more documentation.

core/api/src/main/java/com/blazebit/persistence/spi/CriteriaBuilderConfigurationContributor.java
jwgmeligmeyling: See `CriteriaBuilderConfigurationContributorComparator`.

beikov (approved):
Awesome, looks great now!

beikov (approved):

Issue #254, opened 2016-09-29
Implement support for chunk-processing


beikov:
See #253 for details about the discussion.
Basically what we need is an API that allows chunk-wise processing similar to what Hibernate offers with it's scrollable result.

JWGmeligMeyling:
JPA 2.2 will bring `Query#getStream()`, that defaults to `Query#getResultList().stream()` but which ORM vendors may override with a more performant implementation. Hibernate 5 already offers a `Query#stream()` method but barely optimises this. Optimisation of `Query#stream()` (and the future `Query#getStream()`) is planned for Hibernate 6.0. 

Simply put, I think that if you want to use this today, you have to use Hibernate 6.0 and call `getQuery().getStream()` on your criteria builder. Even then, the actual optimisation may not yet be fully implemented (I am uncertain what the current state of development of HH6 is).

beikov:
Hibernate 6 is currently expected to be released by the end of this year. I'd like to avoid the requirement of using streams, after all, it is based on an Iterator/Spliterator and the scrollable API of Hibernate exposes this, so there is no need to force the Stream API on the user. By simply implementing proper iterators, we can support both, normal Iterable usage and Stream usage.
Implementing optimizations like the ones the Hibernate team is speculating about is out of scope for this issue.

soberich:
@beikov @JWGmeligMeyling
What do you think about this streaming? Say, we have tens of thousands rows to stream.
Any severe drawbacks/disadvantages? (regardless transaction and considering we have lazy fetching where possible)
```kotlin
    @Transactional(NOT_SUPPORTED)
    override fun streamEmployees(): Stream<Employee> = 
            em.createQuery("SELECT e FROM EmployeeEntity e", EmployeeEntity::class.java)
                    .resultStream.map { evm.convert(it, EmployeeListingView::class.java) }
```

beikov:
Fixed via https://github.com/Blazebit/blaze-persistence/pull/1344

PR #1344, opened 2021-08-30
[#1317] Delegate getResultStream method in Query interfaces

### LINKED ISSUE #1317: Efficient Entity View Result Streaming
JPA 2.2 added the method `getResultStream()` to the `TypedQuery` interface with a default implementation that first loads all the results into memory and then returns a stream.

Blaze Persistence has multiple classes which implement `TypedQuery`, for example `ObjectBuilderTypedQuery` and `TypedQueryWrapper`.

I would suggest adding implementations of `getResultStream()` to those classes so that they do not use the inefficient default implementation.

TypedQueryWrapper.java:
```java
//...
    @Override
    public Stream<X> getResultStream() {
        return delegate.getResultStream();
    }
//...
```

ObjectBuilderTypedQuery.java:
```java
//...
    @Override
    public Stream<X> getResultStream() {
        return super.getResultStream().map(tuple -> {
            if (tuple instanceof Object[]) {
                return builder.build((Object[]) tuple);
            } else {
                return builder.build(new Object[] { tuple });
            }
        });
    }
//...
```

These changes allow entity view results to be streamed without first being fully loaded into memory if the underlying JPA provider supports it. I've tested it to work with Hibernate 5.4.

jwgmeligmeyling:
## Related Issue
Fixes #1317
Perhaps #254 as well


jwgmeligmeyling:
@beikov do you want to add the method to `Fetchable` as well?

integration/hibernate-6.0/src/main/java/com/blazebit/persistence/integration/hibernate/Hibernate60Access.java
```diff
@@ -214,4 +232,33 @@ public ParameterTranslations createParameterTranslations(List<ParameterSpecifica
...
+    private static class ScrollableResultsIterator implements Iterator<Object> {
+        private final ScrollableResultsImplementor scrollableResults;
+
+        public ScrollableResultsIterator(ScrollableResultsImplementor scrollableResults) {
+            this.scrollableResults = scrollableResults;
+        }
+
+        @Override
+        public boolean hasNext() {
+            return !scrollableResults.isClosed() && scrollableResults.next();
```
beikov, line 250: AFAIU `next()` will advance the result, so I think it is more appropriate to check `!isLast()` here and in all other implementations

beikov: On further thought, it might be better to eagerly call `next()` in the constructor and check `scrollableResults.get() != null`

integration/hibernate-5/src/main/java/com/blazebit/persistence/integration/hibernate/Hibernate5Access.java
```diff
@@ -320,4 +337,38 @@ public ParameterTranslations createParameterTranslations(List<ParameterSpecifica
...
+            Object[] next = scrollableResults.get();
+            if (next.length == 1) {
+                return next[0];
+            } else {
+                return next;
+            }
+        }
+
+        @Override
+        public void remove() {
```
beikov, line 369: Remove?

integration/hibernate-4.3/src/main/java/com/blazebit/persistence/integration/hibernate/Hibernate43Access.java
```diff
@@ -315,4 +332,40 @@ public QueryParameters createQueryParameters(
...
+            Object[] next = scrollableResults.get();
+            if (next.length == 1) {
+                return next[0];
+            } else {
+                return next;
+            }
+        }
+
+        @Override
+        public void remove() {
```
beikov, line 365: Remove?

integration/hibernate-4.3/src/main/java/com/blazebit/persistence/integration/hibernate/Hibernate43Access.java
```diff
@@ -315,4 +332,40 @@ public QueryParameters createQueryParameters(
...
+    private static class ScrollableResultsIterator implements Iterator<Object> {
+        private final ScrollableResults scrollableResults;
+
+        public ScrollableResultsIterator(ScrollableResults scrollableResults) {
+            this.scrollableResults = scrollableResults;
+        }
+
+        @Override
+        public boolean hasNext() {
+            return scrollableResults.next();
```
beikov, line 351: Implementation differs from other implementations

integration/hibernate-4.2/src/main/java/com/blazebit/persistence/integration/hibernate/Hibernate4Access.java
```diff
@@ -314,4 +331,40 @@ public QueryParameters createQueryParameters(
...
+            Object[] next = scrollableResults.get();
+            if (next.length == 1) {
+                return next[0];
+            } else {
+                return next;
+            }
+        }
+
+        @Override
+        public void remove() {
```
beikov, line 364: Remove?

integration/hibernate-4.2/src/main/java/com/blazebit/persistence/integration/hibernate/Hibernate4Access.java
```diff
@@ -314,4 +331,40 @@ public QueryParameters createQueryParameters(
...
+    private static class ScrollableResultsIterator implements Iterator<Object> {
+        private final ScrollableResults scrollableResults;
+
+        public ScrollableResultsIterator(ScrollableResults scrollableResults) {
+            this.scrollableResults = scrollableResults;
+        }
+
+        @Override
+        public boolean hasNext() {
+            return scrollableResults.next();
```
beikov, line 350: Again, implementation differs from the other implementations.

integration/hibernate-6.0/src/main/java/com/blazebit/persistence/integration/hibernate/Hibernate60Access.java
jwgmeligmeyling: I have some questions of my own about the desiredness of a check to `isClosed` here. I think invoking `next` on a (likely unexpectedly) closed scrollable result should throw an exception for sure. However, I just copied the implementation from https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/main/java/org/hibernate/query/internal/ScrollableResultsIterator.java#L19-L47 as I assumed there were good reasons to do this (in fact, the entire `ScrollableResultsImplementor` interface appears to be introduced merely for this purpose...)

See: https://github.com/hibernate/hibernate-orm/commit/7cae5ba95bf355d1023c25820e45eaf5f984bcb7
And: https://hibernate.atlassian.net/browse/HHH-9340

core/impl/pom.xml
```diff
@@ -50,11 +50,11 @@
             <artifactId>blaze-common-utils</artifactId>
         </dependency>
 
-        <!-- Actually we only need JPA but since there is no official artifact 
-            we rather use this than provider specific apis -->
+        <!-- JPA 2.2 API for getResultStream signature -->
         <dependency>
-            <groupId>org.apache.openejb</groupId>
-            <artifactId>javaee-api</artifactId>
+            <groupId>javax.persistence</groupId>
```
beikov, line 55: Let's use the following dependency instead and please declare it in the parent/pom.xml like the `javaee-api` one so that we can omit the version here:
```
<dependency>
  <groupId>jakarta.persistence</groupId>
  <artifactId>jakarta.persistence-api</artifactId>
  <version>2.2.3</version>
</dependency>
```

integration/hibernate-4.3/src/main/java/com/blazebit/persistence/integration/hibernate/Hibernate43Access.java
jwgmeligmeyling: Yes, because the call "needed" (I am questioning that in my other comment) was only introduced after 5.2.

beikov:
> @beikov do you want to add the method to `Queryable` as well?

Good question, I think we should, yes. Do you happen to have Java 7 installed to check if this all still runs on Java 7?

integration/hibernate-5/src/main/java/com/blazebit/persistence/integration/hibernate/Hibernate5Access.java
jwgmeligmeyling: Remove was still mandatory in 1.7, and Hibernate 5 doesn't target 1.8.

integration/hibernate-4.3/src/main/java/com/blazebit/persistence/integration/hibernate/Hibernate43Access.java
jwgmeligmeyling: Remove was still mandatory in 1.7, and Hibernate 5 doesn't target 1.8.

integration/hibernate-4.2/src/main/java/com/blazebit/persistence/integration/hibernate/Hibernate4Access.java
jwgmeligmeyling: Yes, because the call "needed" (I am questioning that in my other comment) was only introduced after 5.2.

integration/hibernate-4.2/src/main/java/com/blazebit/persistence/integration/hibernate/Hibernate4Access.java
jwgmeligmeyling: Remove was still mandatory in 1.7, and Hibernate 5 doesn't target 1.8.

jwgmeligmeyling:
> Do you happen to have Java 7 installed to check if this all still runs on Java 7?

I do have Java 7, but do we have an example project that compiles against 7?

beikov:
> > Do you happen to have Java 7 installed to check if this all still runs on Java 7?
> 
> I do have Java 7, but do we have an example project that compiles against 7?

Quickstart? 😆

jwgmeligmeyling:
Let's give it a shot

core/api/src/main/java/com/blazebit/persistence/Queryable.java
```diff
@@ -58,4 +59,11 @@
      */
     public T getSingleResult();
 
+    /**
+     * Execute the query and return the result as a type Stream.
+     *
+     * @return The stream of the results
+     */
+    public Stream<T> getResultStream();
```
jwgmeligmeyling, line 67: Quickstart still works under Java 7, although this change leads to a `ClassNotFoundException` when doing reflection calls on the Criteria Builder API, i.e.:

```java
for (Method method : cb.getClass().getMethods()) {
    System.out.println(method);
}
```

beikov:
Nice work, thanks!

PR #1353, opened 2021-09-11
[#1352, #1349, #1350] Infer nullability from entity view mappings for attributes in GraphQL schema

### LINKED ISSUE #1349: With 1.6.2 all fields are mandatory in GraphQL
While in https://github.com/Blazebit/blaze-persistence/pull/1326 every field has been optional (except ID, while this was also not entirely true, see https://github.com/Blazebit/blaze-persistence/pull/1326#issuecomment-889242680), with the 1.6.2 rewrite of that branch, we end up having every single field as mandatory, except collections.

This is fairly critical, since it makes the `graphqls` definition a lot more strict and clients fail, including client generation and types are wrong due to that issue.

IMHO would suggest to derive the mandatory/optional status from the hibernate model, if possible. If that is yet too much work, I would suggest to go back to all optional, instead of all mandatory (the former is the better worse).

### LINKED ISSUE #1350: JPA Criteria joins over embeddables do not render properly
Hi @beikov ,

Follow up from https://github.com/Blazebit/blaze-persistence/issues/1333 and https://github.com/Blazebit/blaze-persistence/issues/1348#issuecomment-913167000 where you said:

> Looks like a bug in Spring Data 2.5 ;)
> You can make use of a specification to model this though or if you want to use an entity view, use an attribute filter:
> ```
> @EntityView(BundleRequestEntity.class)
> public interface BundleRequestSimpleView {
>     @AttributeFilter(EqualFilter.class)
>     @Mapping("bundleRequestParameters.tokenRange.keyAlias")
>     String getKeyAlias();
> }
> @Repository
> public interface RequestRepository extends JpaRepository<BundleRequestEntity, UUID> {
>     default List<BundleRequestSimpleView> findByBundleRequestParametersTokenRangeKeyAlias(String keyAlias) {
>         return findAll(setting -> setting.withAttributeFilter("keyAlias", keyAlias));
>     }
>     List<BundleRequestSimpleView> findAll(EntityViewSettingProcessor<BundleRequestSimpleView> processor);
> }
> ```
Please find here attached a reproducer. [demo.zip](https://github.com/Blazebit/blaze-persistence/files/7116085/demo.zip)

Run this application and target http://localhost:8080/demo
No exception.

Run the application with  'blaze' as profile and target http://localhost:8080/demo
then you will see the exception

The 'blaze' profile triggers the BP configuration (see `DemoApplication`)

Let me know if I'm missing something here ;-)

Thanks!

### LINKED ISSUE #1352: Infer nullability of GraphQL fields from database model
It would be awesome if nullability could be propagated to the GraphQL fields.

beikov:
## Related Issue

Fixes #1352. Fixes #1349. Fixes #1350

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,11 +8,12 @@ Not yet released
 
 ### New features
 
-None yet
+* Introduce type filter pattern for `GraphQLEntityViewSupportFactory`
+* Infer nullability in the GraphQL schema from the entity view mapping expressions
 
 ### Bug fixes
 
-None yet
+* Fix issues with wrongly marking all attributes as non-null in the GraphQL schema
 
 ### Backwards-incompatible changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Not yet released
 ### Bug fixes
 
 * Fix issues with wrongly marking all attributes as non-null in the GraphQL schema
+* Fix JPA Criteria joins rendering over embeddables
 
 ### Backwards-incompatible changes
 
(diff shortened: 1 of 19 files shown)


jwgmeligmeyling:
This is pretty awesome 😃

PR #1415, opened 2022-01-01
[#1414] Upgrade Quarkus to 2.6.1.Final and fix metadata of extension

### LINKED ISSUE #1414: Upgrade Blaze Persistence to Quarkus 2.6.1.Final and generate proper metadata
As seen in https://github.com/quarkusio/quarkus-platform/pull/417, the Blaze Persistence Quarkus extension does not provide all the necessary metadata for modern Quarkus.

Quarkus 2 has been around for a long while now so let's make the extension more modern.

I'm going to work on this.

gsmet:
Fixes #1414

Note: there's a good chance CI will fail as the license headers need an update but I'll let you do that yourself as the license update also removed a new line on my box and I'm not sure it's expected.

Let me know when I can rebase this one.


integration/quarkus/deployment/src/main/java/com/blazebit/persistence/integration/quarkus/deployment/BlazePersistenceProcessor.java
```diff
@@ -76,17 +73,6 @@
...
-    @BuildStep
-    CapabilityBuildItem capability() {
-        String version = Version.getVersion();
-        int dotIndex = version.indexOf('.');
-        // As of version Quarkus 2, the capabilities are read from the extension descriptor
-        if (dotIndex == -1 || Integer.parseInt(version.substring(0, dotIndex)) >= 2) {
-            return null;
-        }
-        return new CapabilityBuildItem(CAPABILITY);
-    }
```
gsmet, line 88: I don't think it's worth keeping that for new versions given Quarkus 2 is there for quite a while now and there's very little chance people would use new versions of the extension with it.

integration/quarkus/runtime/pom.xml
```diff
@@ -78,6 +78,25 @@
             </resource>
         </resources>
         <plugins>
+            <plugin>
+                <groupId>io.quarkus</groupId>
+                <artifactId>quarkus-bootstrap-maven-plugin</artifactId>
```
gsmet, line 83: @aloubyansky can you check I did things properly? I checked the metadata and they look fine now.

aloubyansky:
Looks good @gsmet, thanks!

famod:
Coming here from https://github.com/quarkiverse/quarkus-logging-manager/issues/125

First load of Dev UI breaks with blaze 1.6.4 on Quarkus 2.6.1. _Without_ that lazy optimization from (which I first thought was the cause) the error pops up eagerly during boot of dev mode (which is even more unpleasant).

famod:
...aaand it breaks on Quarkus 2.5.4 as well (eagerly this time because that lazy tweak was added only to 2.6.0.CR1+).

aloubyansky:
@famod i think @gsmet's PR may fix it.

famod:
@aloubyansky yeah, thats what I reckoned.

It feels there is a test coverage gap here. I mean quarkus integration wise 1.6.4 is kinda doomed, 1.6.5 will probably fix it.
But tbh, I would expect a red light to go on before a release of an extension is published? Is this gap a general problem or just specific to blaze?

No offense really, it's just that 1.6.4 took a while and so it's a bit sad to see it turn out like this.

aloubyansky:
Extensions can include dev mode tests in their test suite (and the the platform one), I suppose that would catch this kind of issue.

gsmet:
TBH, my main interrogation is why it didn't fail earlier.

aloubyansky:
Earlier - meaning 1.6.3?

famod:
1.6.3 didn't have this: #1376 (but not sure if related)

gsmet:
Outch, yes, so that might be it because it removes the metadata generation. My guess is that the `quarkus-extension.properties` file should also be removed from the repo? WDYT @aloubyansky ?

aloubyansky:
Yes, if the plugin is present anyway, it's not worth it having the descriptor as a static resource.

gsmet:
I also updated the Quarkus version in this PR.

famod:
Right, I missed that one.
FTR: So with 2.5.4.Final that one test fails, but due to the lazy tweak in 2.6, 2.6.1.Final doesn't fail (because the test is apparently not calling dev ui).

gsmet:
```
Error:    bad class file: /home/runner/.m2/repository/io/quarkus/arc/arc/2.6.1.Final/arc-2.6.1.Final.jar(io/quarkus/arc/DefaultBean.class)
Error:      class file has wrong version 55.0, should be 52.0
```

So now we have the issue that Quarkus 2 is JDK 11 only.

PR #1416, opened 2022-01-04
[#1414] Fix Quarkus extension metadata take 2

gsmet:
https://github.com/Blazebit/blaze-persistence/pull/1415 is the way to go for the future but Blaze Persistence still supports JDK 8 for now, which makes it a bit impractical.

So for now, we just update the metadata to a format that should work with latest Quarkus.

It should be tested by going to the Dev UI of a 2.6.1.Final project and make sure the BP tile is there.

### DIFF
--- a/integration/quarkus/runtime/src/main/resources/META-INF/quarkus-extension.yaml
+++ b/integration/quarkus/runtime/src/main/resources/META-INF/quarkus-extension.yaml
@@ -1,3 +1,4 @@
+---
 name: "Blaze-Persistence"
 metadata:
   short-name: "BP"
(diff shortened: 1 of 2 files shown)


integration/quarkus/runtime/src/main/resources/META-INF/quarkus-extension.yaml
```diff
@@ -12,4 +13,22 @@ metadata:
...
+  - "io.quarkus:quarkus-agroal"
+  - "io.quarkus:quarkus-arc"
+  - "io.quarkus:quarkus-datasource"
+  - "io.quarkus:quarkus-narayana-jta"
+  - "io.quarkus:quarkus-mutiny"
+  - "io.quarkus:quarkus-smallrye-context-propagation"
+  - "io.quarkus:quarkus-caffeine"
+artifact: "${project.groupId}:${project.artifactId}::jar:${project.version}"
+group-id: "${project.groupId}"
+artifact-id: "${project.artifactId}"
```
gsmet, line 33: These two are for some old versions of Quarkus before we moved to `artifact`.

famod:
> It should be tested by going to the Dev UI of a 2.6.1.Final project and make sure the BP tile is there.

Just did that: Error page on first load of Dev UI is gone and the tile is there. 👍

beikov:
Thanks for testing @famod and thanks for the PR @gsmet

PR #1454, opened 2022-02-18
[#1453] Fix rendering parameters in subqueries in select clause

### LINKED ISSUE #1453: QueryDSL exists subquery in case when expression doesn't work
An expression like the following:

```java
Expressions.cases().when(
JPQLNextExpressions.selectFrom(table).where(predicates).exists()
).then(true).otherwise(false)
```

fails with the error

```
java.lang.IllegalArgumentException: Parameter name "param_1" does not exist
```

Where param_1 seems to be the `true` in the case expression (ParameterManager.satisfyParameter)

jwgmeligmeyling:
## Description

The `SerializerBase` which the `JPQLNextSerializer` is based on maintains a mapping of constants to unique parameter names. This ensures for each unique object (by identity) only a single parameter name is introduced. Because this mapping is shared between the main query and subqueries, a query with parameters in both a subquery and the outer query, where the subquery is rendered in the select clause, results in a raise condition where the parameters for the outer query are set when serializing the subquery.

There are a few options to fix this:

1. Only bind parameters that are managed in the `ParameterManager` for the active builder.
2. Only render parameters and constants for the main query, after all subqueries have already been rendered
3. Use a stack based constant to parameter mapping. This way we can determine which parameters should be bound at which stage of serialization. This can either be achieved through a separate separate mapping (used parameters for each subquery) or by making `constantToLabel` itself a stack. In case of the latter, we also need to ensure that generated parameter names remain unique between subqueries. I.e. by keep a count.

This PR implements approach (1), but comments are welcome 😄 

## Related Issue
Fixes #1453

### DIFF
--- a/integration/querydsl/expressions/src/main/java/com/blazebit/persistence/querydsl/BlazeCriteriaBuilderRenderer.java
+++ b/integration/querydsl/expressions/src/main/java/com/blazebit/persistence/querydsl/BlazeCriteriaBuilderRenderer.java
@@ -124,7 +124,9 @@ public class BlazeCriteriaBuilderRenderer<T> {
 
     public Queryable<T, ?> render(Expression<?> expression) {
         this.criteriaBuilder = (CriteriaBuilder) criteriaBuilderFactory.create(entityManager, Object.class);
-        return (Queryable<T, ?>) serializeSubQuery(this.criteriaBuilder, expression);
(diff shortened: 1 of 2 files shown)


beikov:
Approach 2 sounds "best" to me, but if you think that approach 1 works without problems, I'm fine with that.

PR #1461, opened 2022-02-22
[#1460] Treat modulo operation return type

### LINKED ISSUE #1460: Modulo operator return type clash between Querydsl and JPA
Querydsl assumes the LHS side as return type for the modulo operator. While this is true for various DB dialects, in JPA the spec states that modulo should return an Integer. As a result expressoins like `longExpression.mod(2L).eq(1L)` fail, because the parameter type for the constant is bound to an Integer. `intExpression.mod(2).eq(1)` on the other hand works fine.

While it is Hibernate/JPA that is strict about this and not Blaze-Persistence, users first run into this issue with the Blaze-Persistence Querydsl integration as normally Querydsl attempts to cast number parameters to the required parameter type.

```java
                if (parameterType != null && !parameterType.isInstance(val)) {
                    if (val instanceof Number && Number.class.isAssignableFrom(parameterType)) {
                        val = MathUtils.cast((Number) val, parameterType);
                    }
                }
```

Possible fixes:

1. Change `NumberExpression.mod` to return an `Integer` instead of a `<T>`. This will however be a breaking change in the Querydsl API and it might cause outrage under non-JPA users, because the Expression types are shared between `querydsl-jpa` and `querydsl-sql`, etc.
2. Introduce a utility method `JPAExpressions.mod(a, b)` that returns an `Integer` (could also be done in `JPQLNextExpressions`). This will not prevent migration issues for users that use `NumberExpression.mod`  on long fields, but at least give them a clear migration path.
3. Also cast the parameter values before setting them if these are uncompatible number types. While this would be consistent with Querydsl, I think it is undesirable.
4. Apply a treat operation to hint the query analyzer the desired return type of the operation.

Personally I like option 4 the best. No breaking changes, no unexpected conversions (you ask for a long, you get a long), and if you truly want an integer, you can still explicitly cast or use integers as input.

### Environment

Version:            
JPA-Provider:       
DBMS:               
Application Server:

jwgmeligmeyling:
Fixes #1460

### DIFF
--- a/integration/querydsl/expressions/src/main/java/com/blazebit/persistence/querydsl/JPQLNextExpressions.java
+++ b/integration/querydsl/expressions/src/main/java/com/blazebit/persistence/querydsl/JPQLNextExpressions.java
@@ -32,6 +32,7 @@ import com.querydsl.core.types.dsl.DateExpression;
 import com.querydsl.core.types.dsl.DateTimeExpression;
 import com.querydsl.core.types.dsl.Expressions;
 import com.querydsl.core.types.dsl.NumberExpression;
+import com.querydsl.core.types.dsl.SimpleOperation;
 import com.querydsl.core.types.dsl.StringExpression;
 import com.querydsl.jpa.JPAExpressions;
 import com.querydsl.jpa.JPQLOps;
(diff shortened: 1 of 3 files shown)


beikov (approved):
LGTM

PR #1489, opened 2022-05-20
Implement more annotation support for graphql type factory - fixes #1488

### LINKED ISSUE #1488: Respect common annotations on EntityViews for GraphQL type generations
We would like to improve the generated types for GraphQL so that one can use (more) of the common annotations on the `EntityViews` to let the type-generation understand, what should be done (like the annotations that are already respected to rename fields).

The aspects we are implementing are
- `javax.validation.constraints.NotNull` - respect if a field should not be nullable, even though the hibernate entity might suggest / reflect it. Just to override that.
- `com.fasterxml.jackson.annotation.JsonIgnore` - respect if a field is marked to be ignored (using thee usual jackson annotation)

EugenMayer:
Implements #1488

TODO / Questions:
  

- [x] is the `isEntityView` check required in `isNotNull` or is it not needed, since it can only be called on EntityViews (filtered before)


integration/graphql/src/main/java/com/blazebit/persistence/integration/graphql/GraphQLEntityViewSupportFactory.java
```diff
@@ -1156,7 +1160,18 @@ protected GraphQLOutputType getScalarType(GraphQLSchema.Builder schemaBuilder, C
         return new GraphQLTypeReference(typeName);
     }
 
+    /**
+     * Either determines the nullability of a field from the annotations present on the entity view or if no overrides
+     * present, derive it from the hibernate entity referenced by the view.
+     */
     protected boolean isNotNull(SingularAttribute<?, ?> attribute, EntityMetamodel entityMetamodel) {
+        boolean isEntityView = attribute.getDeclaringType().getJavaType().isAnnotationPresent(EntityView.class);
```
beikov, line 1168: This is unnecessary, because this is only called for entity view attributes. Please remove this part.

EugenMayer: done

EugenMayer:
Trying to build the project locally to understand the issues in the tests, but failing

```bash
mvn archetype:generate "-DarchetypeGroupId=com.blazebit" "-DarchetypeArtifactId=blaze-persistence-archetype-core-sample" "-DarchetypeVersion=1.6.6"
```

```
[ERROR] Failed to execute goal com.mycila:license-maven-plugin:3.0:check (check-license) on project blaze-persistence: Some files do not have the expected license header -> [Help 1]
```

Used to build the entire project using (JDK 17)

```
set jdk8.home=~/.jenv/versions/1.8
mvn install -pl '!website,!documentation' -DskipTests
```

But that fails with the very same error

beikov:
>Trying to build the project locally to understand the issues in the tests, but failing

Don't worry about that, you simply can't statically refer to the javax class as that might not be on the class path, but I'll fix that when integrating your work.

>But that fails with the very same error

The license error you mean? That shouldn't fail, unless you have an old version of the source tree or are running from a subdirectory.

EugenMayer:
> The license error you mean? That shouldn't fail, unless you have an old version of the source tree or are running from a subdirectory.

Sorry that was the case, i was not aware my repo was actually based on my fork. Merged downstream now and refreshed the PR, the local build is working now. Trying to update the spring dgs example to a more recent DGS version

EugenMayer:
> Don't worry about that, you simply can't statically refer to the javax class as that might not be on the class path, but I'll fix that when integrating your work.

do you mean that the test issues are nothing to worry about?

beikov:
This was fixed as part of https://github.com/Blazebit/blaze-persistence/commit/7be4bda106599954b61b3d171a1be042349060b8

PR #1464, opened 2022-02-22
[#1443] Fix query cloning in Querydsl

jwgmeligmeyling:
Fixes #1443

PR #1526, opened 2022-08-02
Various querydsl fixes

### LINKED ISSUE #1443: ClassCastException when cloning BlazeJPAQuery (querydsl)
### Description

When a BlazeJPAQuery is cloned by clone Method the following error occurs:
`class com.blazebit.persistence.querydsl.Binds cannot be cast to class com.querydsl.core.types.Operation`

The AbstractBlazeJPAQuery#clone Method contains the following line (at 251)
`this.binds.addBinds(((Operation) query.binds.accept(new ReplaceVisitor<Void>(), null)).getArgs());`

The cast should be to Binds not to Operation.

### Steps to reproduce

`new BlazeJPAQuery<T>(entityManager, cbf).clone();`

### Environment

Version: 1.6.6           
JPA-Provider: Hibernate 5.6.4     
DBMS: PostgreSQL 13.3              
Application Server: Spring Boot 
QueryDSL: 5.0.0

### LINKED ISSUE #1460: Modulo operator return type clash between Querydsl and JPA

### LINKED ISSUE #1516: Set operation of multiple BlazeJPAQuery fails during execution
### Description

Creating a union of multiple BlazeJPAQuery and executing that doesn't work.

### Expected behavior

Should produce a result

### Actual behavior

```
java.lang.IllegalStateException: Modifications to a query after connecting with a set operation is not allowed!
	at com.blazebit.persistence.impl.AbstractCommonQueryBuilder.prepareForModification(AbstractCommonQueryBuilder.java:3526) ~[blaze-persistence-core-impl-1.6.6.jar:1.6.6]
	at com.blazebit.persistence.impl.AbstractFullQueryBuilder.prepareForModification(AbstractFullQueryBuilder.java:134) ~[blaze-persistence-core-impl-1.6.6.jar:1.6.6]
	at com.blazebit.persistence.impl.CriteriaBuilderImpl.prepareForModification(CriteriaBuilderImpl.java:48) ~[blaze-persistence-core-impl-1.6.6.jar:1.6.6]
	at com.blazebit.persistence.impl.AbstractCommonQueryBuilder.with(AbstractCommonQueryBuilder.java:531) ~[blaze-persistence-core-impl-1.6.6.jar:1.6.6]
	at com.blazebit.persistence.impl.AbstractCommonQueryBuilder.with(AbstractCommonQueryBuilder.java:518) ~[blaze-persistence-core-impl-1.6.6.jar:1.6.6]
	at com.blazebit.persistence.impl.AbstractCommonQueryBuilder.with(AbstractCommonQueryBuilder.java:508) ~[blaze-persistence-core-impl-1.6.6.jar:1.6.6]
	at com.blazebit.persistence.querydsl.BlazeCriteriaBuilderRenderer$JPQLNextExpressionSerializer.visitOperation(BlazeCriteriaBuilderRenderer.java:1106) ~[blaze-persistence-integration-querydsl-expressions-1.6.6.jar:1.6.6]
	at com.querydsl.core.support.SerializerBase.visit(SerializerBase.java:273) ~[querydsl-core-5.0.0.jar:na]
	at com.blazebit.persistence.querydsl.BlazeCriteriaBuilderRenderer$JPQLNextExpressionSerializer.visit(BlazeCriteriaBuilderRenderer.java:1160) ~[blaze-persistence-integration-querydsl-expressions-1.6.6.jar:1.6.6]
	at com.blazebit.persistence.querydsl.BlazeCriteriaBuilderRenderer$JPQLNextExpressionSerializer.visit(BlazeCriteriaBuilderRenderer.java:1022) ~[blaze-persistence-integration-querydsl-expressions-1.6.6.jar:1.6.6]
	at com.querydsl.core.types.OperationImpl.accept(OperationImpl.java:88) ~[querydsl-core-5.0.0.jar:na]
	at com.blazebit.persistence.querydsl.BlazeCriteriaBuilderRenderer.renderCTEs(BlazeCriteriaBuilderRenderer.java:368) ~[blaze-persistence-integration-querydsl-expressions-1.6.6.jar:1.6.6]
	at com.blazebit.persistence.querydsl.BlazeCriteriaBuilderRenderer.access$400(BlazeCriteriaBuilderRenderer.java:106) ~[blaze-persistence-integration-querydsl-expressions-1.6.6.jar:1.6.6]
	at com.blazebit.persistence.querydsl.BlazeCriteriaBuilderRenderer$1.visit(BlazeCriteriaBuilderRenderer.java:255) ~[blaze-persistence-integration-querydsl-expressions-1.6.6.jar:1.6.6]
	at com.querydsl.core.types.SubQueryExpressionImpl.accept(SubQueryExpressionImpl.java:57) ~[querydsl-core-5.0.0.jar:na]
	at com.blazebit.persistence.querydsl.BlazeCriteriaBuilderRenderer.serializeSubQuery(BlazeCriteriaBuilderRenderer.java:131) ~[blaze-persistence-integration-querydsl-expressions-1.6.6.jar:1.6.6]
	at com.blazebit.persistence.querydsl.BlazeCriteriaBuilderRenderer.setExpressionSubqueries(BlazeCriteriaBuilderRenderer.java:841) ~[blaze-persistence-integration-querydsl-expressions-1.6.6.jar:1.6.6]
	at com.blazebit.persistence.querydsl.BlazeCriteriaBuilderRenderer.renderWhere(BlazeCriteriaBuilderRenderer.java:431) ~[blaze-persistence-integration-querydsl-expressions-1.6.6.jar:1.6.6]
	at com.blazebit.persistence.querydsl.BlazeCriteriaBuilderRenderer.access$800(BlazeCriteriaBuilderRenderer.java:106) ~[blaze-persistence-integration-querydsl-expressions-1.6.6.jar:1.6.6]
	at com.blazebit.persistence.querydsl.BlazeCriteriaBuilderRenderer$1.visit(BlazeCriteriaBuilderRenderer.java:260) ~[blaze-persistence-integration-querydsl-expressions-1.6.6.jar:1.6.6]
	at com.querydsl.core.types.SubQueryExpressionImpl.accept(SubQueryExpressionImpl.java:57) ~[querydsl-core-5.0.0.jar:na]
	at com.querydsl.core.support.FetchableSubQueryBase.accept(FetchableSubQueryBase.java:150) ~[querydsl-core-5.0.0.jar:na]
	at com.blazebit.persistence.querydsl.BlazeCriteriaBuilderRenderer$1.visit(BlazeCriteriaBuilderRenderer.java:230) ~[blaze-persistence-integration-querydsl-expressions-1.6.6.jar:1.6.6]
	at com.querydsl.core.types.OperationImpl.accept(OperationImpl.java:88) ~[querydsl-core-5.0.0.jar:na]
	at com.blazebit.persistence.querydsl.BlazeCriteriaBuilderRenderer$1.visit(BlazeCriteriaBuilderRenderer.java:172) ~[blaze-persistence-integration-querydsl-expressions-1.6.6.jar:1.6.6]
	at com.querydsl.core.types.OperationImpl.accept(OperationImpl.java:88) ~[querydsl-core-5.0.0.jar:na]
	at com.blazebit.persistence.querydsl.BlazeCriteriaBuilderRenderer$1.visit(BlazeCriteriaBuilderRenderer.java:252) ~[blaze-persistence-integration-querydsl-expressions-1.6.6.jar:1.6.6]
	at com.querydsl.core.types.SubQueryExpressionImpl.accept(SubQueryExpressionImpl.java:57) ~[querydsl-core-5.0.0.jar:na]
	at com.querydsl.core.support.FetchableSubQueryBase.accept(FetchableSubQueryBase.java:150) ~[querydsl-core-5.0.0.jar:na]
	at com.blazebit.persistence.querydsl.BlazeCriteriaBuilderRenderer.serializeSubQuery(BlazeCriteriaBuilderRenderer.java:131) ~[blaze-persistence-integration-querydsl-expressions-1.6.6.jar:1.6.6]
	at com.blazebit.persistence.querydsl.BlazeCriteriaBuilderRenderer.render(BlazeCriteriaBuilderRenderer.java:127) ~[blaze-persistence-integration-querydsl-expressions-1.6.6.jar:1.6.6]
	at com.blazebit.persistence.querydsl.AbstractBlazeJPAQuery.getQueryable(AbstractBlazeJPAQuery.java:223) ~[blaze-persistence-integration-querydsl-expressions-1.6.6.jar:1.6.6]
	at com.blazebit.persistence.querydsl.AbstractBlazeJPAQuery.createQuery(AbstractBlazeJPAQuery.java:174) ~[blaze-persistence-integration-querydsl-expressions-1.6.6.jar:1.6.6]
	at com.blazebit.persistence.querydsl.AbstractBlazeJPAQuery.createQuery(AbstractBlazeJPAQuery.java:165) ~[blaze-persistence-integration-querydsl-expressions-1.6.6.jar:1.6.6]
	at com.querydsl.jpa.impl.AbstractJPAQuery.fetch(AbstractJPAQuery.java:242) ~[querydsl-jpa-5.0.0.jar:na]
	at com.blazebit.persistence.querydsl.SetExpressionImpl.fetch(SetExpressionImpl.java:61) ~[blaze-persistence-integration-querydsl-expressions-1.6.6.jar:1.6.6]
```

### Steps to reproduce

Use a query like this `new BlazeJPAQuery<Tuple>(criteriaBuilderFactory).union(mainSelect1, mainSelect2, mainSelect3);` where the individual selects are simple queries.

### Environment

Version:            1.6.6
JPA-Provider:       Hibernate
DBMS:               -
Application Server: -

### LINKED ISSUE #1524: Support CTE's for left-nested set queries
### Description

The `StartOngoingSetOperationCriteriaBuilder` does not extend `CTEBuilder`. Therefore its currently not possible to add CTE's to a left-nested set query.

### Expected behavior
Adding CTE's to a left-nested set query is possible.

### Actual behavior

The method does not appear in the API.

### Environment

Version:            
JPA-Provider:       
DBMS:               
Application Server:

### LINKED ISSUE #1525: Bind entity shorthand does not work for complex embeddables in Querydsl
### Description

When producing a query with entity subqueries, like:

```java
    JPQLNextQuery<Tuple> select = queryFactory
            .from(selectFrom(person).where(person.id.eq(literal(1l))), new QPerson("a"))
            .from(selectFrom(person).where(person.id.eq(literal(2l))), new QPerson("b"))
            .innerJoin(selectFrom(person).where(person.id.eq(literal(3l))), new QPerson("c"))
                .on(Expressions.ONE.eq(Expressions.ONE))
            .select(new QPerson("a"), new QPerson("c"));
```

Currently the above may fail if the entity has some more complex (embeddable) mappings.

This is because the `BlazeCriteriaBuilderRenderer` does not currently utilise `JpaUtils.getEmbeddedPropertyPaths` to find the embedded property paths.

FWIW:  I also considered changing the code to use `fromEntitySubquery` instead of `fromSubquery` (and consorts) when we intend to perform an entity subquery. This would eliminate the need for the query renderer to render its own entity bindings.  However, I opted not to, because that way we can also support returning entities in set operations that are used as subquery. This is a feature that is rather difficult to support in core.

### Expected behavior
Entity subselect works in all cases.

### Actual behavior
Entity subselect does not work in all cases.

### Environment

Version:            
JPA-Provider:       
DBMS:               
Application Server:

jwgmeligmeyling:
## Description
Closes #1443
Closes #1460
Closes #1461
Closes #1464
Closes #1516
Closes #1524
Closes #1525

### DIFF
--- a/integration/querydsl/expressions/src/main/java/com/blazebit/persistence/querydsl/BlazeCriteriaBuilderRenderer.java
+++ b/integration/querydsl/expressions/src/main/java/com/blazebit/persistence/querydsl/BlazeCriteriaBuilderRenderer.java
@@ -55,6 +55,7 @@ import com.blazebit.persistence.WindowContainerBuilder;
 import com.blazebit.persistence.WindowFrameBetweenBuilder;
 import com.blazebit.persistence.WindowFrameBuilder;
 import com.blazebit.persistence.WindowFrameExclusionBuilder;
+import com.blazebit.persistence.impl.JpaUtils;
 import com.blazebit.persistence.parser.EntityMetamodel;
 import com.blazebit.persistence.parser.util.JpaMetamodelUtils;
 import com.blazebit.persistence.spi.ExtendedAttribute;
@@ -283,16 +284,12 @@ public class BlazeCriteriaBuilderRenderer<T> {
 
                             ExtendedManagedType<?> managedType = metamodel.getManagedType(ExtendedManagedType.class, pathExpression.getType());
                             Map<String, ? extends ExtendedAttribute<?, ?>> ownedSingularAttributes = managedType.getOwnedSingularAttributes();
+                            Collection<String> embeddedPropertyPaths = JpaUtils.getEmbeddedPropertyPaths(ownedSingularAttributes, null, false, false);
 
-                            for (Map.Entry<String, ? extends ExtendedAttribute<?,?>> ownedSingularAttribute : ownedSingularAttributes.entrySet()) {
-                                String attributeName = ownedSingularAttribute.getKey();
-                                ExtendedAttribute<?, ?> attribute = ownedSingularAttribute.getValue();
-
-                                if (!JpaMetamodelUtils.isAssociation(attribute.getAttribute())) {
-                                    final SelectBuilder<?> bindBuilder = selectBaseCriteriaBuilder.bind(attributeName);
-                                    BeanPath<?> beanPath = new BeanPath<Object>(attribute.getElementClass(), pathExpression, attributeName);
-                                    setExpressionSubqueries(beanPath, null,  bindBuilder, SelectBuilderExpressionSetter.INSTANCE);
-                                }
+                            for (String embeddedPropertyPath : embeddedPropertyPaths) {
+                                final SelectBuilder<?> bindBuilder = selectBaseCriteriaBuilder.bind(embeddedPropertyPath);
+                                BeanPath<?> beanPath = new BeanPath<>(Object.class, pathExpression, embeddedPropertyPath);
+                                setExpressionSubqueries(beanPath, null,  bindBuilder, SelectBuilderExpressionSetter.INSTANCE);
                             }
                         } else {
                             for (int i = 0; i < projection.size(); i++) {
--- a/integration/querydsl/expressions/src/main/java/com/blazebit/persistence/querydsl/BlazeCriteriaBuilderRenderer.java
+++ b/integration/querydsl/expressions/src/main/java/com/blazebit/persistence/querydsl/BlazeCriteriaBuilderRenderer.java
@@ -59,7 +59,6 @@ import com.blazebit.persistence.WindowFrameBuilder;
 import com.blazebit.persistence.WindowFrameExclusionBuilder;
 import com.blazebit.persistence.impl.JpaUtils;
 import com.blazebit.persistence.parser.EntityMetamodel;
-import com.blazebit.persistence.parser.util.JpaMetamodelUtils;
 import com.blazebit.persistence.spi.ExtendedAttribute;
 import com.blazebit.persistence.spi.ExtendedManagedType;
 import com.querydsl.core.JoinExpression;
@@ -448,6 +447,8 @@ public class BlazeCriteriaBuilderRenderer<T> {
 
     private <X extends FromBuilder<X>> X renderJoins(QueryMetadata metadata, FromBaseBuilder<X> fromBuilder) {
         X criteriaBuilder = null;
+        String rootAlias = null;
+
         for (final JoinExpression joinExpression : metadata.getJoins()) {
             boolean fetch = joinExpression.hasFlag(JPAQueryMixin.FETCH);
             boolean hasCondition = joinExpression.getCondition() != null;
@@ -518,6 +519,7 @@ public class BlazeCriteriaBuilderRenderer<T> {
                             }
                         }
 
+                        rootAlias = alias;
                         break;
                     default:
                         JoinType joinType = getJoinType(joinExpression);
@@ -530,7 +532,10 @@ public class BlazeCriteriaBuilderRenderer<T> {
                             if (!hasCondition) {
                                 throw new IllegalStateException("No on-clause for entity join!");
                             }
-                            final JoinOnBuilder<X> xJoinOnBuilder = criteriaBuilder.joinOn(entityPath.getType(), alias, joinType);
+                            if (rootAlias == null) {
+                                throw new IllegalArgumentException("An explicit base join node is required when multiple root nodes are used!");
+                            }
+                            final JoinOnBuilder<X> xJoinOnBuilder = criteriaBuilder.joinOn(rootAlias, entityPath.getType(), alias, joinType);
                             setExpressionSubqueries(joinExpression.getCondition(), null, xJoinOnBuilder, JoinOnBuilderExpressionSetter.INSTANCE);
                         } else if (!hasCondition) {
                             // If there is no alias, assume a default join
@@ -568,6 +573,7 @@ public class BlazeCriteriaBuilderRenderer<T> {
                         criteriaBuilder = o instanceof FinalSetOperationCTECriteriaBuilder ?
                                 ((FinalSetOperationCTECriteriaBuilder<X>) o).end() :
                                 ((FullSelectCTECriteriaBuilder<X>) o).end();
+                        rootAlias = alias;
                         break;
                     }
                     default: {
@@ -588,7 +594,10 @@ public class BlazeCriteriaBuilderRenderer<T> {
                             if (isLateral) {
                                 String subqueryAlias = subQueryExpression.getMetadata().getJoins().get(0).getTarget().accept(new JoinTargetAliasPathResolver(), null).getMetadata().getName();
                                 if (entityJoin) {
-                                    joinOnBuilderFullSelectCTECriteriaBuilder = criteriaBuilder.joinLateralOnSubquery(target.getType(), alias, joinType);
+                                    if (rootAlias == null) {
+                                        throw new IllegalArgumentException("An explicit base join node is required when multiple root nodes are used!");
+                                    }
+                                    joinOnBuilderFullSelectCTECriteriaBuilder = criteriaBuilder.joinLateralOnSubquery(rootAlias, target.getType(), alias, joinType);
                                 } else {
                                     joinOnBuilderFullSelectCTECriteriaBuilder = criteriaBuilder.joinLateralOnSubquery(renderExpression(fromPath), alias, subqueryAlias, joinType);
                                 }
@@ -596,7 +605,10 @@ public class BlazeCriteriaBuilderRenderer<T> {
                                 if (!entityJoin) {
                                     throw new IllegalStateException("Entity join to association");
                                 }
-                                joinOnBuilderFullSelectCTECriteriaBuilder = criteriaBuilder.joinOnSubquery(target.getType(), alias, joinType);
+                                if (rootAlias == null) {
+                                    throw new IllegalArgumentException("An explicit base join node is required when multiple root nodes are used!");
+                                }
+                                joinOnBuilderFullSelectCTECriteriaBuilder = criteriaBuilder.joinOnSubquery(rootAlias, target.getType(), alias, joinType);
                             }
 
                             Object o = serializeSubQuery(joinOnBuilderFullSelectCTECriteriaBuilder, target);
@@ -610,7 +622,10 @@ public class BlazeCriteriaBuilderRenderer<T> {
                                 String subqueryAlias = subQueryExpression.getMetadata().getJoins().get(0).getTarget().accept(new JoinTargetAliasPathResolver(), null).getMetadata().getName();
 
                                 if (entityJoin) {
-                                    xFullSelectCTECriteriaBuilder = criteriaBuilder.joinLateralSubquery(target.getType(), alias, joinType);
+                                    if (rootAlias == null) {
+                                        throw new IllegalArgumentException("An explicit base join node is required when multiple root nodes are used!");
+                                    }
+                                    xFullSelectCTECriteriaBuilder = criteriaBuilder.joinLateralSubquery(rootAlias, target.getType(), alias, joinType);
                                 } else {
                                     xFullSelectCTECriteriaBuilder = criteriaBuilder.joinLateralSubquery(renderExpression(fromPath), alias, subqueryAlias, joinType);
                                 }
--- a/integration/querydsl/expressions/src/main/java/com/blazebit/persistence/querydsl/BlazeCriteriaBuilderRenderer.java
+++ b/integration/querydsl/expressions/src/main/java/com/blazebit/persistence/querydsl/BlazeCriteriaBuilderRenderer.java
@@ -34,6 +34,7 @@ import com.blazebit.persistence.GroupByBuilder;
 import com.blazebit.persistence.HavingBuilder;
 import com.blazebit.persistence.JoinOnBuilder;
 import com.blazebit.persistence.JoinType;
+import com.blazebit.persistence.LeafOngoingFinalSetOperationCriteriaBuilder;
 import com.blazebit.persistence.LimitBuilder;
 import com.blazebit.persistence.MultipleSubqueryInitiator;
 import com.blazebit.persistence.ObjectBuilder;
@@ -47,6 +48,7 @@ import com.blazebit.persistence.SelectCTECriteriaBuilder;
 import com.blazebit.persistence.SelectRecursiveCTECriteriaBuilder;
 import com.blazebit.persistence.SetOperationBuilder;
 import com.blazebit.persistence.StartOngoingSetOperationBuilder;
+import com.blazebit.persistence.StartOngoingSetOperationCriteriaBuilder;
 import com.blazebit.persistence.SubqueryBuilder;
 import com.blazebit.persistence.SubqueryInitiator;
 import com.blazebit.persistence.WhereBuilder;
@@ -75,6 +77,7 @@ import com.querydsl.core.types.Order;
 import com.querydsl.core.types.OrderSpecifier;
 import com.querydsl.core.types.ParamExpression;
 import com.querydsl.core.types.Path;
+import com.querydsl.core.types.Predicate;
 import com.querydsl.core.types.SubQueryExpression;
 import com.querydsl.core.types.TemplateExpression;
 import com.querydsl.core.types.Visitor;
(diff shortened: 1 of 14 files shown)


integration/querydsl/expressions/src/main/java/com/blazebit/persistence/querydsl/BlazeCriteriaBuilderRenderer.java
```diff
@@ -1457,4 +1479,72 @@ public Expression<?> visit(TemplateExpression<?> expr, BindResolverContext conte
 
     }
 
+    /**
+     * Recursively find CTE's and render these against the CriteriaBuilder.
+     *
+     * @since 1.6.7
+     */
+    private class CTERenderVisitor implements Visitor<Object, Object> {
```
beikov, line 1487: Maybe I'm not seeing it, but can't this class be static?

integration/querydsl/expressions/src/main/java/com/blazebit/persistence/querydsl/JPQLNextQueryFactory.java
```diff
@@ -16,122 +16,276 @@
...
+import java.util.Collection;
+import java.util.List;
 
 /**
  * Query factory to simplify {@link BlazeJPAQuery} instantiation.
  *
  * @author Jan-Willem Gmelig Meyling
  * @since 1.6.2
  */
-public class JPQLNextQueryFactory implements JPQLQueryFactory {
```
beikov, line 38: The change from class to interface is incompatible and clients must recompile their code. Is this change really necessary? Maybe it's worth adding a new subclass and promote the use of that, but still keep this class around?

integration/querydsl/expressions/src/main/java/com/blazebit/persistence/querydsl/JPQLNextQuery.java
```diff
@@ -35,12 +36,11 @@
  * Query interface for JPQL.Next queries
  *
  * @param <T> Query result type
- * @param <Q> Concrete query type
  * @author Jan-Willem Gmelig Meyling
  * @since 1.5.0
  */
 @SuppressWarnings("unused")
-public interface JPQLNextQuery<T, Q extends JPQLNextQuery<T, Q>> extends JPQLQuery<T> {
+public interface JPQLNextQuery<T> extends JPQLQuery<T>, ExtendedFetchable<T> {
```
beikov, line 43: I know it sucks, but do you think it's possible to keep around this type variable to stay compatible with the old version? We can remove the type variable in 2.0 if you want. If you say this is necessary though, I'm fine with breaking user code, but AFAICT, it's a cosmetic change.

integration/querydsl/expressions/src/main/java/com/blazebit/persistence/querydsl/JPQLNextExpressions.java
```diff
@@ -1122,7 +1123,7 @@ public static <T> Expression<T> cast(Class<T> result, Expression<?> expression)
      * @param <T> Expression type
      * @return The treated expression
      */
-    public static <T> Expression<T> treat(Class<T> result, Expression<?> expression) {
+    public static <T> SimpleOperation<T> treat(Class<T> result, Expression<?> expression) {
```
beikov, line 1126: Is this change necessary? I couldn't find uses of the method in the changed code, so maybe this turns out to be just a cosmetic change as well? If you want, we can change the signatures in 2.0.

beikov (commented):
LGTM, but overall I would prefer if we could avoid signature changes. If you think we can do without signature changes, I would further suggest you compile a list of changes that you'd like to do in 2.0 in an issue and target 2.0

integration/querydsl/expressions/src/main/java/com/blazebit/persistence/querydsl/JPQLNextQueryFactory.java
jwgmeligmeyling: I doubt any users will currently use this class:
- Its was only recently introduced
- The documentation does not mention it or how to use it, all examples just use `new BlazeJPAQuery<>()` to construct new queries
- The class was pretty useless as it were without the signature changes.

integration/querydsl/expressions/src/main/java/com/blazebit/persistence/querydsl/BlazeCriteriaBuilderRenderer.java
jwgmeligmeyling: I don't think so, it needs to invoke some methods on the outer renderer.

integration/querydsl/expressions/src/main/java/com/blazebit/persistence/querydsl/JPQLNextQuery.java
jwgmeligmeyling: I don't think anyone currently uses the `JPQLNextQuery` class directly, because everywhere we just returned `BlazeJPAQuery` and none of our examples mentioned this API. None of my codebases interacts with  `JPQLNextQuery` directly currently.

jwgmeligmeyling: We can postpone the signature change though, but then I think in 2.0 the impact will be slightly more drastic, because with the new `JPQLQueryFactory` interface people might actually start programming against `JPQLNextQuery<T, ?>` instead of `BlazeJPAQuery<T>` which would then later need to be replaced with `JPQLNextQuery<T>`.

integration/querydsl/expressions/src/main/java/com/blazebit/persistence/querydsl/JPQLNextExpressions.java
jwgmeligmeyling: Not necessary no. I'll revert it.

jwgmeligmeyling:
> I would further suggest you compile a list of changes that you'd like to do in 2.0 in an issue and target 2.0

These signature changes would be commit https://github.com/Blazebit/blaze-persistence/pull/1526/commits/13099181b965f68ef5ddcaf4583790bc3ecf7e72 in entirety.

integration/querydsl/expressions/src/main/java/com/blazebit/persistence/querydsl/JPQLNextQuery.java
jwgmeligmeyling: I think eventually I would like to separate `-api` and `-impl` as separate modules in 2.0 (also to get rid of the provided scoped dependency). This is kind of a step towards that goal by properly separating API interfaces from implementation.

integration/querydsl/testsuite/src/test/java/com/blazebit/persistence/querydsl/BasicQueryTest.java
```diff
@@ -214,10 +230,29 @@ public void testFetchCount() {
...
+                    .from(person)
+                    .select(person, person)
+                    .where(bitwiseAndLong(person.id, 1L));
+
+            assertNotNull(query.fetchResults());
+        });
+    }
+
+    public static BooleanExpression bitwiseAndLong(NumberExpression<Long> expression, long bit) {
+        NumberExpression<Long> mod = expression.divide(1L << bit).floor().mod(2L);
```
jwgmeligmeyling, line 246: According to the Postgresql documentation, `ceil` should return its input type.

However:

```
portal=# select pg_typeof(5::bigint/2::bigint);
 pg_typeof 
-----------
 bigint
(1 row)

portal=# select pg_typeof(floor(5::bigint/2::bigint));
    pg_typeof     
------------------
 double precision
```

This then causes issues with the modulo operator, failing with `Caused by: org.postgresql.util.PSQLException: ERROR: function mod(double precision, bigint) does not exist`.

Should we cast the result of `floor` or just leave it to the user to solve and just add the cast to the test here to fix the issue with Postgres?

integration/querydsl/expressions/src/main/java/com/blazebit/persistence/querydsl/JPQLNextQueryFactory.java
beikov: Ok, that's convincing!

integration/querydsl/expressions/src/main/java/com/blazebit/persistence/querydsl/JPQLNextQuery.java
beikov: Sounds ok to me. Please just add a note to the changelog that people will probably have to recompile code when using the new version.

integration/querydsl/testsuite/src/test/java/com/blazebit/persistence/querydsl/BasicQueryTest.java
beikov: According to this (https://www.postgresql.org/docs/current/functions-math.html#id-1.5.8.9.6.2.2.9.1.1.1) it will return either numeric or double precision. Since `5/2` is a literal, PostgreSQL will probably infer `double precision` rather than `numeric`. Overall, I think this is something Hibernate 6.x should fix (can you please create an issue?). For the test, I'd say let's just cast `floor`, yes. But let's not change anything in the implementation.

jwgmeligmeyling: Added cast

integration/querydsl/expressions/src/main/java/com/blazebit/persistence/querydsl/JPQLNextExpressions.java
jwgmeligmeyling: The change was made to use in JPQLNextSerializer which initially replaced the mod operation expression with the expression returned from `treat(operator.getType(), operator)`. 

However this has since changed to simply `"TREAT_" + type.getSimpleName().toUpperCase()` because I didn't like the code.

This was a left over from that change and I have reverted it now.

beikov (approved):
LGTM

integration/querydsl/testsuite/src/test/java/com/blazebit/persistence/querydsl/BasicQueryTest.java
```diff
@@ -243,7 +244,7 @@ public void testModuloLongOperation() {
     }
 
     public static BooleanExpression bitwiseAndLong(NumberExpression<Long> expression, long bit) {
-        NumberExpression<Long> mod = expression.divide(1L << bit).floor().mod(2L);
+        NumberExpression<Long> mod = castToNum(Long.class, expression.divide(1L << bit).floor()).mod(2L);
```
jwgmeligmeyling, line 247: From the remaining test failure, it appears as if MySQL does not allow `cast(x as bigint)`. I believe I am using `CAST_LONG` from the [cast and treat functions](https://persistence.blazebit.com/documentation/1.6/core/manual/en_US/index.html#cast-and-treat-functions). Shall I just exclude the test for one or the other dialect?

jwgmeligmeyling: According to MySQL docs, `bigint` should have been fine: https://dev.mysql.com/doc/refman/8.0/en/integer-types.html

PR #1611, opened 2022-11-02
Fix maven artifact references for Jakarta JPA

thjanssen:
Getting started guide referenced the wrong maven artifacts for Jakarta JPA


beikov:
Thanks a lot!

PR #1659, opened 2022-12-29
[#1557] Add BlazeCriteriaQuery.applyToCriteriaBuilder to allow using a JPA Criteria query as set operand

### LINKED ISSUE #1557: Allow BlazeCriteriaQuery to apply to a BaseCriteriaBuilder
As shown in the [user question](https://stackoverflow.com/questions/73987553/using-multiple-cte-with-blaze-persistence-and-h2) it is hard to compose JPA Criteria parts into a bigger query. My suggestion is to add the following to `BlazeCriteriaQuery`:

```java
    public void applyToCriteriaBuilder(BaseCriteriaBuilder<T> builder);
```

This will allow users to compose a query like this:

```
CriteriaBuilder<Tuple> cb1 = factory.create(em, Tuple.class);
query1.applyToCriteriaBuilder( cb1 );
LeafOngoingSetOperationCriteriaBuilder<Tuple> cb2 = cb.unionAll();
query2.applyToCriteriaBuilder( cb2 );
FinalSetOperationCriteriaBuilder<Tuple> finalCb = cb2.endSet();
finalCb.orderByDesc("statusChangedAt").orderByDesc("id")
```

To fully solve the problem for the user we'd also need https://github.com/Blazebit/blaze-persistence/issues/1556 though.

beikov:
## Related Issue

Fixes #1557

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ Changes that happened in releases
 
 ### New features
 
-None yet
+* Add `BlazeCriteriaQuery.applyToCriteriaBuilder` to allow using a JPA Criteria query as set operand
 
 ### Bug fixes
 
--- a/jpa-criteria/api/src/main/java/com/blazebit/persistence/criteria/BlazeCriteriaQuery.java
+++ b/jpa-criteria/api/src/main/java/com/blazebit/persistence/criteria/BlazeCriteriaQuery.java
@@ -16,6 +16,8 @@
 
 package com.blazebit.persistence.criteria;
 
+import com.blazebit.persistence.BaseCriteriaBuilder;
+import com.blazebit.persistence.BaseQueryBuilder;
 import com.blazebit.persistence.CriteriaBuilder;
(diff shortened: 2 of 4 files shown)


donalpmccarthy (commented):
It would be super helpful if this apply functionality was available more generically.  I have a requirement to support a filter on an analytic function - but these window functions aren’t supported in the where clause.  The work around is either a select from a CTE or a sub query.  It would be helpful if you already have a criteria query to be able to pass it into a fromWith or fromSubquery.  So we end up with either:

select * from (…) a where a.rank = 1

or

with a as (…) select * from a where a.rank =1

In this example rank is a window function in the CTE or sub query of the form “rank() over (partition by … order by …) as rank”

When you already have a JPA criteria query but simply wish to wrap it in a parent select like above, I cannot see an easy way to achieve this without rebuilding the whole query from scratch using the BlazeCriteriaBuilder.  Any help appreciated on this request!

thanks - d

Issue #1653, opened 2022-12-19
GraphQL: Over-fetching on nested entity views


david-kubecka:
### Description

Given this example model
```
@EntityView(Document::class)
@GraphQLName("Document")
interface DocumentView {
    var type: DocumentType
    val files: List<FileView>
}

@EntityView(File::class)
@GraphQLName("File")
interface FileView {
    val name: String?
    val size: Long
}
```
and corresponding entities and GraphQL types, let's issue the following query:
```
query {
		documents { type }
}
```

### Expected behavior

Since `File`s are not queried at all I would expect that the resulting SQL query would select only the `document` table.

### Actual behavior

Both `document` and `file` tables are selected (joined).

### Analysis

The issue is probably caused by unexpected [fields](https://github.com/Blazebit/blaze-persistence/blob/382a66cd0fef3172e704b5350d62146bfa191c40/integration/graphql/src/main/java/com/blazebit/persistence/integration/graphql/GraphQLEntityViewSupport.java#L689) content coming from graphql-java. It contains `documents, documents.type` and the `documents` field causes [the whole subtree to be fetched](https://github.com/Blazebit/blaze-persistence/blob/139d83f7433dd80b0341ea6506d6d58ad2a62b62/entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/EntityViewConfiguration.java#L165) later once the setting is applied to a criteria builder.

The suggested solution is to simply ignore any non-leaf fields in the `applyFetches` method.

### Environment

BP: 1.6.8
Hibernate: 5.6

PR #1662, opened 2023-01-04
[#1653] Fix GraphQL overfetching issue

david-kubecka:
Make sure that EntityViewSetting.fetches are set up correctly,
i.e. they do not include intermediate/non-leaf fields as those
cause the whole subtree to be fetched from the DB.

### DIFF
--- a/examples/spring-data-graphql/pom.xml
+++ b/examples/spring-data-graphql/pom.xml
@@ -209,24 +209,24 @@
     </build>
 
     <profiles>
+        <!-- Use Spring Boot 2.0 as a minimum version because 1.* is no longer compatible with newer versions of graphql-spring-boot-starter -->
         <profile>
             <id>spring-data-1.11.x</id>
             <properties>
-                <!-- For some strange reason this doesn't work with 4.3.9.RELEASE. Since this is old anyway, let's just use 4.3.3.RELEASE here -->
-                <version.spring>4.3.3.RELEASE</version.spring>
-                <version.spring.boot>${version.spring-data-1.x-spring-boot}</version.spring.boot>
+                <version.spring>${version.spring-data-2.0-spring}</version.spring>
+                <version.spring.boot>${version.spring-data-2.0-spring-boot}</version.spring.boot>
             </properties>
 
             <dependencies>
                 <dependency>
                     <groupId>org.springframework.data</groupId>
                     <artifactId>spring-data-jpa</artifactId>
-                    <version>${version.spring-data-1.x}</version>
+                    <version>${version.spring-data-2.0}</version>
                 </dependency>
                 <dependency>
                     <groupId>org.aspectj</groupId>
                     <artifactId>aspectjweaver</artifactId>
-                    <version>1.8.9</version>
+                    <version>1.9.4</version>
                 </dependency>
                 <dependency>
                     <groupId>com.fasterxml.jackson.core</groupId>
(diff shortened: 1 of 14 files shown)


integration/graphql/pom.xml
```diff
@@ -59,6 +59,11 @@
             <artifactId>junit</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <version>3.6.28</version>
+        </dependency>
```
beikov, line 66: Please use `<scope>test</scope>`

beikov (commented):
Also, please add the issue number to the commit messages e.g. `[#1662] Fix GraphQL overfetching issue` and add a bug fix entry to the CHANGELOG.

integration/graphql/pom.xml
david-kubecka: done

david-kubecka:
> Also, please add the issue number to the commit messages e.g. `[#1662] Fix GraphQL overfetching issue` and add a bug fix entry to the CHANGELOG.

done

beikov (approved):
Changes look good, but please squash the commits to one.

david-kubecka:
@beikov In the end I had to modify the fix a little bit because it didn't work for custom/additional fields. Please have a look. Will you please run the CI workflow just for sure?

Also, I can surely squash the commits but why? :-) Don't you find the split into testing/refactoring/fixing commits easier to follow?

beikov:
>Also, I can surely squash the commits but why? :-) Don't you find the split into testing/refactoring/fixing commits easier to follow?

I prefer having one commit per issue :)
When you squash things, can you please make sure the license headers are up-to-date?

examples/spring-data-graphql/pom.xml
```diff
@@ -85,14 +85,21 @@
             <version>${project.version}</version>
         </dependency>
         <dependency>
-            <groupId>com.graphql-java</groupId>
+            <groupId>com.graphql-java-kickstart</groupId>
```
beikov, line 88: Looks like this change doesn't work for Spring Data 1.11.x:

```
Caused by: org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean found for dependency [graphql.kickstart.servlet.context.GraphQLServletContextBuilder]: expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {@org.springframework.beans.factory.annotation.Autowired(required=false)}
```

david-kubecka: Yeah, the newer versions included in the upgraded `graphql-java-kickstart` no longer works with boot 1.11.*. I suggest dropping the profile (or faking it into 2.0 as is already done in https://github.com/Blazebit/blaze-persistence/blob/020ebd73b188613d824bdda8f18172365013a062/examples/spring-data-dgs/pom.xml#L206).

beikov: Faking it sounds fine to me, can you please make the changes and then squash everything? :)

david-kubecka: I realized that dropping the profile would still lead to running the test, although with the default profile version (2.7 in this case). So it really doesn't matter and I faked it into 2.0.

beikov:
Great work, thanks!

PR #1663, opened 2023-01-06
stop ignoring tests in newer versions of spring-boot-starter-test

david-kubecka:
Since 2.4 the starter no longer includes junit-vintage-engine
for junit 4 tests so it has to be included explicitly.


beikov:
Thanks!

PR #1669, opened 2023-01-09
reproducer for [#1670]

david-kubecka:


beikov:
Also, please rebase this here @david-kubecka

david-kubecka:
@beikov I've rather updated the issue with link to now merged commit.

beikov:
So is https://github.com/Blazebit/blaze-persistence/issues/1670 also fixed now?

david-kubecka:
As I write in that issue I've made a quick fix which IMO is not very nice and perhaps not working in all cases, though. If you think it's good/safe enough we can close the issue.

beikov:
Gotcha, I'll keep the issue open for now then and think about it a bit more

Issue #1655, opened 2022-12-19
GraphQL: runtime error on union type in selection set


david-kubecka:
### Description

Let's have the following entity view model:
```
@EntityView(Person::class)
@GraphQLName("Person")
interface PersonView {
    var name: String
    var cat: CatView
}

@EntityView(Cat::class)
@GraphQLName("Cat")
interface CatView {
    var name: String
}

@EntityView(Dog::class)
@GraphQLName("Dog")
interface DogView {
    var name: String
}
```

with a single `QueryMapping` (using [spring-graphql](https://github.com/spring-projects/spring-graphql)):

```
    @QueryMapping
    fun getPerson(dfe: DataFetchingEnvironment, @Argument id: Long) =
         val setting = graphQLEntityViewSupport.createSetting(dfe)
         val cb0 = criteriaBuilderFactory.create(entityManager, Person::class.java).where(...)
         val cb = entityViewManager.applySetting(setting, cb0)
         ....
```

Corresponding GraphQL schema is
```
type Person {
    name: String!
    animal: Animal
}

union Animal = Cat | Dog

type Cat {
    name: String!
}

type Dog {
    name: String!
}

query {
      getPerson(id: ID!): Person
}
```

The query under test is following:
```
query {
	getPerson(id: 1312) {
		id
		name
		animal {
			__typename
		}
	}
}
```

Note that the `animal` union is in the selection set.

### Expected behavior

The `getPerson` query executes ok and returns data according to implementation.

### Actual behavior

The query throws error "java.lang.RuntimeException: Could not apply fetches" with the cause
```
Caused by: java.lang.ClassCastException: class graphql.schema.GraphQLUnionType cannot be cast to class graphql.schema.GraphQLFieldsContainer (graphql.schema.GraphQLUnionType and graphql.schema.GraphQLFieldsContainer are in unnamed module of loader 'app')
	at com.blazebit.persistence.integration.graphql.GraphQLEntityViewSupport.applyFetches(GraphQLEntityViewSupport.java:708)
```

The problematic cast is [here](https://github.com/Blazebit/blaze-persistence/blob/382a66cd0fef3172e704b5350d62146bfa191c40/integration/graphql/src/main/java/com/blazebit/persistence/integration/graphql/GraphQLEntityViewSupport.java#L708).

### Steps to reproduce

Create a model according to the description. There's no need to actually insert data as the query fails in the "planning" phase.

### Environment

BP: 1.6.8.
Hibernate: 5.6

PR #1671, opened 2023-01-09
[#1655]: GraphQL: fix runtime error on union type in selection set

david-kubecka:


integration/graphql/src/main/java/com/blazebit/persistence/integration/graphql/GraphQLEntityViewSupport.java
```diff
@@ -624,50 +602,41 @@ public void applyFetches(DataFetchingEnvironment dataFetchingEnvironment, Entity
...
-                String key = field.getQualifiedName();
-                if (key.length() < prefix.length()) {
+        OUTER:
+        for (SelectedField field : selectionSet.getFields()) {
+            if (!isLeaf(field.getType())) {
+                // Skip non-leaf types because these intermediate objects types would
+                // lead to fetching the whole subtree in EntityViewConfiguration#getFetches
+                continue;
+            }
+            String[] fieldParts = field.getFullyQualifiedName().split("/");
```
beikov, line 613: Is that the new format? Fields separated by `/` and every field has the format `TYPE_NAME.FIELD_NAME`? This graphql-java runtime surely is breaking things all the time..

Anyway, I think we should add caching for the fetch resolving. AFAIU, since EVs forbid cycles, there should only be a limited set of field paths possible, and since the paths are now apparently typed, we can use the FQN as cache key. Can you please add a `ConcurrentHashMap` as a cache for this?

integration/graphql/src/main/java/com/blazebit/persistence/integration/graphql/GraphQLEntityViewSupport.java
```diff
@@ -624,50 +602,41 @@ public void applyFetches(DataFetchingEnvironment dataFetchingEnvironment, Entity
...
-                        sb.append('.');
-                        fieldStartIndex = sb.length();
-                    } else {
-                        sb.append(c);
-                    }
+                if (META_FIELDS.contains(fieldName)) {
+                    // If this is a GraphQL meta field (e.g. __typename) then there's no EV mapping for it but still
+                    // we must fetch info from DB. So put the intermediate field path into fetches, even if it leads
+                    // into fetching the whole subtree in EntityViewConfiguration#getFetches (FIXME)
+                    continue;
```
beikov, line 628: Fetching the id here is a good idea

integration/graphql/src/main/java/com/blazebit/persistence/integration/graphql/GraphQLEntityViewSupport.java
```diff
@@ -624,50 +602,41 @@ public void applyFetches(DataFetchingEnvironment dataFetchingEnvironment, Entity
...
+                String mappedFieldPart = applyFieldMapping(typeName, fieldName);
+                if (mappedFieldPart == null) {
+                    // fieldName cannot be mapped to an entity view field -> ignore the whole field
+                    continue OUTER;
                 }
+                mappedFields.add(mappedFieldPart);
+            }
+            if (!mappedFields.isEmpty()) {
+                setting.fetch(String.join(".", mappedFields));
             }
```
beikov, line 639: Would be very happy if you could append to a `StringBuilder` directly instead of building an intermediate list first :)
Also, the `StringBuilder` can be reused by calling `setLength(0)` at the start of the loop :)

beikov (changes requested):
We're getting there :)
I left a few comments but the first draft looks good 👍

integration/graphql/src/main/java/com/blazebit/persistence/integration/graphql/GraphQLEntityViewSupport.java
```diff
@@ -389,48 +388,25 @@ public <T> EntityViewSetting<T, CriteriaBuilder<T>> createSetting(DataFetchingEn
...
-            return ((GraphQLObjectType) type).getName();
-        } else {
-            return ((GraphQLInterfaceType) type).getName();
-        }
+        return getElementType(dataFetchingEnvironment, elementRoot).getName();
     }
 
-    public GraphQLType getElementType(DataFetchingEnvironment dataFetchingEnvironment, String elementRoot) {
-        GraphQLType type = dataFetchingEnvironment.getFieldDefinition().getType();
+    public GraphQLUnmodifiedType getElementType(DataFetchingEnvironment dataFetchingEnvironment, String elementRoot) {
```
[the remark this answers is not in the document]
david-kubecka: It is actually compatible because the other types (not GraphQLUnmodifiedType) were precisely those which were manually unwrapped (List and NoNull). This is now done via the `unwrapAll` util function.

integration/graphql/src/main/java/com/blazebit/persistence/integration/graphql/GraphQLEntityViewSupport.java
```diff
@@ -624,50 +602,41 @@ public void applyFetches(DataFetchingEnvironment dataFetchingEnvironment, Entity
...
+            if (!isLeaf(field.getType())) {
+                // Skip non-leaf types because these intermediate objects types would
+                // lead to fetching the whole subtree in EntityViewConfiguration#getFetches
+                continue;
+            }
+            String[] fieldParts = field.getFullyQualifiedName().split("/");
+            String[] elementRootParts = elementRoot.isEmpty() ? new String[0] : elementRoot.split("/");
+            List<String> mappedFields = new ArrayList<>();
+            for (int i = 0; i < fieldParts.length; i++) {
+                String[] fieldPartComponents = fieldParts[i].split("\\.");
```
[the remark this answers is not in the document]
david-kubecka: Did you perform any perf measurements when you originally decided on the char-by-char processing? I guess the perf penalty would be pretty negligible here. If anything deserves optimization here, we should avoid processing the same subpaths again and again.

So what if I introduced a local (request only) cache here? I think this would bring real perf benefits while not compromising the readability of the code.

integration/graphql/src/main/java/com/blazebit/persistence/integration/graphql/GraphQLEntityViewSupport.java
david-kubecka: Ah, just made the same suggestion in the other comment :-)

david-kubecka: > This graphql-java runtime surely is breaking things all the time...

To be fair, this time they did not break the contract. `getFullyQualifiedName` is a new method.

integration/graphql/src/main/java/com/blazebit/persistence/integration/graphql/GraphQLEntityViewSupport.java
david-kubecka: Again, did you test this? From what I read (e.g. https://stackoverflow.com/questions/585860/string-join-vs-stringbuilder-which-is-faster) the `String.join` will not be significantly slower and might be actually faster in our case as it should be quite optimized.

integration/graphql/src/main/java/com/blazebit/persistence/integration/graphql/GraphQLEntityViewSupport.java
david-kubecka: > Can you please add a ConcurrentHashMap as a cache for this?

You mentioning ConcurrentHashMap suggests that you want to make the cache's lifecycle tied to the app lifecycle. I originally dismissed this thought from the start because you would have to somehow do cache pruning (see my comment above about request-only cache) but thinking about it more, the number of field paths in the schema is limited by definition, although it can be quite large when taking all combinations. Perhaps it would suffice to just cache the mapping for each `SelectedField` and not also all its subfields/subpaths? Then the cache size would be limited by the number of leaf nodes in the schema which should certainly be manageable.

WDYT?

beikov: I think we should aim for a single big cache for the whole app and add a config option to disable the caching.

integration/graphql/src/main/java/com/blazebit/persistence/integration/graphql/GraphQLEntityViewSupport.java
beikov: It's not about being faster, it's about the allocation rate due to the other unnecessary objects, `ArrayList` and the containing `Object[]`. Possible `byte[]` resizes etc.

integration/graphql/src/main/java/com/blazebit/persistence/integration/graphql/GraphQLEntityViewSupport.java
david-kubecka: Ok, will do that. Can I consider the other discussions regarding memory allocations as superseded by this?

david-kubecka:
@beikov Do you have any further comments on this PR? I believe I addressed all your points.

beikov (approved):

PR #1686, opened 2023-02-02
[#1457] Add Netflix DGS GraphQL integration that supports mutations

### LINKED ISSUE #1457: Update to DGS 4.9.19
Create a `InputObjectMapper` implementation to support updatable entity views.

https://github.com/Netflix/dgs-framework/discussions/775

beikov:
## Related Issue

Fixes #1457


EugenMayer:
Thank you so mutch @beikov , we will test drive this thing and try to get it up and running!

EugenMayer:
I cannot build the branch due to 

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.0.0:check (verify-style) on project blaze-persistence-integration-graphql: You have 1 Checkstyle violation. -> [Help 1]
```

In 
```
[INFO] Blazebit Persistence Integration GraphQL 0.0.0-DGS-MUTATOR-SNAPSHOT FAILURE 
```

Can i skip those checkstyles somehow? tried

```
mvn install -pl '!website,!documentation' -DskipTests -Dcheckstyle.skip
```

But it did not work. Pitty :)

EugenMayer:
Now i run into 

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.7.0:compile (default-compile) on project blaze-persistence-integration-graphql-dgs: Compilation failure: Compilation failure: 
[ERROR] /home/em/Development/repos/blaze-persistence/integration/graphql-dgs/src/main/java/com/blazebit/persistence/integration/graphql/dgs/BlazePersistenceDgsAutoConfiguration.java:[37,24] package javax.annotation does not exist
[ERROR] /home/em/Development/repos/blaze-persistence/integration/graphql-dgs/src/main/java/com/blazebit/persistence/integration/graphql/dgs/BlazePersistenceDgsAutoConfiguration.java:[57,6] cannot find symbol
[ERROR]   symbol:   class PostConstruct
[ERROR]   location: class com.blazebit.persistence.integration.graphql.dgs.BlazePersistenceDgsAutoConfiguration
[ERROR] -> [Help 1]
```

Seems like you now force to compile with java8 only? By accident?

EugenMayer:
I hate to be anoying, but know i'am back to checkstyle issues with `a44dcb2`

EugenMayer:
We could test and confirm that this indeed adds support for using Create/Update views as input types (and i guess any other view too). This is awesome!

We are using Enums and other things in our tests, it seems to work just as expected.

When you switch you will need to switch the integration from `com.blazebit:blaze-persistence-integration-graphql` to `com.blazebit:blaze-persistence-integration-graphql-dgs` and also remove your current type-registry ingtegration, since `com.blazebit:blaze-persistence-integration-graphql-dgs` does offer it build-in via `BlazePersistenceDgsAutoConfiguration`.

EugenMayer:
I'am yet not sure what the actual cause or pattern is, but i though i explain my observation and maybe you can direct me where to debug and help finding the cause.

Right now, with the changes you commited, most of the string "IDs" of our EVs are no deserialized to `Long` (all have been `String` in the past).

Interestingly some are `Long` but some are `Integer`. Both changed form `String` to one of the other.
We are talking of plain, non composite ID's, i cannot see anything special on the views or on the actual entities. All EV are interfaces, all getters are abstract.

Do you have a smell what the cause of this might be?

beikov:
>Right now, with the changes you commited, most of the string "IDs" of our EVs are no deserialized to Long (all have been String in the past).

This was part of the PoC to get this to work, but after some thinking and discussing on other channels, I think that it would be best to switch back to using `ID` as type and do conversion/coercion within the `InputObjectMapper`.

>Interestingly some are Long but some are Integer. Both changed form String to one of the other.

If some are `Integer` even though you use `Long` in your Java model, please report some more details, as that would be a bug.

EugenMayer:
@beikov as far as i know could narrow it, it is now "sometimes its long or integer", it is rather always integer.

Even non EVs like

```
  ldapCheckServerConfig(
        "Server ID" id: ID!
    ): Long!
```

Which returned a Long before your PR, now return an integer

PR #1699, opened 2023-02-26
Add spring-data-base-3.0 and spring-data-3.0 based on jakarta

EugenMayer:
Ensure we support spring-boot 3.0.0 with

- [x] Add spring-data-base-3.0 based on jakarta / java 17
- [x] Add spring-data-3.0 based on jakarta / java 17
- [ ] migrate away from spring.factories for auto-configuration classes
- [ ] `blaze-persistence-integration-spring-data-webmvc` needs adjustments
  - [ ]  Migrate away from WebMvcConfigurerAdapter to `WebMvcConfigurer`
  - [ ] migrate away from spring.factories

## Discussions
- [ ] `org.springframework.data:spring-data-jpa` version management seems overcomplicated. Why cant we just use a bom?
- [ ] `implement com.blazebit.persistence.spring.data.impl.query.EntityViewAwareRepositoryInformation#getFragments` since (new interface with spring-data 6.x)
- [ ] `implement com.blazebit.persistence.spring.data.impl.query.EntityViewAwareRepositoryMetadataImpl#getFragments`


integration/spring-data/3.0/pom.xml
```diff
@@ -0,0 +1,265 @@
...
+            <artifactId>jakarta.persistence-api</artifactId>
+            <version>${version.jakarta-jpa-api}</version>
+            <scope>provided</scope>
+        </dependency>
+
+        <!-- At runtime the user is going to need all dependencies anyway, so let's just give them through transitively -->
+        <dependency>
+            <groupId>com.blazebit</groupId>
+            <artifactId>blaze-persistence-integration-spring-data-base-3.0</artifactId>
+            <version>${project.version}</version>
```
beikov, line 120: This should not be necessary and could be the cause of the issues you mentioned on PM.
You are missing two entries for the new modules in parent/pom.xml in the `dependencyManagement` section, which will make this work without `<version>${project.version}</version>`

EugenMayer: Thanks, fixed both.

EugenMayer: This indeed fixed the issue with the snapshot - thank you!

EugenMayer:
> hateos We will need new modules for all spring related things that are for Spring 5+.

@beikov  Should i create spring-data-hateos-3.0 (spring data 3.x) module? You surely mean Spring 6.x (sb 3 is based on 6.x) - though spring data has a different versioning, seems to follow spring-boot versioning ( https://github.com/spring-projects/spring-data-jpa ) - if you could give me a hint i can finish that task

beikov:
>Should i create spring-data-hateos-3.0 (spring data 3.x) module?

Yes please.

beikov:
The problem with the tests seems to be a missing dependency:

> `Caused by: java.lang.ClassNotFoundException: jakarta.annotation.PostConstruct`

Please, also rebase the PR.

EugenMayer:
It was not really missing, it is just all mixed up because of instead of relying on the bom and the transitive dependenecies, we include way to much ourself and the the classpath is mixed up with versions that do not match

`jakarta.annotation-api` was already present, but in the version of 1.3.x (which is the javax package variant) and no longer compatible with spring boot 3.0.

This should fix the issue (thank for the hint of course .. i should have scrolled up the entire screen .. :) ). also rebased as you asked for

fk-currenxie:
🚀

PR #1717, opened 2023-03-18
Use basic view as Node type for as an connection.edge.node type by default fixes #1707 #1719

### LINKED ISSUE #1707: Having __typename in a GQL query causes the query planer to fetch the PK only
Due to the changes on `GraphQLEntityViewSupport#applyFetches` specifically on case when meta fields like `__typename` are detected in the field selection

```java
if (META_FIELDS.contains(fieldName)) {
    // If this is a GraphQL meta field (e.g. __typename) then there's no EV mapping for it but still
    // we must fetch the data from DB to populate it. If the enclosing EV is a standard view then use
    // its id field. Otherwise, skip this fieldPart and effectively put the intermediate field path
    // into fetches - in reality this wouldn't lead to any unnecessary joins since the EV in that case
    // corresponds to an embeddable entity.
    for (String metaFieldTypeName: field.getObjectTypeNames()) {
        // Meta field might have more parent types (in case of a union type). Try to find a first
        // ViewType among them and use its id field to ensure the type info can be obtained from DB.
        ManagedViewType<?> managedViewType = typeNameToViewType.get(metaFieldTypeName);
        if (managedViewType instanceof ViewType) {
            mappedFields.add(((ViewType<?>) managedViewType).getIdAttribute().getName());
            break;
        }
    }
    continue;
}
```

We see a regression (all query fields null in the results), that for example this query

```graphql
query GetLdapConfigurations {
  ldapServerConfigurations(first: 100, offset: 0) {
    edges {
      node {
        id
        hostname
        enabled
        groupConfiguration {
          id
          __typename
        }
        __typename
      }
    }
  }
}
```

fails, since the created EntityViewSetting include a fetcher for `id` which we assume causes the query planer to fetch the identity of the view (only).

When we compare the old implementation, the same query has no fetchers (as to be expected), with the new one, we have the id fetcher. If we remove `__typename`, we see the old behavior, no fetchers and the query works alright.

### LINKED ISSUE #1719: Paginated GraphQL calls use an XViewNode type as response
When  we declare a paginated data fetcher to return a view `CatView`, the type definitions that are generated for GraphQL will include 2 types (and more). 

- `CatView` (since this type exists as a view / java type)
- `CatViewNode` - this type is introduced by the BP GraphQL type generation

The problem is, that a paginated call will return `CatViewNodeViewConnection` in the graphqls schema, which itself nests `CatViewEdge` which then nests `CatViewNode`. 

The latter one `CatViewNode` should actually be `CatView` and those types are identical in the generated type definition, but they  are not the same.

The problem becomes more visible when the graphql client uses the schema to generate a `client-sdk` and uses types to deal with cache and flusing those. 

Is there any reason we introduced `XViewNode` in the first place? The entire type seems redundant to me.

EugenMayer:
Beforehand, we created a specific type `ViewNode` for the connection.edge.node type, which was a clone of the actual `View` that was to be returned. With this PR this is changed, by default, the actual `View` is used as an connection.edge.node type.

See #1719 for more information.

You can activate the old bahavior by setting `GraphQLEntityViewSupportFactory.setCreateNodeViewTypeForEdge(true)`

Misc:
- Also reproduces the `__typename` issue in #1707 (for DGS)

Fixes #1707. Fixes #1719

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Changes that happened in releases
 * Fix issue with unions in GraphQL selection set
 * Fix issue with `EMBEDDING_VIEW` macro when used in subquery within batch loaded mapping
 * Fix issue with resolving function return types in Hibernate 6 integration
+* Fix fetch determination for relay node types
 
 ### Backwards-incompatible changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ Changes that happened in releases
 ### Backwards-incompatible changes
 
 * The GraphQL integration bumped the graphql-java version to 17.4 and had to do some signature changes to fully support union types
+* Default for `GraphQLEntityViewSupportFactory.setDefineDedicatedRelayNodeTypes` changed to `false` to reuse object types for relay edge
 
 ## 1.6.8
 
--- a/examples/spring-data-dgs/src/main/java/com/blazebit/persistence/examples/spring/data/dgs/fetcher/CatGraphQLApi.java
+++ b/examples/spring-data-dgs/src/main/java/com/blazebit/persistence/examples/spring/data/dgs/fetcher/CatGraphQLApi.java
@@ -78,13 +78,4 @@ public class CatGraphQLApi {
         }
         return null;
     }
-
-    @DgsData(parentType = "CatWithOwnerViewNode", field = "theData")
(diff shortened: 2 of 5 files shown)


EugenMayer:
@beikov this one is ready for review.

I think the changelog message should be checked / polished by you at least, not sure what you need / write there usually

Issue #1722, opened 2023-04-11
Selecting a model subset (via fetches) leads to NPE when @EntityViewInheritance is involved


david-kubecka:
I don't have a simple reproducer for this but the situation is roughly following:
- EV A contains (possibly deeply nested) a collection of EV B which has `@EntityViewInheritance` defined
- I create an `EntityViewSetting` with **`fetches` which don't contain the Bs collection**
- I apply the setting to the CB and fetch `resultList`

Expected: The query executes ok.

Actual: NPE in https://github.com/Blazebit/blaze-persistence/blob/08793f4ab72cd57762bb6d222ad1fa9421f23e22/entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/objectbuilder/transformator/TupleTransformatorFactory.java#L135

Details: The `classMappingIndex` not present in the `tuple` corresponds to a "class type" select expression related to the Bs collection. That's why I think it's related to `@EntityViewInheritance`

david-kubecka:
Update: There's a crude fix/workaround https://github.com/Blazebit/blaze-persistence/pull/1723

PR #1723, opened 2023-04-11
[#1722] avoid NPE in TupleTransformatorFactory

david-kubecka:
For some unknown reason a classMappingIndex sometimes
points to a non-present field in the tuple (filtered
via fetches mechanism). The correct fix should be somewhere
else but that this is probably good enough - just don't
perfomr any tuple transformation for non-present fields.

Also, fix a small bug in hasSubFetches. The wrongly
implemented equality check lead to overfetching.


beikov:
Can you please add a test case that reproducers this issue? I'd like to make sure we don't reintroduce such a regression again.

beikov:
Superseded by https://github.com/Blazebit/blaze-persistence/pull/2018

PR #1731, opened 2023-05-22
Fixed wrong dependency in documentation

bramhaag:
Changed `blaze-persistence-entity-view-processor` to `blaze-persistence-entity-view-processor-jakarta`


beikov:
Thanks a lot!

PR #1734, opened 2023-06-01
Issues/1733 reset entity view id holder on success and failure

### LINKED ISSUE #1733: Possible entity view id leakage in Spring WebMVC integration
### Description
`com.blazebit.persistence.spring.data.webmvc.impl.json.EntityViewIdHandlerInterceptor` is not guaranteed to reset the ThreadLocal in  `com.blazebit.persistence.spring.data.webmvc.impl.json.EntityViewIdValueHolder` because the wrong handler method is used. The current implementation uses `postHandle` which is only invoked on handler success. Instead, `afterCompletion` should be used.

Mobe91:
Fixes #1733

### DIFF
--- a/integration/spring-data/testsuite/webmvc/pom.xml
+++ b/integration/spring-data/testsuite/webmvc/pom.xml
@@ -125,6 +125,56 @@
         </dependency>
     </dependencies>
 
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.bsc.maven</groupId>
+                <artifactId>maven-processor-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>process-test</id>
+                        <goals>
+                            <goal>process-test</goal>
+                        </goals>
+                        <phase>generate-test-sources</phase>
+                        <configuration>
+                            <!-- source output directory -->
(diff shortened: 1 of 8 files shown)


integration/spring-data/testsuite/webmvc/pom.xml
```diff
@@ -125,6 +125,56 @@
         </dependency>
     </dependencies>
 
+    <build>
```
beikov, line 128: The `<id>hibernate-apt</id>` profile already includes the annotation processor for testing. Is that not enough?

Mobe91: This is about the entity-view-processor, not about the Hibernate static metamodel.

beikov: That profile also generate the entity view static metamodel classes.

beikov (approved):

beikov:
Ahh, seems I didn't see you are using the builder in `DocumentControllerTest`. In that case, it might be better to simply make sure the hibernate-apt profile still works properly even if you declare the bsc-plugin in the main build section.

Issue #1758, opened 2023-07-10
Entity with `@Any`-mapped interface field in Hibernate 6 causes exception during Blaze configuration


ryanlarkin:
### Description
With Hibernate 6, `@Any`-mapped interface fields cause exception when calling `CriteriaBuilderConfiguration.createCriteriaBuilderFactory()`. Doing this would work in Hibernate 5.6 using Blaze 1.6.9.

### Expected behavior
`CriteriaBuilderConfiguration.createCriteriaBuilderFactory()` succeeds.

### Actual behavior
Hibernate throws `java.lang.IllegalArgumentException: Not a managed type`. E.g.:
```
java.lang.IllegalArgumentException: Not a managed type: interface org.hibernate.tutorial.em.Property
	at org.hibernate.metamodel.model.domain.internal.JpaMetamodelImpl.managedType(JpaMetamodelImpl.java:192)
	at org.hibernate.metamodel.model.domain.internal.MappingMetamodelImpl.managedType(MappingMetamodelImpl.java:467)
	at org.hibernate.metamodel.model.domain.internal.MappingMetamodelImpl.managedType(MappingMetamodelImpl.java:97)
	at com.blazebit.persistence.impl.EntityMetamodelImpl.discoverEnumTypes(EntityMetamodelImpl.java:309)
	at com.blazebit.persistence.impl.EntityMetamodelImpl.collectColumnNames(EntityMetamodelImpl.java:409)
	at com.blazebit.persistence.impl.EntityMetamodelImpl.<init>(EntityMetamodelImpl.java:127)
	at com.blazebit.persistence.impl.CriteriaBuilderFactoryImpl.<init>(CriteriaBuilderFactoryImpl.java:110)
	at com.blazebit.persistence.impl.CriteriaBuilderConfigurationImpl.createCriteriaBuilderFactory(CriteriaBuilderConfigurationImpl.java:2078)
	...
```

### Steps to reproduce

Add `Property`, its implementations, and `PropertyHolder` from [the Hibernate User Guide section on using `@Any`](https://docs.jboss.org/hibernate/orm/6.2/userguide/html_single/Hibernate_User_Guide.html#associations-any-property):
```Java
public interface Property<T> {
    String getName();
    T getValue();
}


@Entity
@Table(name="integer_property")
public class IntegerProperty implements Property<Integer> {
    @Id
    private Long id;

    @Column(name = "`name`")
    private String name;

    @Column(name = "`value`")
    private Integer value;

[...]

PR #1763, opened 2023-07-13
Exclude if the field is an interface

tapegram:
## Description
Most of the context is laid out in this issue https://github.com/Blazebit/blaze-persistence/issues/1758

Basically, hibernate no longer returns 'null' for Any types: https://github.com/hibernate/hibernate-orm/commit/18a52483ef6c8dd99a8e8499df59ad4ae38d66cb#diff-16e6d3e8450829e9b077f8b4ace8d630ae0a0faa56094f88f759a9ba6584b339L507-R512

So this line here errors out: https://github.com/Blazebit/blaze-persistence/blob/c3cab22c10a76420e5e13f033c1f9cb469673043/core/impl/src/main/java/com/blazebit/persistence/impl/EntityMetamodelImpl.java#L309C91-L309C122

The change is not an *ideal* fix, this just makes it stop erroring out for interfaces, which is our company's current use case and is blocking a critical upgrade to Spring 3. We also tried forking this repo and fixing it ourselves but are currently stuck getting it to build/deploy at all even without any changes :( getting in a quick patch to resolve this would be a huge win for us.

Our current backup plans are to completely remove all instances of [Any](https://docs.jboss.org/hibernate/orm/6.2/userguide/html_single/Hibernate_User_Guide.html#associations-any-property) or to remove all instances of blaze-persistence, which are both big lifts, complex, and kind of shuts down all dev for the whole company for an extended period of time while we rewrite everything :(

## Related Issue

https://github.com/Blazebit/blaze-persistence/issues/1758

## Motivation and Context

Included in description. Any help with this would be greatly appreciated. @beikov (tagging for visibility, we are kind of against the wall and would love to find a way to get this resolved as quickly as possible)

Thanks!


core/impl/src/main/java/com/blazebit/persistence/impl/EntityMetamodelImpl.java
```diff
@@ -305,7 +305,7 @@ private void discoverEnumTypes(Set<Class<?>> seenTypesForEnumResolving, Map<Stri
             }
 
             discoverEnumTypes(seenTypesForEnumResolving, enumTypes, enumTypesForLiterals, pluralAttribute.getElementType(), baseType, newParent, pluralAttribute, false);
-        } else if (!seenTypesForEnumResolving.contains(fieldType)) {
+        } else if (!seenTypesForEnumResolving.contains(fieldType) && !fieldType.isInterface()) {
```
beikov, line 308: The following would probably be more appropriate, because any attributes have a null `PersistentAttributeType`.

```suggestion
        } else if (!seenTypesForEnumResolving.contains(fieldType) && attribute.getPersistentAttributeType() != null) {
```

Mobe91:
I'm taking a look at this and try to add a test for it.

Mobe91:
@beikov I've updated this with the test I've created. But as discussed I have compilation issues with it and don't really get what's the problem. Maybe you can have a look at it.

beikov:
Thanks for the pre-work @Mobe91 and @tapegram. I fixed the remaining issues and squashed everything. Closing this PR as it is superseded by https://github.com/Blazebit/blaze-persistence/pull/1820

PR #1806, opened 2023-09-28
[#1805] Add logic to EntityViewManager and ProxyFactory to maintain e…

### LINKED ISSUE #1805: ProxyFactory#getProxy seems to re-use EVM instances from the cache without properly settings optionalParameters
Assuming that one uses `entityViewConfiguration.addOptionalParameters` to insert default parameters for an EVM to carry (specifically beans on needs to use during Pre/Post remove listeners), it seems that during the invocation attempt for thus listeners, and EVM instance is created via `context.getEntityViewManager().getSerializableDelegate(view.$$_getEntityViewClass())`
 which ultimately will use the `ProxyFactory` to create or **reuse* an EVM.

When an EVM is "reused" it is not ensured, that this EVM has the optionalParameters applied and thus we can end up having an EVM without those parameters included.

Mobe91:
Fixes #1805

### DIFF
--- a/entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/EntityViewManagerImpl.java
+++ b/entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/EntityViewManagerImpl.java
@@ -1450,6 +1450,11 @@ public class EntityViewManagerImpl implements EntityViewManager {
         return serializableDelegates.get(proxyFactory.getProxy(this, metamodel.managedViewOrError(entityViewClass)));
     }
 
+    @Override
+    public void close() {
+        proxyFactory.clear();
+    }
+
     /**
      * @author Christian Beikov
      * @since 1.2.0
--- a/documentation/src/main/asciidoc/entity-view/manual/en_US/getting_started.adoc
+++ b/documentation/src/main/asciidoc/entity-view/manual/en_US/getting_started.adoc
@@ -323,6 +323,11 @@ public class EntityViewManagerProducer {
         evm = config.createEntityViewManager(criteriaBuilderFactory);
     }
 
+    @PreDestroy
+    public void closeEvm() {
+        evm.close();
+    }
+
     @Produces
     @ApplicationScoped
     public EntityViewManager createEntityViewManager() {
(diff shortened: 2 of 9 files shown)


entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/EntityViewManagerImpl.java
```diff
@@ -1450,6 +1450,11 @@ public EntityViewManager getSerializableDelegate(Class<?> entityViewClass) {
         return serializableDelegates.get(proxyFactory.getProxy(this, metamodel.managedViewOrError(entityViewClass)));
     }
 
+    @Override
+    public void close() {
+        proxyFactory.clear();
```
beikov, line 1455: I think we should add a `static volatile boolean closed` as well and set this here, as well as check on every public method if it is closed to throw an exception when that is the case.
Apart from that, do you know if the static implementations (generated by the annotation processor) are "unloaded" properly as well?

EugenMayer:
Would be great to have this on the main branch to be able to test it all together!

entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/EntityViewManagerImpl.java
Mobe91: > I think we should add a static volatile boolean closed as well and set this here, as well as check on every public method if it is closed to throw an exception when that is the case.

Why would we do that? What's the point? The contract of the close method clearly states that any invocations on an entity view manager that has been closed is undefined.

> Apart from that, do you know if the static implementations (generated by the annotation processor) are "unloaded" properly as well?

No, I need to check that.

beikov: >Why would we do that? What's the point? The contract of the close method clearly states that any invocations on an entity view manager that has been closed is undefined.

Yes, but undefined is not very user friendly. If we get into some state when bad things might happen after closing, then we should ensure a user is not using the EVM anymore after a close. I created https://github.com/Blazebit/blaze-persistence/issues/1813 for this task as I don't want to hold back the release for this any further than necessary, but if you find time to work on that, I'd very much appreciate it.

Mobe91: > Apart from that, do you know if the static implementations (generated by the annotation processor) are "unloaded" properly as well?

Judging from `com.blazebit.persistence.view.impl.proxy.ProxyFactory#loadImplementation` static implementations should be handled properly as well by this change.

PR #1836, opened 2023-11-21
[#1835] Implement more types for GraphQl

### LINKED ISSUE #1835: FR: Support more types for GraphQL
Currently, we only support types like String, Lists, Sets, Int/Long and that's about it.

We  should also support types like `DateTime, Date, Time, BigInteger` and others when generating the GraphQL schema from EntityViews.

This has already been prepared but never fully implemented (was targeted for MicroProfile)

EugenMayer:
implements/ fixes #1835

### DIFF
--- a/integration/graphql/src/main/java/com/blazebit/persistence/integration/graphql/GraphQLEntityViewSupportFactory.java
+++ b/integration/graphql/src/main/java/com/blazebit/persistence/integration/graphql/GraphQLEntityViewSupportFactory.java
@@ -95,8 +95,9 @@ import java.util.regex.Pattern;
  */
 public class GraphQLEntityViewSupportFactory {
 
-    private static final Map<Class<?>, String> TYPES;
-    private static final Map<Class<?>, String> MP_TYPES;
+    private static final Map<Class<?>, String> BASIC_TYPES;
+    private static final Map<Class<?>, String> ADVANCED_TYPES;
+    private static final Map<Class<?>, String> SUPPORTED_TYPES;
 
     static {
         Map<Class<?>, String> types = new HashMap<>();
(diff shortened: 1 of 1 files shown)


EugenMayer:
Implemented the advanced type support using a class file test as we discussed @beikov - see https://github.com/Blazebit/blaze-persistence/pull/1836/files#diff-9f332e007d90470e77f0360a7cfbd8720be501ce3fc4f6839e237d34b20c25c9R157

Tell me if there is anything to adjust

beikov:
Superseded by https://github.com/Blazebit/blaze-persistence/pull/1846

EugenMayer:
Since i do not see any reason to not being open minded, i might as well be bold.

IMHO, it is not a good practice to  throw away work and re-roll the patch instead of giving feedback and let people finish the contribution. This does not exactly motivate to contribute in general. I have no hard feelings here, always glad if thing just move on. But i cannot say i'am happy either. 

Just take it as my personal feedback as a developer, not more or less :)

beikov:
Sorry, but it would have taken us both longer if I had written down the things that weren't quite right and I just wanted to save both our time so you can get your hands on this faster.
You were splitting up the collections into "standard" and "extended" types, but the standard ones contained types that are actually not standard. The extended ones were missing a few entries and as you mentioned, giving "Long" the wrong name.
You're still the commit author, so I'm not removing credit or anything. Thanks for your feedback though.

EugenMayer:
I'am aware that it would have taken longer, so i really see your motivation. Thanks for taking the feedback in!

But i guess, at least my POV, one has to take the route to guide the people. This way they align to the project ideas, structure, conduct and so forth - and thus contributions become better over time.

I just like open minded and of course, as written, no hard feelings. We are all pulling into the same direction in the end, that is all that matters to me / is essential to me :)

Issue #1842, opened 2023-11-27
Incompatibility with Spring 6.1(.1)


kuhnroyal:
### Description

The Spring integration uses assertions from `org.springframework.util.Assert` which now seem to require a message parameter.

```
Description:

An attempt was made to call a method that does not exist. The attempt was made from the following location:

    com.blazebit.persistence.integration.view.spring.impl.AnnotationEntityViewConfigurationSource.<init>(AnnotationEntityViewConfigurationSource.java:58)

The following method did not exist:

    'void org.springframework.util.Assert.notNull(java.lang.Object)'
```

### Steps to reproduce
Update a SpringBoot project to 3.2.0.

### Environment

Version:            1.6.10
JPA-Provider:       Hibernate 6.2.13.Final
DBMS:               Postgres
Application Server: Spring Boot 3.2.0

PR #1848, opened 2023-12-18
Fix spring boot 3.2 compatibility

rajadilipkolli:
From Spring 6.2, `Assert.notNull(Object)` is removed, hence adding supplier message to fix it.

This is stopping users from migrating to 3.2.x version of spring boot

### DIFF
--- a/integration/entity-view-spring/src/main/java/com/blazebit/persistence/integration/view/spring/impl/AnnotationEntityViewConfigurationSource.java
+++ b/integration/entity-view-spring/src/main/java/com/blazebit/persistence/integration/view/spring/impl/AnnotationEntityViewConfigurationSource.java
@@ -55,9 +55,9 @@ public class AnnotationEntityViewConfigurationSource extends AbstractEntityViewC
     public AnnotationEntityViewConfigurationSource(AnnotationMetadata metadata, Class<? extends Annotation> annotation,
                                                    ResourceLoader resourceLoader, Environment environment) {
         super(environment);
-        Assert.notNull(metadata);
-        Assert.notNull(annotation);
-        Assert.notNull(resourceLoader);
+        Assert.notNull(metadata,() -> "metadata cannot be null");
+        Assert.notNull(annotation,() -> "annotation cannot be null");
+        Assert.notNull(resourceLoader,() -> "resourceLoader cannot be null");
 
         this.attributes = new AnnotationAttributes(metadata.getAnnotationAttributes(annotation.getName()));
         this.configMetadata = metadata;
--- a/integration/entity-view-spring/src/main/java/com/blazebit/persistence/integration/view/spring/impl/AnnotationEntityViewConfigurationSource.java
+++ b/integration/entity-view-spring/src/main/java/com/blazebit/persistence/integration/view/spring/impl/AnnotationEntityViewConfigurationSource.java
@@ -55,9 +55,9 @@ public class AnnotationEntityViewConfigurationSource extends AbstractEntityViewC
     public AnnotationEntityViewConfigurationSource(AnnotationMetadata metadata, Class<? extends Annotation> annotation,
                                                    ResourceLoader resourceLoader, Environment environment) {
         super(environment);
-        Assert.notNull(metadata,() -> "metadata cannot be null");
-        Assert.notNull(annotation,() -> "annotation cannot be null");
-        Assert.notNull(resourceLoader,() -> "resourceLoader cannot be null");
+        Assert.notNull(metadata, "metadata cannot be null");
+        Assert.notNull(annotation, "annotation cannot be null");
+        Assert.notNull(resourceLoader, "resourceLoader cannot be null");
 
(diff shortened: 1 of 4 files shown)


rajadilipkolli:
Missed this in #1842

integration/entity-view-spring/src/main/java/com/blazebit/persistence/integration/view/spring/impl/AnnotationEntityViewConfigurationSource.java
```diff
@@ -55,9 +55,9 @@ public class AnnotationEntityViewConfigurationSource extends AbstractEntityViewC
     public AnnotationEntityViewConfigurationSource(AnnotationMetadata metadata, Class<? extends Annotation> annotation,
                                                    ResourceLoader resourceLoader, Environment environment) {
         super(environment);
-        Assert.notNull(metadata);
-        Assert.notNull(annotation);
-        Assert.notNull(resourceLoader);
+        Assert.notNull(metadata,() -> "metadata cannot be null");
+        Assert.notNull(annotation,() -> "annotation cannot be null");
+        Assert.notNull(resourceLoader,() -> "resourceLoader cannot be null");
```
beikov, line 60: Please change to this

```suggestion
        Assert.notNull(metadata, "metadata cannot be null");
        Assert.notNull(annotation, "annotation cannot be null");
        Assert.notNull(resourceLoader, "resourceLoader cannot be null");
```

rajadilipkolli: Hi @beikov , the reason to use supplier is for the purpose of performance gain , as per documentation below supplier get triggered only if it fails, where as if we dont specify it will be initialized during loading.

```
Params:
object – the object to check messageSupplier – a supplier for the exception message to use if the assertion fails
```

rajadilipkolli: Since backward compatibility is must, accepted the code changes

rajadilipkolli:
@beikov, added message in other missing places. After this merge, as code is compatible with Hibernate 6.4.0 Can you please release 1.6.10 version? So that users can update to Spring boot 3.2.1 and taking advantage of Hibernate 6.4.0

EugenMayer:
The current main branch works with an 3.2.1 and hub 6.4.1 without any issues already

EugenMayer (changes requested):
This changes do not fix sb 3.2.x, changes made to the wrong projects. The needed changes have already been made, the ones here would affect sb < 2.x instead

rajadilipkolli:
Hi @EugenMayer , is there snapshot way which I can test ?

EugenMayer:
@rajadilipkolli AFAIK you have to build one yourself (that is what we do), not sure if there are any snapshot builds from the CI published anywhere

beikov:
Snapshots are published to https://oss.sonatype.org/content/repositories/snapshots

rajadilipkolli:
> Snapshots are published to https://oss.sonatype.org/content/repositories/snapshots

In current workflow, publishing to snapshot is failing

beikov:
Sonatype repository uploads are a bit flaky, but I'll ensure the publish action runs through and let you know once it's done.

rajadilipkolli:
> Sonatype repository uploads are a bit flaky, but I'll ensure the publish action runs through and let you know once it's done.

Thanks @beikov , Can you also add instructions in Read Me on how to use Snapshots version. I want to check if my blocker which is blocking migrating to 3.2 is fixed or not

beikov:
Documented it [here](https://github.com/Blazebit/blaze-persistence?tab=readme-ov-file#using-snapshots) and the snapshot is also already published.

rajadilipkolli:
> Documented it [here](https://github.com/Blazebit/blaze-persistence?tab=readme-ov-file#using-snapshots) and the snapshot is also already published.

Thank you @beikov , it is working with spring boot 3.2.1.


QQ : Spring boot 3.2.1 comes with Hibernated 6.4 so is there a plan to provide below module?

```
        <dependency>
            <groupId>com.blazebit</groupId>
            <artifactId>blaze-persistence-integration-hibernate-6.4</artifactId>
            <version>${blaze-persistence.version}</version>
        </dependency>
```

beikov:
The 6.2 integration is compatible with 6.4. I don't have plans right now to add a relocation pom. Just use the 6.2 integration.

rajadilipkolli:
> Documented it [here](https://github.com/Blazebit/blaze-persistence?tab=readme-ov-file#using-snapshots) and the snapshot is also already published.

Hi @beikov ,

I am trying to access 1.6.16-SNAPSHOT as per above documentation. I am unable to access it. 

But I see github action to push the snapshot repo is successful.

Am I missing something here?

beikov:
Snapshots are now published elsewhere. See https://github.com/Blazebit/blaze-persistence?tab=readme-ov-file#using-snapshots

PR #1861, opened 2024-01-25
[#1858] Fix EntityViewAwareMappingJackson2HttpMessageConverter to use Spring Boot's ObjectMapper

### LINKED ISSUE #1858: EntityViewAwareMappingJackson2HttpMessageConverter doesn't use Spring Boot's ObjectMapper
### Description

[See Zulip discussion here](https://blazebit.zulipchat.com/#narrow/stream/338263-blaze-persistence/topic/Custom.20LocalDateTime.20Deserializer.20with.20Entity.20Views)

The `EntityViewAwareMappingJackson2HttpMessageConverter` class, which is registered as a converter in the `BlazePersistenceWebConfiguration` class, does not use the `ObjectMapper` provided by the Spring Boot context. This results in the default `ObjectMapper` from the `AbstractJackson2HttpMessageConverter` class being used, which may not include any custom deserializers or other configuration that the user has applied, such as custom configuration with `Jackson2ObjectMapperBuilder`.

### Expected behavior

The `EntityViewAwareMappingJackson2HttpMessageConverter` should use the `ObjectMapper` provided by the Spring Boot context. This would ensure that any custom configuration that the user has applied is used by Blaze Persistence.

### Actual behavior

The `EntityViewAwareMappingJackson2HttpMessageConverter` uses the default `ObjectMapper` from the `AbstractJackson2HttpMessageConverter` class. This may result in incorrect deserialization if the custom `ObjectMapper` includes any custom deserializers or other configuration that is not present in the default `ObjectMapper`.

### Steps to reproduce

1. Download the zip file containing the test case ([BlazePersistenceTestCaseCustomObjectMapperIssue.zip](https://github.com/Blazebit/blaze-persistence/files/14040432/BlazePersistenceTestCaseCustomObjectMapperIssue.zip));
2. Run the command `./mvnw clean install` at the root of the project;
3. Run the Spring Boot project with your IDE of choice or using the command `java -jar target/blaze-persistence-test-case-custom-object-mapper-issue-0.0.1-SNAPSHOT.jar`;
4. Make a POST request to `http://localhost:8080/users` with the following body:
```
{
  "username": "random_username",
  "createdAt": "2024-01-24T12:00:00+03:00"
}
```
The error will occur. You can debug and notice that the `LocalDateTimeCustomDeserializer` is not called by the Blaze Persistence converter (even though it is registered in Spring Boot through the `JacksonConfig` class. You can check this in the `UserController` class by debugging the injected `ObjectMapper` object). If you send the request without the `createdAt` field, it will succeed normally.

### Environment

Version: 1.6.11
Spring Boot: 3.2.2
JPA-Provider: Hibernate 6.4.1.Final
DBMS: h2, postgresql
Application Server: N/A

pvncoder:
## Description

The `EntityViewAwareMappingJackson2HttpMessageConverter` now uses the `ObjectMapper` provided by the Spring Boot context. This ensures that users' custom configurations are used.

## Related Issue

Fixes #1858

## Motivation and Context

This change allows users to use custom deserializers and other `ObjectMapper` configurations with Blaze Persistence Entity Views.

### DIFF
--- a/integration/spring-data/webmvc-jakarta/src/main/java/com/blazebit/persistence/spring/data/webmvc/impl/BlazePersistenceWebConfiguration.java
+++ b/integration/spring-data/webmvc-jakarta/src/main/java/com/blazebit/persistence/spring/data/webmvc/impl/BlazePersistenceWebConfiguration.java
@@ -48,11 +48,9 @@ public class BlazePersistenceWebConfiguration implements WebMvcConfigurer {
     protected final ObjectMapper objectMapper;
     private final EntityViewManager entityViewManager;
 
-    public BlazePersistenceWebConfiguration(
-      EntityViewManager entityViewManager,
-      @Qualifier("mvcConversionService") ObjectFactory<ConversionService> conversionService,
-      @Autowired(required = false) ObjectMapper objectMapper
-    ) {
+    public BlazePersistenceWebConfiguration(EntityViewManager entityViewManager,
+            @Qualifier("mvcConversionService") ObjectFactory<ConversionService> conversionService,
+            @Autowired(required = false) ObjectMapper objectMapper) {
         this.entityViewManager = entityViewManager;
         this.conversionService = conversionService;
         this.objectMapper = objectMapper == null ? new ObjectMapper() : objectMapper.copy();
@@ -87,7 +85,7 @@ public class BlazePersistenceWebConfiguration implements WebMvcConfigurer {
     @Override
     public void extendMessageConverters(List<HttpMessageConverter<?>> converters) {
         // Add it to the beginning, so it has precedence over the builtin
-        converters.add(0, new EntityViewAwareMappingJackson2HttpMessageConverter(entityViewManager, blazeWebmvcIdAttributeAccessor()));
+        converters.add(0, new EntityViewAwareMappingJackson2HttpMessageConverter(entityViewManager, blazeWebmvcIdAttributeAccessor(), objectMapper()));
     }
 
     @Override
--- a/integration/spring-data/webmvc-jakarta/src/main/java/com/blazebit/persistence/spring/data/webmvc/impl/json/EntityViewAwareMappingJackson2HttpMessageConverter.java
+++ b/integration/spring-data/webmvc-jakarta/src/main/java/com/blazebit/persistence/spring/data/webmvc/impl/json/EntityViewAwareMappingJackson2HttpMessageConverter.java
@@ -20,6 +20,7 @@ import com.blazebit.persistence.integration.jackson.EntityViewAwareObjectMapper;
 import com.blazebit.persistence.integration.jackson.EntityViewIdValueAccessor;
 import com.blazebit.persistence.view.EntityViewManager;
 import com.fasterxml.jackson.databind.JavaType;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import org.springframework.http.HttpInputMessage;
 import org.springframework.http.MediaType;
 import org.springframework.http.converter.HttpMessageNotReadableException;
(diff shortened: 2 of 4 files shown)


beikov:
Thanks a lot!

pvncoder:
No problem! I'm glad I could help.

EugenMayer:
This is a braking change. Hopefully this has been tested widely

EugenMayer:
in the previous version

```java
 module.setDeserializerModifier(new BeanDeserializerModifier() {
      public JsonDeserializer<?> modifyDeserializer(DeserializationConfig config, BeanDescription beanDesc, JsonDeserializer<?> deserializer) {
        ManagedViewType<?> view = entityViewManager.getMetamodel().managedView(beanDesc.getBeanClass());
        return (JsonDeserializer)(view != null ? new EntityViewReferenceDeserializer(entityViewManager, view, objectMapper, beanDesc.getIgnoredPropertyNames(), entityViewIdValueAccessor) : deserializer);
      }
    });
```

This part, 

```java
ManagedViewType<?> view = entityViewManager.getMetamodel().managedView(beanDesc.getBeanClass());
```

Is never called before the PR

beikov:
@pvncoder what do you think about copying the passed `ObjectMapper`? The problem @EugenMayer is experiencing is due to `EntityViewAwareObjectMapper` now altering Spring Boot's `ObjectMapper`.

I suppose the sole purpose of "using" Spring Boot's `ObjectMapper` is to allow the entity view serialization take advantage of the Jackson modules that you register in your Spring application, which is fine, but using a copy of the provided `ObjectMapper` would suffice to achieve that.

pvncoder:
@beikov yeah, the main purpose of utilizing Spring Boot's `ObjectMapper` is to leverage any custom Jackson modules registered within the Spring application context. If I understand correctly, the issue at hand is that Blaze Persistence might inadvertently modify the Spring Boot's `ObjectMapper`?

If yes, I've implemented a solution by creating a copy of the Spring Boot's `ObjectMapper` in the constructor of `BlazePersistenceWebConfiguration`:
```
public BlazePersistenceWebConfiguration(EntityViewManager entityViewManager,
        @Qualifier("mvcConversionService") ObjectFactory<ConversionService> conversionService,
        @Autowired(required = false) ObjectMapper objectMapper) {
    // source code omitted
    this.objectMapper = objectMapper == null ? new ObjectMapper() : objectMapper.copy();
}
```
When passing it to `EntityViewAwareMappingJackson2HttpMessageConverter`, I use the copied `ObjectMapper` via the `objectMapper()` getter method.

Is the part referenced by @EugenMayer modifying the original `ObjectMapper` from Spring's context instead of the copied one?

EugenMayer:
I do not think this will help my cause with the NPE when mocking - but it also does not need to. I think the mocking thing is something that we need to resolve on our end, most probably, we can no longer mock.

Nevertheless i think protecting spring boots ObjectMapper while still base on it, is a solid idea. The problem might be, that deciding to copy this within the constructor can be surprising  to anybody passing anything different then SB OM - maybe the direct reference is expected or even wanted.

So i ask myself, if we should actually let the user copy the OM when passing it, or not. Still i think we are generally heading into the right direct (IMHO)

beikov:
Yes @pvncoder, copying like you have shown is what I had in mind. Could you please provide a PR with this change?

pvncoder:
@beikov I apologize for the confusion in my previous comment. The code I sent in the current PR already copies the `ObjectMapper` in the constructor. In fact, that line was there before I committed the change.

image

As @EugenMayer mentioned, one of the potential problems with this approach is that it may surprise users who expect to pass a different reference to the `ObjectMapper`. This could be a problem for users who want to use the original reference. I'm not sure how to proceed with this issue.

Mobe91:
@EugenMayer 
What exactly is your use case where the copying would be an issue?

Right now users who want to avoid the copying could override the objectMapper() method.

```java
@Configuration
public class CustomBlazePersistenceWebConfiguration extends BlazePersistenceWebConfiguration {

    private final ObjectMapper originalObjectMapper;

    public BlazePersistenceWebConfiguration(EntityViewManager entityViewManager,
            @Qualifier("mvcConversionService") ObjectFactory<ConversionService> conversionService,
            @Autowired(required = false) ObjectMapper objectMapper) {
        super(entityViewManager, conversionService, objectMapper);
        this.entityViewManager = entityViewManager;
        this.conversionService = conversionService;
        this.originalObjectMapper = objectMapper;
    }

    @Override
    protected ObjectMapper objectMapper() {
        return this.originalObjectMapper;
    }
}
```
Though that might require the built-in `BlazePersistenceWebConfiguration` to be excluded from the context configuration somehow.

Issue #1870, opened 2024-02-18
SQL Server Dialect should use varchar(max) instead of varchar


Aniket-Singla:
### Description
While using Blaze Persistence, it was found that CAST_STRING function is casting the columns to varchar instead of varchar(max). The problem is varchar datatype in Sql Server is just 30 characters by default and hence rest of the value from column will be trimmed after the first 30 characters.

### Expected behavior
CAST_STRING , or any other String reference should use larger values and probably varchar(max).

### Actual behavior
CAST_STRING , or any other String dependent function is using varchar(30).

### Steps to reproduce
Just use any function like CAST_STRING to validate the current behavour.
More Details at : https://github.com/Blazebit/blaze-persistence/discussions/1868

Version:            
JPA-Provider: Spring Data JPA (Hibernate 6.2 as ORM)
DBMS: Sql Server
Application Server: Spring

PR #1871, opened 2024-02-18
[# 1870] Override MSSQLDbmsDialect to use varchar(max) for String

Aniket-Singla:
## Description
Overridden MSSQLDbmsDialect String equivalent in sql type to varchar(max).

## Related Issue
Fixes : [#1870](https://github.com/Blazebit/blaze-persistence/issues/1870)

## Motivation and Context
SQL Server has default length of 30 for varchar which will fall short for most of the users for doing string operations like CONCAT , CAST_STRING etc.


Aniket-Singla:
1. I have followed the same way its being done in DB2DbmsDialect . My personal Preference would be to make this behaviour change in other constructor as well like below to make it consistent until user specifically overrides String type with help of this constructor : 
```
public DB2DbmsDialect(Map<Class<?>, String> childSqlTypes) {
        Map<Class<?>, String> types = getSqlTypes();
        if(childSqlTypes!=null)
           types.putAll(childSqlTypes)
        super(types);
    }

```

2. I don't see any dialect specific test cases in existing code. So, I have not added the same. Also I have verified that existing test cases pass by using a sql server instance and below command:
```
mvn test -Pmssql,hibernate-5.6,deltaspike-1.9,spring-data-2.7.x -Djdbc.password=<my-local-pass>
```
@beikov let me know if we want to add an explicit TC for this. Currently I believe it is being covered with other Test Cases.

beikov (approved):

beikov:
Thanks!

PR #1893, opened 2024-04-19
[#1878] Avoid overriding a user provided deserializer

### LINKED ISSUE #1878: Deserializing EntityViews with jackson in Springboot yields null instead of an EntityView
### Description
When deserializing an EntityView in a DTO as payload in a `@RequestBody` the new Blaze version deserializes null while the old version can transform the id that is set, e.g. 3000 to the respective EntityView

The difference is, when deserializing, jackson (in `MethodProperty#deserializeAndSet`)  uses the `EntityViewReferenceDeserializer` in the newer case (probably as a fallback implementation) while when using the old Blaze version `<EntityIdViewName>Deserializer` is used.
The `EntityViewReferenceDeserializer` (new version) starts looking for an `{ idAttribute: <Id>}` `(com.blazebit.persistence.integration.jackson.EntityViewReferenceDeserializer#retrieveId)`  in a `JsonNode` that only contains the value of the id (e.g. 3000) and thus cannot find an id and cannot create the `EntityView`.

Probably in the new version the `<EntityIdViewName>Deserializer` is not registered with the `ObjectMapper`.

The payload we would send looks something like this:
```
{ niceEntity: 3000, ... }
```
With a DTO of:
```
@Data
....
NiceEntityDTO {
	private NiceEntityIdView niceEntity;
    ...
}
```
With the idView being something like this:
```
@EntityView(NiceEntity::class)
interface LongIdView {
    @get:IdMapping
    val id: Long?
}

```
In this example in the old version of Blaze, jackson would use the `NiceEntityIdViewDeserializer` while in the new version `EntityViewReferenceDeserializer`  is used to deserialize the `niceEntity` property in `NiceEntityDTO`.

### Expected behavior
After deserialization the DTO class should contain a NiceEntityIdView with an id of 3000.

### Actual behavior
After deserialization the DTO class should contains null for the niceEntity field.

### Environment
Current Version:  latest master build,
Before Version:    based on 191e13067
JPA-Provider:        Hibernate 6.4.4.Final

beikov:
## Related Issue

Fixes #1878

### DIFF
--- a/integration/jackson/src/main/java/com/blazebit/persistence/integration/jackson/EntityViewAwareObjectMapper.java
+++ b/integration/jackson/src/main/java/com/blazebit/persistence/integration/jackson/EntityViewAwareObjectMapper.java
@@ -28,6 +28,7 @@ import com.fasterxml.jackson.databind.JsonDeserializer;
 import com.fasterxml.jackson.databind.MapperFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.ObjectReader;
+import com.fasterxml.jackson.databind.deser.AbstractDeserializer;
 import com.fasterxml.jackson.databind.deser.BeanDeserializerModifier;
 import com.fasterxml.jackson.databind.introspect.VisibilityChecker;
 import com.fasterxml.jackson.databind.module.SimpleModule;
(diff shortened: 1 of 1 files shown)


EugenMayer:
@beikov seems like this strategy leads to quiet some issues, we have a good some of tests failing now. IMHO this should not be merged yet. i will report back with more details on the failing tests

EugenMayer:
TLDR: LGTM

@tom-mayer debugged this one and it turned out this patch unmasked an existing bug in our codebase, so the issues we had with this PR are 'self made'.

Still, IMHO this PR is still very valuable and valid, without this, custom deserializes would never be used.

PR #1905, opened 2024-05-16
[#1904] LocalDateBasicUserType incorrectly converting to java.sql.Timestamp (Multiset fetching)

### LINKED ISSUE #1904: Exception thrown when using LocalDate in an EntityView being fetched using the Multiset strategy
### Description
When using the Multiset fetch strategy on a collection whose entity view contains a LocalDate typed value, an exception is thrown when attempting to convert the database column to a java.sql.Timestamp before converting to a LocalDate.

### Expected behavior
The database column for a LocalDate should be converted from a java.sql.Date instead of a java.sql.Timestamp object.

### Actual behavior
```
java.lang.IllegalArgumentException: Timestamp format must be yyyy-mm-dd hh:mm:ss[.fffffffff]
    at java.sql/java.sql.Timestamp.valueOf(Timestamp.java:196)
    at com.blazebit.persistence.view.impl.type.LocalDateBasicUserType.fromString(LocalDateBasicUserType.java:36)
    at com.blazebit.persistence.view.impl.type.LocalDateBasicUserType.fromString(LocalDateBasicUserType.java:30)
    at com.blazebit.persistence.view.impl.objectbuilder.transformer.MultisetTupleTransformer.transform(MultisetTupleTransformer.java:78)
    at com.blazebit.persistence.view.impl.objectbuilder.transformator.TupleTransformator.transform(TupleTransformator.java:84)
    at com.blazebit.persistence.view.impl.objectbuilder.transformator.TupleTransformator.transform(TupleTransformator.java:77)
    at com.blazebit.persistence.view.impl.objectbuilder.ChainingObjectBuilder.build(ChainingObjectBuilder.java:51)
    at com.blazebit.persistence.impl.builder.object.PreProcessingObjectBuilder.build(PreProcessingObjectBuilder.java:46)
    at com.blazebit.persistence.impl.query.ObjectBuilderTypedQuery.getResultList(ObjectBuilderTypedQuery.java:71)
    at com.blazebit.persistence.impl.query.ObjectBuilderTypedQuery.getSingleResult(ObjectBuilderTypedQuery.java:49)  
```

### Steps to reproduce
Introduce a collection on an entity view that uses the Multiset fetch strategy whose entity view contains a LocalDate field

### Environment
Version:                       1.6.11        
JPA-Provider:             hibernate
DBMS:                         postgres 13.3
Application Server:    tomcat

dsarlo:
## Description
- Changed fromString method to convert date/date time strings to a LocalDate instead of to a java.sql.Timestamp first
- Added a LocalDate field to DocumentForCollections entity and the MultiSet fetch view associated with said backing entity
- Updated multiset fetch test case to use the LocalDate to make sure exception is no longer thrown and test passes

## Related Issue

Fixes #1904 

## Motivation and Context

Allows us to safely use a LocalDate in an entity view fetched using the Multiset fetch strategy without having to use a custom BasicUserType to override the behavior.

### DIFF
--- a/entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/type/LocalDateBasicUserType.java
+++ b/entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/type/LocalDateBasicUserType.java
@@ -18,22 +18,46 @@ package com.blazebit.persistence.view.impl.type;
 
 import com.blazebit.persistence.view.spi.type.BasicUserType;
 
-import java.sql.Timestamp;
 import java.time.LocalDate;
-import java.time.ZoneOffset;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeFormatterBuilder;
+import java.time.format.DateTimeParseException;
 
 /**
- *
  * @author Christian Beikov
  * @since 1.5.0
  */
 public class LocalDateBasicUserType extends TimestampishImmutableBasicUserType<LocalDate> {
 
     public static final BasicUserType<LocalDate> INSTANCE = new LocalDateBasicUserType();
+    private static final DateTimeFormatter OPTIONAL_TIME_OFFSET_AND_ZONE_FORMATTER = new DateTimeFormatterBuilder()
+            .parseCaseInsensitive()
+            .append(DateTimeFormatter.ISO_LOCAL_DATE)
+            .optionalStart()
+            .optionalStart()
+            .appendLiteral(' ')
+            .optionalEnd()
+            .optionalStart()
+            .appendLiteral('T')
+            .optionalEnd()
+            .append(DateTimeFormatter.ISO_LOCAL_TIME)
+            .optionalStart()
+            .appendOffsetId()
+            .optionalEnd()
+            .optionalStart()
+            .appendLiteral('[')
+            .appendZoneRegionId()
+            .appendLiteral(']')
+            .optionalEnd()
+            .optionalEnd().toFormatter();
 
     @Override
     public LocalDate fromString(CharSequence sequence) {
-        return Timestamp.valueOf(sequence.toString()).toInstant().atZone(ZoneOffset.UTC).toLocalDate();
+        String input = sequence.toString();
+        try {
+            return LocalDate.parse(input, OPTIONAL_TIME_OFFSET_AND_ZONE_FORMATTER);
+        } catch (DateTimeParseException e) {
+            throw new IllegalArgumentException("Invalid date or date-time format: " + input, e);
+        }
     }
-
 }
--- a/entity-view/testsuite/src/test/java/com/blazebit/persistence/view/testsuite/collections/entity/simple/DocumentForCollections.java
+++ b/entity-view/testsuite/src/test/java/com/blazebit/persistence/view/testsuite/collections/entity/simple/DocumentForCollections.java
@@ -28,6 +28,7 @@ import javax.persistence.OneToMany;
 import javax.persistence.OrderColumn;
 import javax.persistence.Table;
 import java.io.Serializable;
+import java.time.LocalDate;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -48,6 +49,7 @@ public class DocumentForCollections implements Serializable {
 
     private Long id;
     private String name;
+    private LocalDate dateCollected;
     private PersonForCollections owner;
     private Set<PersonForCollections> partners = new HashSet<PersonForCollections>();
     private Map<Integer, PersonForCollections> contacts = new HashMap<Integer, PersonForCollections>();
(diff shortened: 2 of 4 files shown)


dsarlo:
@beikov We can also do something like this to be compatible with someone using a datetime db type for a LocalDate. Lmk if you want me to change it to this

```
@Override
    public LocalDate fromString(CharSequence sequence) {
        String input = sequence.toString();

        try {
            return LocalDate.parse(input, DateTimeFormatter.ISO_LOCAL_DATE);
        } catch (DateTimeParseException e) {
            try {
                LocalDateTime dateTime = LocalDateTime.parse(input, DateTimeFormatter.ISO_LOCAL_DATE_TIME);
                return dateTime.toLocalDate();
            } catch (DateTimeParseException ex) {
                throw new IllegalArgumentException("Invalid date or date-time format");
            }
        }
    }
```

beikov:
Could you please use construct a `DateTimeFormatter` as static variable, that has an optional time part?

rajadilipkolli:
Hi @dsarlo , since you have worked on this. can you please extend the same to `OffsetDateTimeBasicUserType` as well. There as well its same issue.

dsarlo:
Thanks y'all! I’ll take a look and will make the changes in a bit

dsarlo:
@rajadilipkolli I do not see an issue with OffsetDateTimeBasicUserType. When using an OffsetDateTime in an entity view to trigger it, it's parsed correctly without issue.

@beikov Changes pushed - lmk if you think it's a bit overkill but personally I'd rather cover all bases

```
private static final DateTimeFormatter OPTIONAL_TIME_ZONE_AND_OFFSET_FORMATTER = new DateTimeFormatterBuilder()
            .parseCaseInsensitive()
            .append(DateTimeFormatter.ISO_LOCAL_DATE)
            .optionalStart()
            .optionalStart()
            .appendLiteral(' ')
            .optionalEnd()
            .optionalStart()
            .appendLiteral('T')
            .optionalEnd()
            .append(DateTimeFormatter.ISO_LOCAL_TIME)
            .optionalStart()
            .appendOffsetId()
            .optionalEnd()
            .optionalStart()
            .appendLiteral('[')
            .appendZoneRegionId()
            .appendLiteral(']')
            .optionalEnd()
            .optionalEnd().toFormatter();
```
Supports all of the following string formats

ZonedDateTime
OffsetDateTime
Instant
LocalDateTime
LocalDate
Date
Timestamp

entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/type/LocalDateBasicUserType.java
```diff
@@ -18,22 +18,39 @@
...
+        try {
+            return LocalDate.parse(input, OPTIONAL_TIME_FORMATTER);
+        } catch (DateTimeParseException e) {
+            throw new IllegalArgumentException("Invalid date or date-time format");
+        }
     }
 
+    @Override
+    public String toStringExpression(String expression) {
+        return "TO_CHAR(" + expression + ", 'YYYY-MM-DD')";
```
beikov, line 54: I think you will have to revert this change and make the `DateTimeFormatter` smarter if necessary, since the `TO_CHAR` function is not available on all databases.

dsarlo: Sounds good, I'll remove the override and will give the `DateTimeFormatter` another look. I had a version of it that supported a lot of different formats so I'll maybe add that back (Handled OffsetDateTime, ZonedDateTime, etc)

dsarlo: @beikov Made the suggested changes, however, on the `toStringExpression` comment.

```
@Override
    public String toStringExpression(String expression) {
        return "TO_CHAR(" + expression + ", 'YYYY-MM-DD HH24:MI:SS.US')";
    }
 ```

This is what is in the base class `TimestampishImmutableBasicUserType` so I overrode it and removed the time portion. If we don't want to use TO_CHAR, this should probably be changed and we should provide one without the time, but I'm not clear on what/how this is used so maybe it's better if that's something someone with more knowledge could address unless you can provide some guidance there.

File: entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/type/LocalDateBasicUserType.java [hunk 2]
```diff
@@ -18,22 +18,46 @@
...
+            .optionalEnd().toFormatter();
 
     @Override
     public LocalDate fromString(CharSequence sequence) {
-        return Timestamp.valueOf(sequence.toString()).toInstant().atZone(ZoneOffset.UTC).toLocalDate();
+        String input = sequence.toString();
+        try {
+            return LocalDate.parse(input, OPTIONAL_TIME_OFFSET_AND_ZONE_FORMATTER);
+        } catch (DateTimeParseException e) {
+            throw new IllegalArgumentException("Invalid date or date-time format");
```
beikov, line 60:
```suggestion
            throw new IllegalArgumentException("Invalid date or date-time format: " + input, e);
```

beikov:
Thanks for your work. I'll try to get this through the finish line as part of https://github.com/Blazebit/blaze-persistence/pull/1911

PR #1918, opened 2024-07-11
Fix table inheritance not being able to find the entity metamodel

EugenMayer:


EugenMayer:
@beikov our test-suite was passing just fine with this fix, including the problematic part. Seems like the BP suite does pass too

EugenMayer:
@beikov any chance to get this merged? Is anything left to do?

jwgmeligmeyling (approved):
Change looks obvious to me 😄

PR #1933, opened 2024-08-29
Add missing TREAT, CAST and LITERAL functions

jwgmeligmeyling:
I noticed a few more treat functions were added recently. I ended up needing them in the Querydsl integration so I decided to add them 😉

### DIFF
--- a/integration/querydsl/expressions/pom.xml
+++ b/integration/querydsl/expressions/pom.xml
@@ -8,6 +8,8 @@
     </parent>
     <properties>
         <module.name>com.blazebit.persistence.querydsl</module.name>
+        <maven.compiler.source>1.8</maven.compiler.source>
+        <maven.compiler.target>1.8</maven.compiler.target>
     </properties>
     <modelVersion>4.0.0</modelVersion>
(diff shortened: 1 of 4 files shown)


integration/querydsl/expressions/pom.xml
```diff
@@ -8,6 +8,8 @@
     </parent>
     <properties>
         <module.name>com.blazebit.persistence.querydsl</module.name>
+        <maven.compiler.source>1.8</maven.compiler.source>
+        <maven.compiler.target>1.8</maven.compiler.target>
```
jwgmeligmeyling, line 12: Wasn't the whole project changed to Java 8 minimally already? I thought that was the blocker for `TREAT_INSTANT` and others to be added to begin with?

beikov: It wasn't, the support of `java.time` types was doable without a hard dependency on Java 8. Version 2.0 which I'm working on currently will be based on Java 11 though, so we can get rid of some of the hacks.

jwgmeligmeyling: Ok, then we'll let this PR sit until 2.0 as well 😄

Issue #1522, opened 2022-07-27
Unable to make a paginated query (GraphQl) for an entity view with inheritance configured


dsarlo-viso:
### Description

I have an entity view that has EntityViewInheritance configured. I am trying to create a paginated graphql query wrapped in a GraphQLRelayConnection.

This is my entity view and graphql schema:
```
@JsonTypeInfo (use = JsonTypeInfo.Id.NAME, property = "type", visible = true)
@JsonSubTypes({
    @JsonSubTypes.Type (value = EventNotification.class, name = Notification.EVENT_TYPE),
    @JsonSubTypes.Type (value = TestNotification.class, name = Notification.TEST_TYPE),
    @JsonSubTypes.Type (value = TipNotification.class, name = Notification.TIP_TYPE),
})
@EntityView (Notification.class)
@EntityViewInheritance
public abstract class NotificationView {
    @IdMapping
    public abstract Long getId();

    public abstract Long getUserId();
}

type Notification {
    id: ID!
    userId: ID!
}

type NotificationConnection {
    edges: [NotificationEdge]
    pageInfo: PageInfo
}

type NotificationEdge {
    node: Notification!
    cursor: String!
}

type PageInfo {
    startCursor: String
    endCursor: String
}
```

The following query is being made:
```
 CriteriaBuilder<Notification> cb = cbf.create(em, Notification.class).from(Notification.class);

        cb = cb.where("userId").eq(userId);

        EntityViewSetting<NotificationView, PaginatedCriteriaBuilder<NotificationView>> setting = graphQLEntityViewSupport.createPaginatedSetting(NotificationView.class, dataFetchingEnvironment);
        setting.addAttributeSorter("id", Sorters.ascending());
        if (setting.getMaxResults() == 0) {
            return new GraphQLRelayConnection<>(Collections.emptyList());
        }
        PaginatedCriteriaBuilder<NotificationView> paginatedCriteriaBuilder = evm.applySetting(setting, cb);
        return new GraphQLRelayConnection<>(paginatedCriteriaBuilder.getQuery().getResultList());
```

### Expected behavior

I expect that I would get the entity view's back as edge nodes in the response

### Actual behavior

All of the node data - the entity views I'm querying for - are coming back null

```
"data": {
    "findAllReadAndNewNotificationsForCurrentUser": {
      "edges": [
        null,
        null,
        null,
        null,
        null
      ],
      "pageInfo": {
[...]

PR #1960, opened 2024-11-27
[#1522] Unable to make a paginated query (GraphQl) for an entity view with inheritance configured

kurbaniec:
## Description
Adapted `ViewTypeObjectBuilder` to fetch discriminator column instead of selecting `NULL` when inheritance and `public <X> SubGraph<X> fetch(String path)` are in use.

## Related Issue
#1522 

## Motivation and Context
I encountered a similar issue at my company, though it wasn’t directly related to GraphQL. The problem occurred with certain `public <X> SubGraph<X> fetch(String path)` calls of `EntityViewSetting<T, Q extends FullQueryBuilder<T, Q>>` for inheritance queries. From what I understand of the codebase, this mechanism is also involved in the GraphQL query flow.

Because I approached the issue from a slightly different angle, I’m not entirely certain this fix resolves every aspect of the GraphQL case, but based on my tests, I believe it should help. 

This is my first time contributing to Blaze-Persistence, so I’m not entirely sure if I’ve done everything correctly. I’d really appreciate any guidance or feedback to help me improve this PR. Specifically:

* Does the approach I’ve taken make sense, or are there better ways to solve this?
* Are there any additional tests or scenarios I should include?
* Would it help if I added more comments or documentation to explain the changes?

I’m eager to learn and happy to make any changes needed. Thank you so much for taking the time to review this! 😊

### DIFF
--- a/entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/objectbuilder/ViewTypeObjectBuilder.java
+++ b/entity-view/impl/src/main/java/com/blazebit/persistence/view/impl/objectbuilder/ViewTypeObjectBuilder.java
@@ -9,6 +9,7 @@ import com.blazebit.persistence.FullQueryBuilder;
 import com.blazebit.persistence.ObjectBuilder;
 import com.blazebit.persistence.ParameterHolder;
 import com.blazebit.persistence.SelectBuilder;
+import com.blazebit.persistence.view.impl.objectbuilder.mapper.AliasExpressionTupleElementMapper;
 import com.blazebit.persistence.view.impl.objectbuilder.mapper.TupleElementMapper;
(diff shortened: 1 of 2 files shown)


beikov:
Thanks for this work. The direction of the approach looks good. I would like to understand what effect this change is going to have on generated queries though.
My main concern is that a join could be created that is unexpected/unnecessary. If an entity view has a polymorphic subview, it might create a join for that subview even though none of its properties are fetched/requested. You seem to have tests that cover this scenario logically, but the check in the `isInheritance()` method looks like it might be too lenient. Maybe you can reassure me a bit or you see a problem in your code? :)

beikov:
Thanks for the great work! I debugged this locally and it looks good to me :)

Issue #1998, opened 2025-02-14
Upgrade Quarkus to 3.8 and switch to `@ConfigMapping` for the configuration


gsmet:
At some point in the future, we will retire the legacy config classes so let's migrate to the new @ConfigMapping infrastructure.

So let's upgrade to Quarkus 3.8 as the minimum version and update to the new config infrastructure.

I will create a PR shortly.

PR #1999, opened 2025-02-14
[#1998] Upgrade to Quarkus 3.8 and switch to `@ConfigMapping`

gsmet:
## Description

Update Quarkus to 3.8 as the minimum version and switch to the new `@ConfigMapping` infrastructure.

## Related Issue

#1998 

## Motivation and Context

At some point in the future, we will retire the legacy config classes so let's migrate to the new @ConfigMapping infrastructure.

I was conservative and only updated to Quarkus 3.8.

Note that there might be more optimizations possible with having all elements in a single map but I tried to be as less intrusive as possible as I don't know the codebase.

### DIFF
--- a/examples/pom.xml
+++ b/examples/pom.xml
@@ -49,6 +49,25 @@
                 <module>spring-hateoas</module>
                 <module>it-service-management</module>
                 <module>quarkus</module>
+                <module>microprofile-graphql</module>
+            </modules>
+        </profile>
+        <profile>
+            <id>jdk17+</id>
+            <activation>
+                <jdk>[17,)</jdk>
+            </activation>
+            <modules>
+                <module>showcase</module>
+                <module>deltaspike-data-rest</module>
+                <module>spring-data-webmvc</module>
+                <module>spring-data-webflux</module>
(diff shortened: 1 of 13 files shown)


beikov:
Hi @gsmet and thanks for your PR to get rid of the legacy config root model. I suppose the test errors are because you bumped the Quarkus version to build and test against to 3.8, but didn't update the Hibernate ORM version in the `deployment/pom.xml` profiles to 6.4 or 6.6 (I forgot what version you use in Quarkus 3.8).

Is RHBQ 3.2 out of support already? If so, I'm happy to drop support for it and upgrade the minimum to Quarkus 3.8, but just wanted to make sure before we do this.

gsmet:
Thanks for pointing me to the profiles.

I pushed something that should be a bit better.

Note that I enforce 6.4 for testing for anything else than 6.6. And for 6.6, I test with Quarkus 3.15. What I was a bit unsure of is which profile you use to release. As we wouldn't want to use 3.15 for the release.

I confirm 3.2 is out of support, both in community and RHBQ.

beikov:
Thanks a lot for taking on this task!

Issue #2002, opened 2025-02-21
Get paginated results with inline ID query disabled doesn't work with jakarta 3.x


kizzywuzzit:
I'm using jakarta 3.1.0 and noticed that when getting paginated results on a query that does a join to a oneToMany relationship that it was failing with
> Unexpected value type (expected : java.lang.Long) : [Ljava.lang.Object;@31d7d44c ([Ljava.lang.Object;)

The problem appears to be that when the ID query is generated it does so by producing a `TypedQuery<Object[]>` which results in the IDs being mapped as `List<Object[]>` where each ID is stored in a single element object array within the list. I was able to resolve this by just having the return type as `TypedQuery<Object>` so that the ID mapping gets correctly generated as `List<Long>`or any other ID type.

PR https://github.com/Blazebit/blaze-persistence/pull/2003

PR #2003, opened 2025-02-21
#2002 Play nice with jakarta 3.x and disabled inline id queries

kizzywuzzit:
## Description
Changed response type for getIdQuery from `TypedQuery<Object[]>` to `TypedQuery<Object>` to avoid doubly listing the ID parameters

## Related Issue
#2002

## Motivation and Context
Solves getting paginated results on jakarta 3.x with inline ID query disabled

### DIFF
--- a/core/impl/src/main/java/com/blazebit/persistence/impl/PaginatedCriteriaBuilderImpl.java
+++ b/core/impl/src/main/java/com/blazebit/persistence/impl/PaginatedCriteriaBuilderImpl.java
@@ -1001,7 +1001,7 @@ public class PaginatedCriteriaBuilderImpl<T> extends AbstractFullQueryBuilder<T,
 
     private TypedQuery<Object> getIdQuery(String idQueryString, boolean normalQueryMode, Set<JoinNode> keyRestrictedLeftJoins, List<JoinNode> entityFunctions) {
         if (normalQueryMode && isEmpty(keyRestrictedLeftJoins, ID_QUERY_CLAUSE_EXCLUSIONS)) {
-            TypedQuery<Object> idQuery = em.createQuery(idQueryString, Object.class);
+            TypedQuery<Object> idQuery = (TypedQuery<Object>) em.createQuery(idQueryString);
             if (isCacheable()) {
                 mainQuery.jpaProvider.setCacheable(idQuery);
             }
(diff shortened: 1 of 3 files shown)


core/testsuite/src/test/java/com/blazebit/persistence/testsuite/PaginationTest.java
```diff
@@ -285,6 +285,29 @@ public void testSelectIndexedWithParameterForceIdQuery() {
         cb.setParameter("contactNr", 1).getResultList();
     }
 
+    @Test
```
kizzywuzzit, line 288: This test technically passes before my changes, because there is no existing configuration for jakarta 3.x in this repo. Once one is added, this test should sufficiently cover the scenario (I verified separately with the code I'm using that ha jakarta 3.1.0)

beikov:
I'd be happy to merge your PR, but could you please look into the failures?

core/impl/src/main/java/com/blazebit/persistence/impl/PaginatedCriteriaBuilderImpl.java
```diff
@@ -1001,7 +1001,7 @@ private Map.Entry<TypedQuery<T>, ObjectBuilder<T>> getObjectQuery(boolean normal
 
     private TypedQuery<Object> getIdQuery(String idQueryString, boolean normalQueryMode, Set<JoinNode> keyRestrictedLeftJoins, List<JoinNode> entityFunctions) {
         if (normalQueryMode && isEmpty(keyRestrictedLeftJoins, ID_QUERY_CLAUSE_EXCLUSIONS)) {
-            TypedQuery<Object> idQuery = em.createQuery(idQueryString, Object.class);
+            TypedQuery<Object> idQuery = (TypedQuery<Object>) em.createQuery(idQueryString);
```
kizzywuzzit, line 1004: Funny thing about this is that pre 6.2 hibernate only skips verification for `Object[].class` but 6.x skips for both `Object.class` and `Object[].class`. Because we're explicitly using Object as the return type here, it's safe to bypass the verification performed by the method which also takes in the return class.

kizzywuzzit:
Seems like it did get all the tests to pass

beikov:
Awesome, thanks a lot!

Issue #2010, opened 2025-03-06
Join on subquery and sort by left joined column causes unknown parameter error when inline ID query is disabled


kizzywuzzit:
The scenario required to hit this is definitely a bit weird, but I just happened to need weird. When a query joins on a windowed CTE, AND sorts by a coalesced left joined column, it produces
> java.lang.IllegalArgumentException: Invalid or unknown parameter with name: ids_0_0

This stems from the fact that the `ids_#` parameters are never registered in the parameter manager, so when `getObjectQueryById` gets called it creates the query with only the `param_#` parameters registered. When `PaginatedTypedQueryImpl` iterates and calls `setParameter` the underlying `AbstractCustomQuery` doesn't have it in the parameter list, and throws the exception.

PR is incoming with fix

Issue #295, opened 2016-11-19
Add support for Oracle cycle clause


beikov:
Since Oracle is the only DB that has cycle detection, some queries only fail on Oracle because of that. The cycles could be ignored in most if not all cases of our testsuite since they occurr only in a tree where we start from the children instead of the parents.
The cycle clause should be configurable through properties since Oracle is the only DB that requires this and I wouldn't want to add an API for that.

See the reference: https://docs.oracle.com/cloud/latest/db112/SQLRF/statements_10002.htm#i2161315

PR #2011, opened 2025-03-06
#2010 unknown ids on window subquery

kizzywuzzit:
## Description
Fix issue where join on window query and sort by left joined column breaks

## Related Issue
https://github.com/Blazebit/blaze-persistence/issues/2010

## Motivation and Context
See issue above for info


kizzywuzzit:
All the tests failed, with a cache error?

beikov:
Thanks for notifying me about this. I updated the GitHub action step versions and rebased your PR.

kizzywuzzit:
Thank you, I've looked through the now failing tests, most of these seem to be due to the join on subquery not supported on hibernate pre-5.1, or unsupported in datanucleus/eclipselink. I've disabled the new test on those. The failures on the oracle DB seem to be related to #295 just as all the CTE tests are disabled for oracle, so I've also disabled testing on that.

beikov:
Looks good to me, can you please squash the commits?

kizzywuzzit:
Squashed and pushed, but isn't that usually done during merge?

beikov:
>Squashed and pushed, but isn't that usually done during merge?


Thanks. It can be done during merge, but then I think that authorship attribution might be messed up. Either way, having the PR follow our standards is the preferred approach.

core/impl/src/main/java/com/blazebit/persistence/impl/query/CustomQuerySpecification.java
```diff
@@ -156,7 +155,9 @@ protected void initialize() {
         List<Query> participatingQueries = new ArrayList<>();
 
         for (Map.Entry<String, Collection<?>> entry : listParameters.entrySet()) {
-            baseQuery.setParameter(entry.getKey(), entry.getValue());
+            if (!entry.getValue().isEmpty()) {
```
beikov, line 158: Is this really necessary? I think this might cause problems if you first execute a query with a list of e.g. 2 elements and then again with an empty list.

core/impl/src/main/java/com/blazebit/persistence/impl/PaginatedCriteriaBuilderImpl.java
```diff
@@ -1062,6 +1061,14 @@ private TypedQuery<T> getObjectQueryById(boolean normalQueryMode, Set<JoinNode>
 
         if (identifierExpressionsToUse.length == 1) {
             parameterListNames.add(ID_PARAM_NAME);
+        } else if (identifierExpressionsToUse.length > 1) {
+            for (Parameter<?> p : baseQuery.getParameters()) {
+                if (p.getName().startsWith(skippedParameterPrefix)) {
+                    parameterManager.registerParameterName(p.getName(), true, null, this);
```
beikov, line 1067: Instead of "dirtying" the parameter manager, I would suggest we create `ParameterManager.ParameterImpl` instances directly in the else branch below and add these instances to the `parameters` collection.

core/impl/src/main/java/com/blazebit/persistence/impl/query/CustomQuerySpecification.java
kizzywuzzit: Yes, the reason it's needed is that the ids_# values aren't set during the initalisation like the param_# values are, they're populated later on. We skip the setting here, otherwise it causes `ERROR: <AST>:0:0: unexpected end of subtree`

core/impl/src/main/java/com/blazebit/persistence/impl/PaginatedCriteriaBuilderImpl.java
kizzywuzzit: @beikov just wanted to follow up if you've had a chance to read above or if you have any suggestions for what I was seeing?

beikov: Hi and sorry it took me so long. If you could spend some time to try and figure out a way to make this work without registering parameters in the parameter manager, that would be really helpful. Otherwise I have to look into this more deeply which I am not sure when I can get to. Unfortunately, I don't have any good suggestions other than what I already wrote. I'm kind of surprised that values parameters play a role here, since there is no values clause in your test case. Maybe you mixed something up there?

kizzywuzzit: Okay, I'll try to take a look. as for
> there is no values clause in your test case. Maybe you mixed something up there?

It's because I've also disabled the inline ID query so when it first fetches the IDs, it's then creating the object query taking in IDs as values intrinsically.

PR #2015, opened 2025-03-13
[#2014] Copying criteria builder with CTE throws exception

### LINKED ISSUE #2014: Copying criteria builder with CTE throws exception
When copying a criteria builder which uses a CTE, the selection fails with
> Usage of CTE 'com.blazebit.persistence.testsuite.entity.DocumentNodeCTE' without definition!

This is due to the copy method copying the CTEs with the original CTE key, but the new criteria builder uses a separate joinManager object, so the lookup fails.

PR https://github.com/Blazebit/blaze-persistence/pull/2015

kizzywuzzit:
## Description
Build CTE Key using copied join manager instead of original so that the CTE can be found during selection.

## Related Issue
Fixes #2014

## Motivation and Context
See ticket above


beikov:
Thanks!

Issue #2042, opened 2025-05-24
LazyKeysetLink#initialize using the same orderByExpression in every loop iteration - returned keyset is wrong


laurinstenz:
When trying to use keyset pagination as follows, I got a hibernate type conversion error:

```
.orderByAsc("enumProperty")
.orderByAsc("stringProperty")
.afterKeyset()
    .with("enumProperty", SomeEnum.VALUE)
    .with("stringProperty", "someStringValue")
```

The problem was that Hibernate was trying to match SomeEnum.VALUE with the stringProperty.

---

I tracked the bug down to the LazyKeysetLink where the returned keyset tuple was `new Serializable[] {SomeEnum.VALUE, SomeEnum.VALUE}` instead of the expected `new Serializable[] {SomeEnum.VALUE, "someStringValue"}`

```java
String expressionString = orderByExpressions.get(0).getExpression().toString();
```
where it actually should be 

```java
String expressionString = orderByExpressions.get(i).getExpression().toString();
```

I'll create a PR with a unit test covering this issue.

PR #2043, opened 2025-05-24
[#2042] Fix LazyKeysetLink#initialize using the same orderByExpression in every loop iteration

laurinstenz:
## Description
LazyKeysetLink#initialize now uses the loops actual orderByExpression instead of reusing the first one over and over again.

## Related Issue
#2042 

## Motivation and Context
This change makes it possible to use the `afterKeyset().with("expression", "value")`, with multiple `with` statements. Before it would reuse the value of the first `with` expression for the whole keyset, this is now fixed.


laurinstenz:
Don't worry, I'm happy to contribute to such a library. Thank you for merging :pray:

PR #2117, opened 2026-06-11
[#2113] Fix SIOOBE when a VALUES clause and an inline CTE join are nested inside another inline CTE

### LINKED ISSUE #2113: SIOOBE in removeSyntheticPredicate with CTEs
In Blaze Persistence 1.6.18 (`-jakarta` artifacts) on Hibernate 7.2.x, using `fromValues(...)` followed by `leftJoinOn(...)` **inside** an outer `with(SomeCte.class)…end()` block trips a `StringIndexOutOfBoundsException` during SQL rendering, at `EntityFunction.removeSyntheticPredicate` (line 110 in 1.6.18). The exact same shape used in the **main query** works correctly.

  ## Versions

  - `blaze-persistence` **1.6.18** (`blaze-persistence-core-impl-jakarta`, `blaze-persistence-integration-hibernate-7.2`)
  - Hibernate ORM **7.2.12.Final**
  - Jakarta Persistence **3.2**
  - PostgreSQL 17 (Testcontainers)
  - Java 25

  ## Minimal reproducer

  Self-contained reproducer with passing/failing JUnit 5 tests: https://github.com/heruan/blaze-issues/pull/1

  The PR contains two tests:

  1. `mainQueryShapeWorks` — `fromValues + leftJoinOn` in the main query (with an inner CTE). **Passes.** Mirrors your own
  `InlineCTETest.testJoinInlineEntityWithLimit`, included as a sanity baseline.
  2. `insideCteShapeCrashes` — same shape wrapped inside an outer `with(CashBalanceCte.class)…end()`. **Crashes** with the SIOOBE.

  ## The shape

  The failing call is roughly:

  ```java
  cb.with(MonthlyAggCte.class)               // inner aggregate CTE (works alone)
      .from(Movement.class, "m")
      .groupBy("FUNCTION('TRUNC_MONTH', m.date)")
      .bind("month").select("FUNCTION('TRUNC_MONTH', m.date)")
      .bind("amount").select("SUM(m.amount)")
      .end();

  cb.with(CashBalanceCte.class)              // <-- the outer wrapper is what trips it
      .fromValues(Movement.class, "date", "ms", months)
      .leftJoinOn(MonthlyAggCte.class, "agg")
          .on("agg.month").eqExpression("ms")
      .end()
      .bind("month").select("ms")
      .bind("balance").select("COALESCE(agg.amount, 0)")
      .end();

  cb.from(CashBalanceCte.class, "c").select("c.month").select("c.balance");
  cb.getResultList();   // boom
```
  `CashBalanceCte` and `MonthlyAggCte` are plain `@CTE @Entity` POJOs with just an `@Id` and one or two basic columns; nothing exotic.

  Stack trace (truncated):
```
  java.lang.StringIndexOutOfBoundsException: ...
      at com.blazebit.persistence.impl.function.entity.EntityFunction
          .removeSyntheticPredicate(EntityFunction.java:110)
      ...
```
  The synthetic-predicate cleanup pass appears to call `sb.charAt(-1)` because it cannot locate the predicate it expects inside the rendered SQL, i.e. the shape detection misses the `fromValues(...) + leftJoinOn(...)` form when it is nested inside an outer `with(...)` definition.

  ## Related issues

  - Possibly related: #2093 (open) — same `leftJoinOn(CTEEntity)` codepath, manifests as malformed SQL (null is null and 999=999) instead of SIOOBE.

  ## Temporary workaround

  Keep `fromValues` + `leftJoinOn` in the main query and only register the inner aggregate as a CTE. The outer query joins the `VALUES` list to the inner CTE via `leftJoinOn`, projects `Tuple` rows to a private record that implements a public view interface — the rest of the application code is shaped as if the outer CTE existed, so reverting once this is fixed is mechanical (entity-view + `@With(CTEProvider) `annotations dropped onto the existing view interface, two service methods collapsed back to a single `applyView(...)` call).

heruan:
Fixes #2113

## Root cause

`JoinManager.buildClause` emits a synthetic marker group (`... and 999=999 and (<alias>.<attr> is null)`) for every VALUES clause and inline CTE node, and one `entity_function` wrapper per node is supposed to consume its own group during SQL rendering. `EntityFunction.render` located "its" marker with a blind `lastIndexOf(" and 999=999")`.

When a CTE body contains both a `fromValues(...)` and a `leftJoinOn(...)` of another inline CTE, `asExpression` wraps the body in nested `entity_function` calls (VALUES node innermost) and the innermost render call sees both marker groups. It picked the inline CTE join's marker, derived the wrong values table alias from it, substituted that alias into the synthetic predicate, and then could not find the predicate in the SQL (`indexOf` returning `-1`), crashing with `StringIndexOutOfBoundsException` at `EntityFunction.removeSyntheticPredicate`. The same shape in the main query works because that path (`CustomQuerySpecification.applySqlTransformations`) iterates over all nodes.

## Fix

`render` now scans marker candidates from last to first and picks the one that verifiably belongs to the current invocation:

- VALUES clause node (non-empty synthetic predicate): the alias-substituted synthetic predicate must immediately precede the marker
- inline CTE node (empty synthetic predicate): the `null is null` placeholder rendered from `NULLFN` must immediately precede the marker

If no candidate matches, it falls back to the last marker, i.e. the previous behavior, so currently working shapes cannot regress.

## Testing

- New `InlineCTETest#testJoinInlineCteInsideInlineCte` reproduces the exact SIOOBE without the fix; `testJoinInlineEntityInsideInlineCte` covers the plain entity join variant
- Full core testsuite green on H2 + Hibernate 5.6 (2313 tests) and on the Hibernate 7.2 jakarta runner (1966 tests)
- Verified against the original report's stack (Hibernate ORM 7.2.12, Jakarta Persistence 3.2, PostgreSQL 17, Java 25) via the standalone reproducer from the issue

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ None yet
 
 ### Bug fixes
 
-None yet
+* Fix `StringIndexOutOfBoundsException` when using a `VALUES` clause and an inline CTE join inside another inline CTE
 
 ### Backwards-incompatible changes
 
--- a/core/impl/src/main/java/com/blazebit/persistence/impl/function/entity/EntityFunction.java
+++ b/core/impl/src/main/java/com/blazebit/persistence/impl/function/entity/EntityFunction.java
@@ -57,26 +57,54 @@ public class EntityFunction implements JpqlFunction {
         // We remove the synthetic predicate here and later extract the values clause alias of it so we can insert the proper SQL values clause at the right place
         String subquery = functionRenderContext.getArgument(0);
         StringBuilder sb = new StringBuilder();
-        int subqueryEndIndex = subquery.lastIndexOf(AND_MARKER);
-        if (subquery.regionMatches(subqueryEndIndex - NULL_IS_NULL.length(), NULL_IS_NULL, 0, NULL_IS_NULL.length())) {
-            subqueryEndIndex -= NULL_IS_NULL.length();
-        } else if (subquery.regionMatches(subqueryEndIndex - NULL_IS_NULL_IN_PARENTHESIS.length(), NULL_IS_NULL_IN_PARENTHESIS, 0, NULL_IS_NULL_IN_PARENTHESIS.length())) {
-            subqueryEndIndex -= NULL_IS_NULL_IN_PARENTHESIS.length();
-        }
-        int aliasEndIndex = subquery.indexOf('.', subqueryEndIndex) ;
-        int aliasStartIndex = aliasEndIndex - 1;
-        while (aliasStartIndex > subqueryEndIndex) {
-            if (!SqlUtils.isIdentifier(subquery.charAt(aliasStartIndex))) {
-                aliasStartIndex++;
-                break;
-            }
-            aliasStartIndex--;
-        }
 
         String entityName = JpqlFunctionUtil.unquoteSingleQuotes(functionRenderContext.getArgument(1));
         String valuesClause = JpqlFunctionUtil.unquoteSingleQuotes(functionRenderContext.getArgument(2));
         String valuesAliases = JpqlFunctionUtil.unquoteSingleQuotes(functionRenderContext.getArgument(3));
         String syntheticPredicate = JpqlFunctionUtil.unquoteSingleQuotes(functionRenderContext.getArgument(4));
+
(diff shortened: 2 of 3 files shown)


beikov (approved):
The fix looks good to me, thanks!

PR #2118, opened 2026-06-11
[#2114] Fix count and id query omitting joins referred to by the WHERE clause through lazy path references

### LINKED ISSUE #2114: getCountQuery() fails on entity-view with self-reference sub
`PaginatedCriteriaBuilder.getCountQuery()` throws `org.hibernate.query.SemanticException: Could not interpret path expression 'reversesMovement_1'` when:

1. The projected `EntityView` declares a **self-reference subview** (same target entity, different view type — required to avoid the recursive-subview cycle Blaze rejects at boot).
2. The underlying `CriteriaBuilder` filters on the **same association** by name, e.g. `cb.where("m.reversesMovement").isNull()`.

The fetch-join Blaze emits for the subview gets a Hibernate-generated alias (`reversesMovement_1`). The COUNT query the paginated builder emits later resolves the WHERE-clause path against that alias and fails.

Walking the FK column directly (`cb.where("m.reversesMovement.id").isNull()`) sidesteps the bug.

### Versions

- `blaze-persistence` **1.6.18** (`-core-impl`, `-entity-view-impl`,
  `-integration-hibernate-7.2`)
- `hibernate-orm` **7.2.12.Final**
- Jakarta Persistence 3.2
- PostgreSQL 17 (Testcontainers)
- JDK 25

### Minimal reproducer

https://github.com/heruan/blaze-issues/pull/2 (branch `repro/paginated_count_self_subview`)

```sh
./gradlew test --tests 'to.lova.blaze.issues.paginated_count_self_subview.*'
```

### Entity

```java
@Entity
public class Movement {
    @Id @GeneratedValue Long id;
    @Column(nullable = false) LocalDate date;
    @Column(nullable = false) BigDecimal amount;
    @ManyToOne Movement reversesMovement;
}
```

### Entity views

```java
@EntityView(Movement.class)
public interface MovementView {
    @IdMapping Long getId();
    LocalDate getDate();
    BigDecimal getAmount();
    MovementReferenceView getReversesMovement();   // self-reference subview
}

@EntityView(Movement.class)
public interface MovementReferenceView {
    @IdMapping Long getId();
    LocalDate getDate();
    BigDecimal getAmount();
}
```

### Failing call

```java
EntityViewSetting<MovementView, ?> setting = EntityViewSetting.create(MovementView.class, 0, 50);
setting.addAttributeSorter("id", Sorters.ascending());

var cb = cbf.create(em, Movement.class, "m");
cb.where("m.reversesMovement").isNull();           // <-- direct association

evm.applySetting(setting, cb).getResultList();     // boom on COUNT
```

### Expected

The paginated query runs and returns the seeded rows (here: 2, both with `reversesMovement` null).

### Actual

```
java.lang.IllegalArgumentException: org.hibernate.query.SemanticException:
        Could not interpret path expression 'reversesMovement_1'
    at org.hibernate.internal.ExceptionConverterImpl.convert(...)
    at org.hibernate.internal.AbstractSharedSessionContract.createQuery(AbstractSharedSessionContract.java:1010)
    at org.hibernate.internal.SessionImpl.createQuery(SessionImpl.java:155)
    at com.blazebit.persistence.impl.PaginatedCriteriaBuilderImpl.getCountQuery(PaginatedCriteriaBuilderImpl.java:452)
    at com.blazebit.persistence.impl.PaginatedCriteriaBuilderImpl.getQuery(PaginatedCriteriaBuilderImpl.java:528)
    at com.blazebit.persistence.impl.PaginatedCriteriaBuilderImpl.getResultList(PaginatedCriteriaBuilderImpl.java:604)
    ...
Caused by: org.hibernate.query.SemanticException:
        Could not interpret path expression 'reversesMovement_1'
    at org.hibernate.query.hql.internal.BasicDotIdentifierConsumer$BaseLocalSequencePart.resolveTerminal(BasicDotIdentifierConsumer.java:180)
    at org.hibernate.query.hql.internal.SemanticQueryBuilder.visitSimplePath(SemanticQueryBuilder.java:5786)
    at org.hibernate.query.hql.internal.SemanticQueryBuilder.visitUnaryIsPredicate(SemanticQueryBuilder.java:2407)
    at org.hibernate.query.hql.internal.SemanticQueryBuilder.visitWhereClause(SemanticQueryBuilder.java:2336)
    ...
    at org.hibernate.query.hql.internal.StandardHqlTranslator.translate(StandardHqlTranslator.java:64)
```

The crash is on the COUNT half of the paginated pair; the data query alone (without `EntityViewSetting`-driven pagination) does not reach the failing code path.

### Workaround

Walk the FK column directly via `.id` instead of the association name. This bypasses the implicit-join codepath that collapses onto the fetch alias and the count emitter is happy:

```java
cb.where("m.reversesMovement.id").isNull();   // ✅ passes on 1.6.18
```

Both forms are covered by the reproducer (`crashesOnDirectAssociationIsNull` and `workaroundFkPathSucceeds`) so the workaround acts as a regression guard for the day the upstream fix lands and the idiomatic association form starts working again.

### Related observation in production code

In a real GraphQL projection (large fetch tree, multiple filters), the correlated subquery form

```java
cb.whereNotExists()
    .from(Movement.class, "rev")
    .where("rev.reversesMovement.id").eqExpression("m.id")
    .end();
```

crashes with the same `Could not interpret path expression 'reversesMovement_1'`, and `OUTER(id)` does not help — Hibernate resolves the path against the fetch alias before the `OUTER` substitution runs. We fall back to an explicit `LEFT JOIN ... ON ... WHERE alias.id IS NULL` anti-join, at the cost of Postgres no longer recognising the predicate as a semantic anti-join in `EXPLAIN`. The minimal repro here only triggers the direct-association variant; the correlated NOT EXISTS variant did not reproduce in the trimmed setup, so an additional factor (fetch tree depth?) appears to be required.

heruan:
Fixes #2114

## Root cause

A predicate on a single valued association like `where("m.reversesMovement").isNull()` does not require a join, so `JoinManager.implicitJoin` produces a `LazyPathReference` and registers the WHERE clause dependency only on the path's base node. When another clause creates a default join node for the same association — in the reported case the SELECT clause, through an entity view self-reference subview mapping — `LazyPathReference.getBaseNode()` resolves to that node at rendering time and the predicate is rendered against the join alias (`reversesMovement_1 IS NULL`).

Since that join node only carries a SELECT clause dependency, the paginated count query (clause exclusions `{SELECT, ORDER_BY, GROUP_BY}`) pruned the join while its rendered WHERE clause still referred to the alias. Hibernate 6+ rejects the resulting JPQL with `SemanticException: Could not interpret path expression 'reversesMovement_1'`; the legacy Hibernate 5 parser silently tolerated it. The same applies to the page id query. Walking the FK (`.id`) sidesteps the bug because the single valued id path never goes through the lazy collapse, which matches the workaround reported in the issue.

## Fix

Registering the WHERE dependency on the collapsed node eagerly at implicit join time is not viable: clause dependencies are copied by `CriteriaBuilder.copy()`, and `SingleValuedAssociationManyToOneTest#manyToOneSingleValuedAssociationIsNullAndSelectCopyResolved` codifies that a copy dropping the SELECT clause must also drop the join and re-render the plain association path.

Instead, when building count and id queries, `getCountAndIdQueryAlwaysIncludedNodes` collects the non-root join nodes that the WHERE clause path expressions *effectively* resolve to (the lazy collapse is final by then, so this also covers default nodes created by later-visited clauses such as ORDER BY) plus their ancestors and ON clause dependencies, and passes them through the pre-existing `alwaysIncludedNodes` mechanism of `JoinManager.buildClause`. No shared state is mutated, so the main query, builder copies and cached query strings are unaffected.

## Testing

- New `PaginationTest#testPaginatedWhereOnAssociationImplicitlyJoinedInSelect` asserts the count query now renders `SELECT COUNT(*) FROM Document d LEFT JOIN d.parent parent_1 WHERE parent_1 IS NULL` (previously the join was missing while the alias was referenced)
- Full core testsuite green on H2 + Hibernate 5.6 (2315 tests) and on the Hibernate 7.2 jakarta runner (1967 tests); entity view testsuite green (6037 tests)
- Verified against the original report's stack (entity view self-reference subview, Hibernate ORM 7.2.12, Jakarta Persistence 3.2, PostgreSQL 17, JDK 25) via the standalone reproducer from the issue: both the failing direct-association form and the `.id` workaround now pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ None yet
 ### Bug fixes
 
 * Fix `StringIndexOutOfBoundsException` when using a `VALUES` clause and an inline CTE join inside another inline CTE
+* Fix count and id query omitting joins that the WHERE clause refers to through an association path that is implicitly joined in the SELECT clause
 
 ### Backwards-incompatible changes
 
--- a/core/impl/src/main/java/com/blazebit/persistence/impl/AbstractFullQueryBuilder.java
+++ b/core/impl/src/main/java/com/blazebit/persistence/impl/AbstractFullQueryBuilder.java
@@ -356,7 +356,7 @@ public abstract class AbstractFullQueryBuilder<T, X extends FullQueryBuilder<T, 
             } else {
                 // Collect usage of collection join nodes to optimize away the count distinct
                 // Note that we always exclude the nodes with group by dependency. We consider just the ones from the identifiers
-                Set<JoinNode> identifierExpressionsToUseNonRootJoinNodes = getIdentifierExpressionsToUseNonRootJoinNodes();
+                Set<JoinNode> identifierExpressionsToUseNonRootJoinNodes = getCountAndIdQueryAlwaysIncludedNodes(getIdentifierExpressionsToUseNonRootJoinNodes());
                 Set<JoinNode> collectionJoinNodes = joinManager.buildClause(sbSelectFrom, clauseExclusions, null, true, externalRepresentation, true, false, optionalWhereClauseConjuncts, whereClauseConjuncts, explicitVersionEntities, countNodesToFetch, identifierExpressionsToUseNonRootJoinNodes, null, true);
                 boolean hasCollectionJoinUsages = collectionJoinNodes.size() > 0;
 
@@ -459,7 +459,7 @@ public abstract class AbstractFullQueryBuilder<T, X extends FullQueryBuilder<T, 
             } else {
                 // Collect usage of collection join nodes to optimize away the count distinct
                 // Note that we always exclude the nodes with group by dependency. We consider just the ones from the identifiers
-                Set<JoinNode> identifierExpressionsToUseNonRootJoinNodes = getIdentifierExpressionsToUseNonRootJoinNodes();
+                Set<JoinNode> identifierExpressionsToUseNonRootJoinNodes = getCountAndIdQueryAlwaysIncludedNodes(getIdentifierExpressionsToUseNonRootJoinNodes());
                 Set<JoinNode> collectionJoinNodes = joinManager.buildClause(sbSelectFrom, COUNT_QUERY_GROUP_BY_CLAUSE_EXCLUSIONS, null, true, externalRepresentation, true, false, optionalWhereClauseConjuncts, whereClauseConjuncts, explicitVersionEntities, countNodesToFetch, identifierExpressionsToUseNonRootJoinNodes, null, true);
                 boolean hasCollectionJoinUsages = collectionJoinNodes.size() > 0;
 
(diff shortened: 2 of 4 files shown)


beikov (approved):

beikov:
Thanks!

PR #2119, opened 2026-06-17
Feature/oracle json object aggregate for multiset V2 

guitarmanmike:
Attempting to implement oracle JSON multiset using JSON_OBJECT / JSON_ARRAYAGG similar to PostgresSql and MySql

## Description

Changed the OracleStringJsonAggFunction.java and OracleToStringJsonFunction.java the latter of which is used by TO_MULTISET to use the JSON_ARRAYAGG and JSON_OBJECT functions like other Dbms impls

## Motivation and Context

The existing oracle implementations were using XML conversion and aggregation.

The fall back to use xml conversion and aggregation works but generates very large queries with nested REPLACE() and CHR() literals to cast and convert values to xml then to json. The JSON functions produce much smaller and easier to read/understand queries and should also be more efficient with regards to query plan parsing etc.


core/impl/src/main/java/com/blazebit/persistence/impl/function/tostringjson/OracleToStringJsonFunction.java
```diff
@@ -5,62 +5,47 @@
...
-
-                StringBuilder sb = new StringBuilder(fromIndex);
-                render(sb, fields, selectItemExpressions);
-                context.addChunk(sb.toString());
-
-                context.addChunk(postChunk);
-                context.addChunk(" from lateral(select ");
+                renderJsonObjectArguments(context, fields, fields); //I am not sure if this part is correct limit syntax for oracle? Might need to use the lateral join from OracleToStringJsonFunction (V1)
+                context.addChunk(POST_CHUNK);
+                context.addChunk(" from (select ");
```
beikov: Yeah, I think you will need `lateral` here

```suggestion
                renderJsonObjectArguments(context, fields, fields);
                context.addChunk(POST_CHUNK);
                context.addChunk(" from lateral(select ");
```

guitarmanmike:
Added the lateral select back for the limit

entity-view/testsuite/src/test/resources/logging.properties
```diff
@@ -23,7 +23,7 @@ net.sf.jsqlparser.parser.level = SEVERE
 org.hibernate.level = SEVERE
 org.hibernate.tool.hbm2ddl.level = OFF
 org.hibernate.tool.schema.internal.ExceptionHandlerLoggedImpl.level = ALL
-#org.hibernate.SQL.level = ALL
+org.hibernate.SQL.level = ALL
```
beikov: Please revert that change.

beikov (approved):

PR #2131, opened 2026-09-08
Fix for #2128 - pass the HttpInputMessage to HttpMessageNotReadableException

### LINKED ISSUE #2128: NoSuchMethodError on Spring Framework 7: EntityViewAwareMappingJackson2HttpMessageConverter calls HttpMessageNotReadableException(String, Throwable), removed in Spring 7
### Description

On Spring Framework 7 / Spring Boot 4, any unreadable request body on a controller method whose `@RequestBody` is an entity view fails with `NoSuchMethodError` instead of the intended `HttpMessageNotReadableException`.

[`EntityViewAwareMappingJackson2HttpMessageConverter.readJavaType`](https://github.com/Blazebit/blaze-persistence/blob/main/integration/spring-data/webmvc/src/main/java/com/blazebit/persistence/spring/data/webmvc/impl/json/EntityViewAwareMappingJackson2HttpMessageConverter.java#L68-L81) throws via the two-arg constructor:

```java
private Object readJavaType(JavaType javaType, HttpInputMessage inputMessage) {
    try {
        ...
        return entityViewAwareObjectMapper.readerFor(javaType).readValue(inputMessage.getBody());
    } catch (IOException ex) {
        throw new HttpMessageNotReadableException("Could not read document: " + ex.getMessage(), ex);
    }
}
```

`HttpMessageNotReadableException(String)` and `HttpMessageNotReadableException(String, Throwable)` were deprecated in Spring 5.1 and **removed in Spring Framework 7.0**. Only the `HttpInputMessage`-bearing constructors remain (`javap` on `spring-web-7.0.8.jar`):

```
public class org.springframework.http.converter.HttpMessageNotReadableException extends org.springframework.http.converter.HttpMessageConversionException {
  public HttpMessageNotReadableException(java.lang.String, org.springframework.http.HttpInputMessage);
  public HttpMessageNotReadableException(java.lang.String, java.lang.Throwable, org.springframework.http.HttpInputMessage);
  public org.springframework.http.HttpInputMessage getHttpInputMessage();
}
```

The removed constructor is still the compiled call target in the latest release, 1.6.20 (`javap -c` on `blaze-persistence-integration-spring-data-webmvc-jakarta-1.6.20.jar`):

```
60: new           #76   // class org/springframework/http/converter/HttpMessageNotReadableException
...
87: invokespecial #96   // Method org/springframework/http/converter/HttpMessageNotReadableException."<init>":(Ljava/lang/String;Ljava/lang/Throwable;)V
90: athrow
```

Since `BlazePersistenceWebConfiguration` auto-registers this converter at index 0, it handles every entity-view request body. The bad call site is only on the `catch (IOException)` path — malformed JSON, a truncated body, a client disconnect mid-request — so the happy path is unaffected.

**Suggested fix.** `readJavaType` already receives `inputMessage`, so the three-arg constructor is a drop-in. It has existed since Spring 5.1, so this compiles unchanged on Spring 5.1+ / 6.x too, with no version branching:

```java
} catch (IOException ex) {
    throw new HttpMessageNotReadableException("Could not read document: " + ex.getMessage(), ex, inputMessage);
}
```

This is the only occurrence of the removed constructor in the repo, and the same line appears in both copies of the class:

- `integration/spring-data/webmvc/src/main/java/com/blazebit/persistence/spring/data/webmvc/impl/json/EntityViewAwareMappingJackson2HttpMessageConverter.java:80`
- `integration/spring-data/webmvc-jakarta/src/main/java/com/blazebit/persistence/spring/data/webmvc/impl/json/EntityViewAwareMappingJackson2HttpMessageConverter.java:80`

Happy to open a PR against `main` with the fix plus a test case if that helps.

**Workaround** for anyone hitting this on Spring Boot 4: exclude the auto-configuration and re-register a patched converter.

```yaml
spring:
  autoconfigure:
    exclude: com.blazebit.persistence.spring.data.webmvc.impl.BlazePersistenceWebConfiguration
```

```java
@Configuration
public class PatchedBlazePersistenceWebConfiguration extends BlazePersistenceWebConfiguration {
    // override extendMessageConverters() to insert a copy of the converter
    // that throws HttpMessageNotReadableException(msg, ex, inputMessage)
}
```

### Expected behavior

An unreadable body produces `HttpMessageNotReadableException`, which Spring MVC maps to `400 Bad Request` and which application `@ExceptionHandler` / `ProblemDetail` handling can act on. The underlying `IOException` stays as the cause, so the log shows the real parse failure.

### Actual behavior

`NoSuchMethodError` escapes the converter:

```
java.lang.NoSuchMethodError: org.springframework.http.converter.HttpMessageNotReadableException: method 'void <init>(java.lang.String, java.lang.Throwable)' not found
    at com.blazebit.persistence.spring.data.webmvc.impl.json.EntityViewAwareMappingJackson2HttpMessageConverter.readJavaType(EntityViewAwareMappingJackson2HttpMessageConverter.java:80)
    at com.blazebit.persistence.spring.data.webmvc.impl.json.EntityViewAwareMappingJackson2HttpMessageConverter.read(EntityViewAwareMappingJackson2HttpMessageConverter.java:66)
    at org.springframework.web.servlet.mvc.method.annotation.AbstractMessageConverterMethodArgumentResolver.readWithMessageConverters(...)
```

wrapped by Spring MVC as:

```
Servlet.service() for servlet [dispatcherServlet] in context with path [] threw exception
[Handler dispatch failed: java.lang.NoSuchMethodError: org.springframework.http.converter.HttpMessageNotReadableException:
method 'void <init>(java.lang.String, java.lang.Throwable)' not found] with root cause
```

Consequences: the response is `500` instead of `400`; the original `IOException` is swallowed, so logs show only the `NoSuchMethodError` and not the actual parse failure; and `@ExceptionHandler` handling registered for `HttpMessageNotReadableException` never runs.

### Steps to reproduce

1. Spring Boot 4.0.x application with `blaze-persistence-integration-spring-data-webmvc-jakarta` on the classpath (auto-configuration active, no custom message-converter setup).
2. A controller method taking an entity view as `@RequestBody`:

    ```java
    @PutMapping("/tags")
    public ResponseEntity<TagView> updateTag(@Valid @RequestBody TagUpdateView view) {
        ...
    }
    ```

3. `PUT` a malformed JSON body with `Content-Type: application/json`:

    ```
    curl -i -X PUT http://localhost:8080/tags -H 'Content-Type: application/json' -d '{"name":'
    ```

4. Expected `400`; actual `500` with the `NoSuchMethodError` above. Any other `IOException` while reading the body (truncated body, client disconnect) reproduces it equally.

No database interaction is needed to reproduce — the failure happens during argument resolution, before the handler method runs. Glad to push a test case to the test-case-template repo if the repro above isn't sufficient.

### Environment

```
Version:            1.6.20 (also reproduced on 1.6.18)
JPA-Provider:       Hibernate 7.1.32.Final
DBMS:               PostgreSQL 18 (not involved in the failure)
Application Server: Spring Boot 4.0.7 / Spring Framework 7.0.8, embedded Tomcat, Java 21
```

Blaze-Persistence integrations in use: `spring-data-webmvc-jakarta`, `spring-data-4.0`, `entity-view-spring-6.0`, `hibernate-7.1`, `graphql-jakarta`.

cc @beikov — looks like it belongs to the same Spring Boot 4 wave as #2100, #2102 and #2103.

dsarlo-viso:
Fixes #2128.

`EntityViewAwareMappingJackson2HttpMessageConverter.readJavaType` throws through `HttpMessageNotReadableException(String, Throwable)`. That constructor was deprecated in Spring 5.1 and **removed in Spring Framework 7.0**, so on Spring Boot 4 the `throw` itself fails with `NoSuchMethodError` before the exception is constructed:

```
java.lang.NoSuchMethodError: org.springframework.http.converter.HttpMessageNotReadableException: method 'void <init>(java.lang.String, java.lang.Throwable)' not found
    at com.blazebit.persistence.spring.data.webmvc.impl.json.EntityViewAwareMappingJackson2HttpMessageConverter.readJavaType(EntityViewAwareMappingJackson2HttpMessageConverter.java:80)
```

The in-flight `IOException` is discarded rather than attached as a cause, the response becomes a 500 instead of a 400, and any `@ExceptionHandler` registered for `HttpMessageNotReadableException` never runs.

`readJavaType` already has the `inputMessage` in hand, so the three-arg `(String, Throwable, HttpInputMessage)` overload is a drop-in. It has existed since Spring 5.1, so this compiles unchanged on every Spring version the build targets — no version branching, no profile-specific source.

## Changes

- `integration/spring-data/webmvc/.../EntityViewAwareMappingJackson2HttpMessageConverter.java` — pass `inputMessage`.
- `integration/spring-data/webmvc-jakarta/.../EntityViewAwareMappingJackson2HttpMessageConverter.java` — same, the jakarta copy of the class.
- `DocumentControllerTest` — `// Test for #2128`, a `PUT` with a truncated JSON body expecting `400`. Committed separately as `Test for #2128`, per CONTRIBUTING.
- `AbstractSpringWebMvcTest` — `putRawJson(...)`, so a test can send a body that is deliberately not valid JSON. Mirrors the existing reflective `putJson(...)` so it keeps working across the Spring versions the testsuite compiles against.
- `CHANGELOG.md` — bug-fix entry under 1.6.21-SNAPSHOT.

## Verification

The new test reproduces the failure only when the testsuite runs against Spring 7, i.e. `-Pspring-data-4.0.x` (`version.spring-data-4.0-spring` = 7.0.8). On earlier Spring versions the removed constructor still exists, so it passes with or without the fix — it is a regression guard for the 4.0 profile.

I was not able to run the full Maven testsuite locally. What I did verify directly: compiling the patched `webmvc-jakarta` converter against `spring-web-7.0.8` succeeds, and the emitted call site is now the surviving constructor.

```
$ javap -c ... EntityViewAwareMappingJackson2HttpMessageConverter | grep 'HttpMessageNotReadableException."<init>"'
75: invokespecial #86  // Method org/springframework/http/converter/HttpMessageNotReadableException."<init>":(Ljava/lang/String;Ljava/lang/Throwable;Lorg/springframework/http/HttpInputMessage;)V
```

Happy to adjust the test's placement (a dedicated `Issue2128Test` rather than a case on `DocumentControllerTest`) or anything else you'd prefer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ None yet
 
 ### Bug fixes
 
-None yet
+* Use the `HttpInputMessage` bearing `HttpMessageNotReadableException` constructor in `EntityViewAwareMappingJackson2HttpMessageConverter` so that unreadable entity view request bodies do not fail with `NoSuchMethodError` on Spring Framework 7
 
 ### Backwards-incompatible changes
 
--- a/integration/spring-data/testsuite/webmvc/src/test/java/com/blazebit/persistence/spring/data/testsuite/webmvc/AbstractSpringWebMvcTest.java
+++ b/integration/spring-data/testsuite/webmvc/src/test/java/com/blazebit/persistence/spring/data/testsuite/webmvc/AbstractSpringWebMvcTest.java
@@ -104,6 +104,26 @@ public abstract class AbstractSpringWebMvcTest extends AbstractSpringTest {
         }
     }
 
+    protected RequestBuilder putRawJson(String path, Object idParam, byte[] content, String accept) {
+        MockHttpServletRequestBuilder builder = idParam == null ? put(path) : put(path, idParam);
+        if (CONTENT != null) {
+            try {
+                CONTENT.invoke(builder, content);
+                CONTENT_TYPE.invoke(builder, MediaType.APPLICATION_JSON);
+                if (accept != null) {
+                    ACCEPT.invoke(builder, (Object) new String[]{ accept });
+                }
+            } catch (Exception e) {
+                throw new RuntimeException(e);
+            }
+            return builder;
+        } else {
+            builder.content(content)
+                    .contentType(MediaType.APPLICATION_JSON);
+            return accept == null ? builder : builder.accept(accept);
+        }
+    }
+
     protected RequestBuilder postJson(String path, Object createView) throws JsonProcessingException {
         MockHttpServletRequestBuilder builder = post(path);
         if (CONTENT != null) {
--- a/integration/spring-data/testsuite/webmvc/src/test/java/com/blazebit/persistence/spring/data/testsuite/webmvc/DocumentControllerTest.java
+++ b/integration/spring-data/testsuite/webmvc/src/test/java/com/blazebit/persistence/spring/data/testsuite/webmvc/DocumentControllerTest.java
@@ -24,6 +24,7 @@ import com.blazebit.persistence.spring.data.testsuite.webmvc.view.DocumentCreate
 import com.blazebit.persistence.spring.data.testsuite.webmvc.view.DocumentCreateOrUpdateViewBuilder;
 import com.blazebit.persistence.spring.data.testsuite.webmvc.view.DocumentUpdateView;
 import com.blazebit.persistence.spring.data.webmvc.impl.BlazePersistenceWebConfiguration;
+import java.nio.charset.StandardCharsets;
 import java.util.Collections;
 import org.junit.Test;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -144,6 +145,18 @@ public class DocumentControllerTest extends AbstractSpringWebMvcTest {
         // attempted because the entity view used is both updatable and creatable. The update would fail like before.
         mockMvc.perform(postJson("/documents", createView))
             .andExpect(status().isOk());
+    }
+
+    // Test for #2128
+    @Test
+    public void testUnreadableRequestBodyIsReportedAsBadRequest() throws Exception {
+        // Given
+        Document d1 = createDocument("D1");
+        byte[] truncatedJson = "{\"name\":".getBytes(StandardCharsets.UTF_8);
+
+        // When / Then
+        mockMvc.perform(putRawJson("/documents/{id}", d1.getId(), truncatedJson, APPLICATION_VND_BLAZEBIT_UPDATE_1_JSON))
+            .andExpect(status().isBadRequest());
     }
 
     private Document createDocument(String name) {
--- a/integration/spring-data/webmvc-jakarta/src/main/java/com/blazebit/persistence/spring/data/webmvc/impl/json/EntityViewAwareMappingJackson2HttpMessageConverter.java
+++ b/integration/spring-data/webmvc-jakarta/src/main/java/com/blazebit/persistence/spring/data/webmvc/impl/json/EntityViewAwareMappingJackson2HttpMessageConverter.java
@@ -77,7 +77,7 @@ public class EntityViewAwareMappingJackson2HttpMessageConverter extends MappingJ
             }
             return entityViewAwareObjectMapper.readerFor(javaType).readValue(inputMessage.getBody());
         } catch (IOException ex) {
-            throw new HttpMessageNotReadableException("Could not read document: " + ex.getMessage(), ex);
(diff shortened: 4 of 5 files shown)


beikov (approved):

beikov:
Can you please fix the CI errors?

PR #2133, opened 2026-09-16
[#2132] Fix keyset predicate binding to the last branch only of a top-level disjunction

### LINKED ISSUE #2132: Keyset predicate binds to the last branch only when the WHERE clause is a top-level OR, returning wrong rows from page 2 on
When the WHERE clause of a keyset-paginated query consists of a single top-level `OR` group, the keyset predicate rendered for page 2 onwards is conjoined to it **without parentheses**, so `AND` binds it to the last branch of the disjunction only:

```sql
WHERE <keyset predicate> AND d.name = :param_0 OR owner_1.name = :param_1
```

instead of

```sql
WHERE <keyset predicate> AND (d.name = :param_0 OR owner_1.name = :param_1)
```

Rows matching any branch but the last are therefore returned again on every page (they ignore the keyset), and rows matching the last branch are paged correctly, so the page contents are wrong without any error. Page 1 (no keyset) and the same query without pagination are correct, which makes this easy to miss.

### Root cause

`SimpleQueryGenerator.visit(CompoundPredicate)` renders a compound predicate with a **single child bare**, and only parenthesizes a child compound when the parent has two or more children. The WHERE root is an `AND` compound, so when its only child is the `OR` group, the clause renders as `A OR B` — correct on its own.

`PaginatedCriteriaBuilderImpl` (id query, object query) and `AbstractCommonQueryBuilder.appendWhereClause` then splice the keyset predicate textually, `" WHERE " + keyset + " AND " + whereManager.buildClausePredicate(...)`, without wrapping the base predicate. The same splice shape applies to the join-derived `whereClauseConjuncts` / `optionalWhereClauseConjuncts` inside `PredicateManager.buildClausePredicate` (`conjunct AND A OR B`).

### Reproducer

Against the `Document`/`Person` model of the core testsuite (doc1..doc6, doc4/5/6 owned by Karl4):

```java
CriteriaBuilder<Tuple> crit = cbf.create(em, Tuple.class).from(Document.class, "d")
        .select("d.name").select("d.owner.name");
crit.whereOr()
        .where("d.name").eq("doc1")
        .where("d.owner.name").eq("Karl4")
    .endOr();
crit.orderByAsc("d.name").orderByAsc("d.id");

PagedList<Tuple> page1 = crit.page(null, 0, 2).getResultList();          // doc1, doc4 — correct
PagedList<Tuple> page2 = crit.page(page1.getKeysetPage(), 2, 2).getResultList();
// expected: doc5, doc6
// actual:   doc1, doc5
```

Rendered page id query on `main`:

```
SELECT d.name, d.id FROM Document d JOIN d.owner owner_1 WHERE (d.name > :_keysetParameter_0 OR (d.name = :_keysetParameter_0 AND d.id > :_keysetParameter_1)) AND d.name = :param_0 OR owner_1.name = :param_1 GROUP BY d.name, d.id ORDER BY d.name ASC, d.id ASC
```

Any second top-level predicate (e.g. `crit.where("d.id").isNotNull()`) makes the generator parenthesize the `OR` child and hides the bug, which is the workaround we are using for now.

### Versions

- `blaze-persistence` **1.6.20** and current `main`
- Reproduced on Hibernate ORM 7.2 / PostgreSQL 17 (entity views + `EntityViewSetting.withKeysetPage`, via the GraphQL integration's `createPaginatedSetting`) and on the core testsuite with H2 / Hibernate 5.6, both keyset predicate renderings (`OPTIMIZED_KEYSET_PREDICATE_RENDERING` true and false)

heruan:
Fixes #2132

## Root cause

`SimpleQueryGenerator.visit(CompoundPredicate)` renders a compound predicate with a single child bare and only parenthesizes a child compound when the parent has two or more children. The WHERE root is an `AND` compound, so when its only child is an `OR` group the clause renders as `A OR B`, which is correct on its own.

The keyset-paginated query builders then splice the keyset predicate textually around that rendering, `" WHERE " + keyset + " AND " + whereManager.buildClausePredicate(...)`, in four places (`PaginatedCriteriaBuilderImpl` id query, inline-id object query and object query; `AbstractCommonQueryBuilder.appendWhereClause`). Nothing wraps the base predicate, so `AND` binds the keyset to the last branch of the disjunction only:

```
WHERE <keyset> AND d.name = :param_0 OR owner_1.name = :param_1
```

Rows matching an earlier branch ignore the keyset and come back on every page. The same shape applies to the join-derived `whereClauseConjuncts` / `optionalWhereClauseConjuncts` that `PredicateManager.buildClausePredicate` conjoins itself (`conjunct AND A OR B`).

## Fix

`PredicateManager.buildClausePredicate` parenthesizes the clause predicate when it renders as a top-level disjunction **and** is conjoined with something else: additional/optional conjuncts it renders itself, or — via a new `asConjunct` flag that the four keyset/page-id call sites pass — a predicate the caller renders around it. "Renders as a top-level disjunction" looks through the single-child compounds the generator renders bare, so `whereAnd().whereOr()…` nests are handled too. Nothing changes when the clause stands alone or already has several top-level conjuncts, which keeps every existing expected-query assertion intact.

## Testing

- `KeysetPaginationTest#testKeysetPredicateOnTopLevelDisjunction` and the `OptimizedKeysetPaginationTest` twin: `whereOr()` over `d.name = 'doc1' OR d.owner.name = 'Karl4'`, page size 2. Page 2 returned `doc1, doc5` instead of `doc5, doc6` on `main`; both assert the rendered query now reads `… AND (d.name = :param_0 OR owner_1.name = :param_1) …`, for both keyset predicate renderings
- Both tests fail on `main` and pass with the fix
- `*Pagination*`, `*Keyset*`, `*Where*`, `*Predicate*`, `*Values*`, `*Cte*`, `*Treat*` (525 tests) green on H2 + Hibernate 5.6 with checkstyle, except `treat.jpql.RootTreatTest`, which fails identically (45/72) on unmodified `main` in that local setup
- Originally hit through the entity-view GraphQL integration (Hibernate ORM 7.2, PostgreSQL 17): a `nameContains` filter built as one `whereOr()` group, paged via `EntityViewSetting.withKeysetPage`, returned rows from page 1 again on page 2


beikov (approved):

core/testsuite/src/test/java/com/blazebit/persistence/testsuite/KeysetPaginationTest.java
```diff
@@ -148,6 +148,39 @@ public void testChangingLimit() {
...
+        assertEquals(2, result.size());
+        assertEquals(4, result.getTotalSize());
+        assertEquals("doc1", result.get(0).get(0));
+        assertEquals("doc4", result.get(1).get(0));
+
+        // The keyset predicate must not bind to the last branch of the disjunction only
+        pcb = crit.page(result.getKeysetPage(), 2, 2);
+        String expectedIdQuery = "SELECT d.name, owner_1.name, d.id FROM Document d JOIN d.owner owner_1"
+                + " WHERE (d.name > :_keysetParameter_0 OR (d.name = :_keysetParameter_0 AND d.id > :_keysetParameter_1)) AND (d.name = :param_0 OR owner_1.name = :param_1)"
+                + " ORDER BY d.name ASC, d.id ASC";
```
beikov, line 175:
```suggestion
                + " ORDER BY " + renderNullPrecedence("d.name", "ASC", "FIRST") + ", d.id ASC";
```

core/testsuite/src/test/java/com/blazebit/persistence/testsuite/OptimizedKeysetPaginationTest.java
```diff
@@ -116,6 +116,39 @@ public void simpleNormalTest() {
...
+        assertEquals(2, result.size());
+        assertEquals(4, result.getTotalSize());
+        assertEquals("doc1", result.get(0).get(0));
+        assertEquals("doc4", result.get(1).get(0));
+
+        // The keyset predicate must not bind to the last branch of the disjunction only
+        pcb = crit.page(result.getKeysetPage(), 2, 2);
+        String expectedIdQuery = "SELECT d.name, owner_1.name, d.id FROM DocumentWithNullableName d JOIN d.owner owner_1"
+                + " WHERE ((d.name > :_keysetParameter_0 OR d.name IS NULL) OR (d.name = :_keysetParameter_0 AND d.id > :_keysetParameter_1)) AND (d.name = :param_0 OR owner_1.name = :param_1)"
+                + " ORDER BY d.name ASC NULLS LAST, d.id ASC";
```
beikov, line 143:
```suggestion
                + " ORDER BY " + renderNullPrecedence("d.name", "ASC", "LAST") + ", d.id ASC";
```

beikov (commented):
Seems the failures are cause because of a missing use of `renderNullPrecedence`

heruan:
Applied the `renderNullPrecedence` suggestion on `OptimizedKeysetPaginationTest`, which sorts on `DocumentWithNullableName.name`.

Reverted it on `KeysetPaginationTest`: `Document.name` is not nullable, so no null precedence is rendered on any dialect and CI failed with

```
KeysetPaginationTest.testKeysetPredicateOnTopLevelDisjunction:176
expected:<... ORDER BY d.name ASC[ NULLS FIRST], d.id ASC>
 but was:<... ORDER BY d.name ASC[], d.id ASC>
```

The plain `d.name ASC` matches the other assertions in that class, e.g. `testChangingLimit`.

